Issue #38, opened 2017-04-29
7.1.1 Charter reviews on a public communication channel?


wseltzer:
The Process says team MUST provide member- and team-confidential channels for charter review comments, and that one of those must be the communications default. As we conduct more conversation in public, should we permit a public list to be used and designated for default communications? 

That possibility could be useful where we're chartering based on work that has been incubated in public.

chaals:
Splitting the formal AC review across member-visible and team-confidential discussions is already problematic. Assuming there is a legitimate reason for the Team to accept feedback in confidence, and I believe that is a reasonable assumption, a member who provides such feedback restricts the Team in how they can manage the subsequent discussion of issues. Moreover, it enables an asymmetric discussion to occur, where someone opposed to the substance of the objection has an enhanced scope for "politicking", by for example seeking support from some significant number of members without needing to engage with the issues that were raised, just relying on the number of AC reps who were convinced to go along with an argument.

Adding to this a dimension that recreates the same capacity for chasing up random public statements - given the enormously self-selecting nature of those who would thus be surveyed - seem like a terrible idea.

wseltzer:
And yet, we want to encourage and participate in discussion in public, since many of the charters come from public community discussion and some members discuss with their communities on public lists (Mozilla, for example). Encouraging members to give their feedback in public helps the conversations to stay synchronized.

chaals:
>Encouraging members to give their feedback in public helps the conversations to stay synchronized.

If they *do* have conversations in public. I have certainly spent a lot of breath and hand cartilage encouraging W3C to hold conversations with members rather than in private. And encouraging 2 levels of confidentiality to turn into 3 **does not** help synchronise anything I can think of.

chaals:
> Encouraging members to give their feedback in public helps the conversations to stay synchronized.

If they *do* have conversations in public. I have certainly spent a lot of breath and hand cartilage encouraging W3C to hold conversations with members rather than in private. And encouraging 2 levels of confidentiality to turn into 3 **does not** help synchronise anything I can think of.

wseltzer:
We're trying to become better at public conversation, so thanks for your encouragement.

Several groups are now drafting charters in public repositories, and working through the issues raised there, in public. That's the mode we're recommending all groups use. In those public discussions, it's hard to incorporate "based on team-confidential or member-confidential input, consider this change."

dwsinger:
[...]

PR #39, opened 2017-05-08
Proposed addition of public channel for AC review

wseltzer:
https://github.com/w3c/w3process/issues/38
supported by discussion in Strategy team

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3922,7 +3922,7 @@ Start of a Review Period</h4>
 	Each Member organization <em class="rfc2119">may</em> send one review,
 	which <em class="rfc2119">must</em> be returned by its [=Advisory Committee representative=].
 
-	The Team <em class="rfc2119">must</em> provide two channels for Advisory Committee review comments:
+	The Team <em class="rfc2119">must</em> provide three channels for Advisory Committee review comments:
 
 	<ol>
 		<li>
@@ -3930,13 +3930,16 @@ Start of a Review Period</h4>
 
 		<li>
 			an archived [=Member-only=] channel.
+
+		<li>
+			an archived public channel.
 	</ol>
 
 	The [=Call for Review=] <em class="rfc2119">must</em> specify
 	which channel is the default for review comments on that Call.
 
 	Reviewers <em class="rfc2119">may</em> send information
-	to either or both channels.
+	to any or all channels.
 	They <em class="rfc2119">may</em> also share their reviews
 	with other Members on the <a href="#ACCommunication">Advisory Committee discussion list</a>.
 


koalie:
Hi all,

I was just going to open that very issue, and #38.

In practice we have been doing this since December 2014. I was going to suggest section 7.1.1 be updated to reflect practice.
See the [December 2014 announcement](https://lists.w3.org/Archives/Member/w3c-ac-members/2014OctDec/0046.html) in the W3C Member list.

Quoting the relevant part (and linking to the public list's mail archive):
> Starting with the next review form that we publish, you will find the following options for sharing your review comments in the first drop-down menu of the review form:
>
> - Member-visible
> - Member-visible and send email to w3c-ac-forum
> - Public and send email to both w3c-ac-forum and [public-new-work](https://lists.w3.org/Archives/Public/public-new-work/)
> - Team-only

frivoal:
AC reviews are AC rep's opinions. They can make their own opinion public no matter what we say. Also we also already give them the opportunity to post their review to a public channel if they so desire, with the current implementation of the system.

I really don't see what is objectionable in this PR, and I think we should merge it.

@chaals , as you were pushing back, can you explain why it's a problem to you?

frivoal:
Agenda+ to unblock. Either we merge or we go into what's wrong with this

chaals:
Where review comments are made at different levels of confidentiality, it tends to stifle rather than encourage open discussion.

It is more or less always possible to publish one side of an otherwise confidential debate - indeed we added to the process some obligations on W3C to do so in the case of formal objections to a Director's decision.

But there is a real risk that instead of bringing the discussion into the open, putting some review in public will push people who disagree further into the shadows. If they don't feel comfortable in public view, people have often in practice responded to a public argument by making their counter-argument **less** visible than AC-wide.

This is a tension that we address repeatedly. This proposal seems to be optimising for **making some/most content public** over **transparency of a discussion to those who are formally parties to it**, and that seems to me a poor choice.

(Reasonable people have been recorded disagreeing about the right way to do this over centuries. So "your mileage may vary")

frivoal:
@chaals do you think this tension would be addressed if the system that enables people to make their comments public only lets them do so by also making them in member private space, so that they can be responded to without further friction by people who wish to remain in confidence?

frivoal:
This repository has been switched to using bikeshed. Pull request now need to be done against the master branch (instead of the gh-pages branch), against index.bs (instead of index.html).
This pull request predates that. Rebasing.

frivoal:
As per https://www.w3.org/2020/01/29-w3process-minutes.html and https://www.w3.org/2020/02/12-w3process-minutes.html, this issue is deferred to a subsequent cycle of the Process.

wseltzer:
Propose to close as no longer of interest to anyone but the requester.

dwsinger:
closing as suggested

frivoal:
If we're closing this (which I'm OK with), we should probably close #63 and #38 as well, as they're the same question.

PR #40, opened 2017-05-11
who can dismiss a group member why?

### LINKED ISSUE #17: Should the Process define Business and Community Groups?
Transferred from https://www.w3.org/community/w3process/track/issues/130
State: Raised

### LINKED ISSUE #23: Director can dismis a AB or TAG participant without giving a cause?
Transferred from https://www.w3.org/community/w3process/track/issues/160
State: Raised

chaals:
Fix #23

### DIFF
--- a/index.html
+++ b/index.html
@@ -792,9 +792,15 @@ Advisory Board, Working Group, Coordination Group, Interest Group, W3C Activity,
 
       <h4 id="AB-TAG-vacated">2.5.3 Advisory Board and Technical Architecture Group Vacated Seats</h4>
 
-      <p>An Advisory Board or TAG participant's seat is vacated when either
-        the participant <a href="#resignation">resigns</a>, or an Advisory Board or TAG participant changes affiliations
-        meaning the <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a> are no longer met.
+      <p>An Advisory Board or TAG participant's seat is vacated when either</p>
+      <ul>
+        <li>the participant <a href="#resignation">resigns</a>, or</li>
+        <li>an Advisory Board or TAG participant changes affiliations meaning the
+          <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a>
+          are no longer met, or</li>
+        <li>the Director dismisses the participant for failing to meet the 
+          <a href="#ParticipationCriteria">criteria for participation</a>.
+          
         
       <p>If a participant change affiliation, but the <a href="#AB-TAG-constraints">particiaption constraints</a> are met,
         that participant's seat becomes vacant at the the next regularly scheduled election for that group.</p>
@@ -832,7 +838,10 @@ Advisory Board, Working Group, Coordination Group, Interest Group, W3C Activity,
        [<a href="#ref-cepc">PUB38</a>] and the participation requirements described in
         <a href="https://www.w3.org/Consortium/Patent-Policy#sec-Disclosure">section 6</a> of the
         <a href="https://www.w3.org/Consortium/Patent-Policy">W3C Patent Policy</a> [<a href="#ref-patentpolicy">PUB33</a>].</p>
+      <p>A Working Group Chair may suspend, or the Director may dismiss, a participant from a Working Group
+        or Interest Group for failure to meet these Criteria.</p>
 
+          
       <h4 id="coi">3.1.1 Conflict of Interest Policy</h4>
 
       <p>Individuals participating materially in W3C work <em class="rfc2119">must</em> disclose significant relationships when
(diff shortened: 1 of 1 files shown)


dwsinger:
I think we need to discuss the whole question of whether members of any group can be dismissed 'for cause' and if so, how. What causes? Violations of agreements, policies, processes, or codes? What about non-performance in the case of TAG/AB?

chaals:
@dwsinger:

> we need to discuss the whole question of whether members of any group can be dismissed 'for cause'

Agreed. But I had this change lying around and I figured I would set it up properly. If we reach a conclusion in #17 that is broadly in line, I can adapt the PR easily…

chaals:
Note the interaction with #51

chaals:
Sigh. Should be fixed now.

dwsinger:
The Director may remove for cause a participant in any group (including the AB and TAG), where cause includes violating conditions of this process, the membership agreement, or applicable laws.

   became

Participants in W3C activity must abide by the terms and spirit of the W3C Code of Ethics and Professional Conduct [PUB38] and the participation requirements described insection 6 of the W3C Patent Policy [PUB33].

The Director may suspend or remove a participant from any W3C Group (including the AB and TAG) or activity for failure to meet these requirements.


   which to my eye doesn’t explicitly say that the Director can dismiss for cause, and doesn’t explicitly say that the process, membership agreement, or legal violations are included. I really prefer (a) using the phrase "for cause" which means "For a legitimate, specific reason; with justification (Read more at http://www.yourdictionary.com/for-cause#PwihAGO1qTfh5ySq.99)" (b) naming explicitly the member agreement and process (though I know the former includes the latter, the process is our primary vehicle and (c) saying that legal violations are fair game (imagine, for example, someone convicted for anti-trust behavior in a W3C WG) and (d) using 'includes' so that if there is some other glaring infraction which the Director sees and everyone agrees is fair game (except the perpetrator), the Director can use it and (e) being clear that the special groups AB & TAG that are elected, not volunteered, are included.

I forgot to ask, what about the Advisory Committee?  I suspect that that should be explicitly listed as well.

I would be fine making this something the Director 'decides' and hence could be appealable, I suppose.
"The Director may decide to remove for cause a participant in any group (including the AC, AB and TAG), where cause includes violating conditions of this process, the membership agreement, or applicable laws."

michaelchampion:
IANAL but ...
the wording "for failure to meet these requirements" seems fine to me.  Not sure why adding "for cause" to that sentence adds anything.  But if saying "for cause if there is a failure to meet these requirements" is important to anyone, I won't object.

The member agreement says  "Each member organization is entitled to appoint one individual to the Advisory Committee. I don't see anything  allowing the Director to remove someone from the AC, and adding verbiage to the Process Document doesn't change that -- the normative reference to the process document in the member agreement has to do with the standards review process itself, not the composition of the AC.  So I would advise against adding the ability to remove an AC rep from the process.

dwsinger:
"for cause" is more than "these requirements" (and "these" is a limited and perhaps rather vague set). For cause means "For a legitimate, specific reason; with justification". I am OK with being silent on the AC. I hope we never have to consider suspending someone from the AC for cause. But I do want to be quite explicit that the Director can remove an elected AB or TAG rep for cause.

chaals:
I'm with Mike. A failure to demonstrate the required social competence for your role is a pretty broad decision-making tool, and I don't think we need to continue to enumerate possible things that are included.

dwsinger:
I'm sorry, the requirements listed are:

1. Technical competence in one's role;
2. The ability to act fairly;
3. Social competence in one's role.

The Director can and should not be dismissing for the first, and it's doubtful that they should be dismissing for the third, social incompetence (many engineers are borderline socially competent, and we live with it). Simple members of a group, unlike chairs and editors, are not under very strict fairness requirements either. In fact the next sentence pretty much says that these three are not the Director's problem: "Advisory Committee representatives who nominate individuals from their organization for participation in W3C activities are responsible for assessing and attesting to the qualities of those nominees."

Which leaves us with "Participants in any W3C activity must abide by the terms and spirit of the W3C Code of Ethics and Professional Conduct [PUB38] and the participation requirements described in section 6 of the W3C Patent Policy [PUB33]." Which leaves out (at least) the Process document itself, the Membership agreement, and applicable laws. Not OK. (Though this sentence is important, as it incorporates those two into the process by reference.)

I am looking for an unambiguous statement that the Director can remove someone, if they can justify it by reference to an appropriate rule (W3C) or law. I see no advantage in being vague here.

Please, what is the *problem* with being really clear, and making it also clear that the Director has to be clear? 

I offer this change to the end of the sentence: 

"The Director may suspend or remove a participant from any W3C Group (including the AB and TAG) or activity, for cause, where cause includes a violation of the W3C Membership agreement, this process, the patent policy, or applicable laws."

PR #50, opened 2017-07-12
Editorial: Obsoletion may be because a Rec is superseded

### LINKED ISSUE #36: Do we need a superseded status for older specs?
We have some older recommendations that are still published as they are important references (e.g. HTML 3, 4). The technology in them has been wholly subsumed by later versions, and we no longer want to maintain them actively. Do we need a process to mark them as superseded?

chaals:
Fix #36


chaals:
Rendered obsolete / superseded by #66

PR #54, opened 2017-07-12
Allow representation of multiple members

### LINKED ISSUE #9: Should participants in Working Groups be allowed to represent more than one organisation?
Transferred from https://www.w3.org/community/w3process/track/issues/162
State: Open

chaals:
fix #9

This removes the constraint that a participant cannot represent more than one member.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1240,7 +1240,7 @@ Advisory Board, Working Group, Coordination Group, Interest Group, W3C Activity,
           <dfn id="public-participant-ig">public participants</dfn>.</p>
         <p>Except where noted in this document or in a group charter, all participants share the same rights and responsibilities
           in a group; see also the <a href="#ParticipationCriteria">individual participation criteria</a>.</p>
-        <p>A participant <em class="rfc2119">must</em> represent at most one organization in a Working Group or Interest Group.</p>
+
         <p>An individual <em class="rfc2119">may</em> become a Working or Interest Group participant at any time during the group's
           existence. See also relevant requirements in
           <a href="https://www.w3.org/Consortium/Patent-Policy#sec-join">section 4.3</a> of the


dwsinger:
I don't think we should commit this until we understand (a) whether more than one AC Rep can nominate the same person and (b) whether the organizations of both AC Reps will be shown as members to the WG.

frivoal:
I think this ban is ineffective and not useful, and should be removed. However, I don't think its removal implies that we need to change anything in terms of tooling. The ability for 2 different AC reps to formally nominate the same person to a WG does not seem to be a pressing need.

* From an IP point of view, it only matters that companies are members of the WG, not which individual is with which company.
* Regardless of formal affiliation, individuals can always relay anybody's opinion (e.g. a Googler can easily say: I have heard from Mozilla engineers that they didn't like this). As long as all companies involved have made the IP commitment, this is is not meaningfully different from someone who works part time at two member companies regardless of whether the ration is 50/50 or 95/5.

dwsinger:
We need text to say that both organizations must be members (have joined) the working group, because of the IPR policy.

dwsinger:
we also need the database to show that this person represents two members...let's check with the system team

frivoal:
> we also need the database to show that this person represents two members...let's check with the system team

We should check, but I would not block because of that.

dwsinger:
@frivoal "We should check, but I would not block because of that." I think I would.  

There are lots of formal questions to ask here, alas.
* are the two organizations required to be members of the WG? (I think so).
* are the two AC reps both supposed to nominate the person? (I think we need some way to know whether the person's claim to represent both is true).
* on the rare occasions that WGs vote, how many votes does this person get (N>1, I assume).
* does the database need to record this person as representing both (the question above)? I fear so, to enable transparency and for the community to know what they say is true.

I fear that this is a hypothetical situation which risks opening a can of worms. At the moment, it's asymmetric; one nominates, and the person has to declare a dual interest. Is this proving to be a problem?

nrooney:
**are the two organizations required to be members of the WG? (I think so).**
Yes

**are the two AC reps both supposed to nominate the person? (I think we need some way to know whether the person's claim to represent both is true).**
Not needed.

**on the rare occasions that WGs vote, how many votes does this person get (N>1, I assume).**
Each org gets one vote in votes (as per process doc) so if this one person represents both orgs they can vote twice as long as these are the only two votes for these organisations.

**does the database need to record this person as representing both (the question above)? I fear so, to enable transparency and for the community to know what they say is true.**
No

note to editors: write up.

frivoal:
> are the two organizations required to be members of the WG? (I think so).

Yes, they need to be, but the removal of "A participant <em class="rfc2119">must</em> represent at most one organization in a Working Group or Interest Group." does nothing to allow non members to get representation. 

So for that point, the process is fine as is.

> are the two AC reps both supposed to nominate the person?

I don't that's necessary. As long a everybody involved is a member (which is already ensured), everything is fine. People need to be clear which hat their are wearing, but I don't think more formality is needed.

To me, the main point of removing this clause is to stop getting in the way of people who already are juggling hats. No more formalism is necessary.

So for that point, the process is fine as is.

> on the rare occasions that WGs vote, how many votes does this person get (N>1, I assume)

what @nrooney , plus it's rare enough that chairs can go check with the organizations if the individual is indeed representing them in case there's any doubt.

So for that point, the process is fine as is.


> does the database need to record this person as representing both (the question above)?

No, I don't think we need anything. Again, the point to me is not to allow people to formally claim multiple hats, since there's not much benefit in doing so, just to stop hindering them from doing so informally. No need for big complicated systems for that.

So for that point, the process is fine as is.

-----

I think we can merge as is.

frivoal:
Agenda+ to suggest merging as is,  based on the rationale in the previous comment.

dwsinger:
I think simply deleting the sentence hides this important conversation. I'd prefer that at least consider stating the opposite: 

<p>A participant <em class="rfc2119">may</em> represent more one organization in a Working Group or Interest Group. Those organizations <em class="rfc2119">must</em> all be members of the group. The chair decides whether decisions (including votes) are taken by considering organizations or individuals.</p>

frivoal:
That is reasonably succinct, and quite clear, I like it.

frivoal:
Actually, isn't "The chair decides whether decisions (including votes) are taken by considering organizations or individuals" in contradiction with the existing process on votes?

> Each organization represented in the group must have at most one vote

Maybe just delete the third sentence of your proposal then? Votes are already ruled to be per org, and the rest of the time, chairs do what they want anyway. If you have a rephrasing that does not seem to contradict the part quoted above, I'm happy to take that too.

dwsinger:
> Actually, isn't "The chair decides whether decisions (including votes) are taken by considering organizations or individuals" in contradiction with the existing process on votes?
> 
> > Each organization represented in the group must have at most one vote
> 
> Maybe just delete the third sentence of your proposal then? Votes are already ruled to be per org, and the rest of the time, chairs do what they want anyway. If you have a rephrasing that does not seem to contradict the part quoted above, I'm happy to take that too.

How so a contradiction? If the chair says "this is a technical decision, we're voting here by organization" then the culprit can claim two votes; if the chair says "this is an operational decision (e.g. a meeting location) we're voting individually" then the culprit gets only one vote. In either case, Each organization represented in the group has at most one vote.

frivoal:
Removed the Agenda+ label from the PR to put it on the corresponding issue.

dwsinger:
OK.  I guess it's a contradiction to say that each org gets a vote, and then the chair says that the person representing more than one org gets only one vote.  Go ahead, delete it; I expect decisions-by-person to be obviously only about personal questions anyway.

frivoal:
PR https://github.com/w3c/w3process/pull/219 is the alternative to this one that uses the first 2 sentence of your proposal.

frivoal:
rejected in favor of #219

PR #61, opened 2017-08-03
Amended Recommendation to facilitate Maintenance

plehegar:
This is intended to address #5

### DIFF
--- a/index.html
+++ b/index.html
@@ -2356,7 +2356,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
       <p>Working Groups <em class="rfc2119">must</em> keep a record as errors are reported by readers and implementers.
         Such error reports <em class="rfc2119">should</em> be processed no less frequently than quarterly. Readers of the
         Recommendation <em class="rfc2119">should</em>
-        be able easily to find and see the errata that apply to that specific Recommendation.</p>
+        be able easily to find and see the errata that apply to that specific Recommendation with their associated tests.</p>
 
       <p> Working Groups decide how to document errata. The best practice is a document that identifies itself as based on the
         Recommendation text and clearly identifies the errata and any proposed corrections; other approaches include various
--- a/index.html
+++ b/index.html
@@ -2,8 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta name="keywords" content="W3C Process, Consortium, Team, Recommendation, Advisory Committee,
-Advisory Board, Working Group, Coordination Group, Interest Group, W3C Activity, Workshop, charter, Working Draft, Process Document, Candidate Recommendation, Director, Proposed Recommendation, Submission request">
+    <meta name="keywords" content="W3C Process, Team, Recommendation, Advisory Committee,
+Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working Draft, Process Document, Candidate Recommendation, Director, Proposed Recommendation, Submission request">
     <link rel="stylesheet" type="text/css" href="https://www.w3.org/StyleSheets/TR/base.css">
     <title>World Wide Web Consortium Process Document</title>
     <style type="text/css">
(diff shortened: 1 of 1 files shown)


plehegar:
This is the proposed for the status of the Document (SOTD), part of the Amended Recommendation:

The Recommendation from which this document is derived was produced by a group operating under the 5 February 2004 W3C Patent Policy. [...] This Amended Recommendation was produced by incorporating errata after that group closed.

index.html
```diff
@@ -1773,7 +1773,8 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
           published as a Working Group Note. Working Drafts do not necessarily represent a consensus of the Working Group,
           and do not imply any endorsement by W3C or its members beyond agreement to work on a general area of technology.</dd>
         <dt id="RecsCR">Candidate Recommendation (CR)</dt>
-        <dd class="changed">A Candidate Recommendation is a document that satisfies the Working Group's technical requirements,
+        <dd class="changed">A Candidate Recommendation is a document that satisfies the technical
+          requirements from the Working Group or contains only substantive corrections if there is no Working Group,
```
chaals: Or has been updated to match important changes if

index.html
```diff
@@ -2052,11 +2053,16 @@ <h3 id="candidate-rec"><a id="last-call">6.4 Candidate Recommendation </a></h3>
       <p>The Director <em class="rfc2119">must</em> announce the publication of a Candidate Recommendation to other W3C groups
         and to the public, and <em class="rfc2119">must</em> begin an Advisory Committee Review on the question of whether the
         specification is appropriate to publish as a W3C Recommendation.</p>
+      <p>If the document is intended to become an <a href="#rec-amended">Amended Recommendation</a>, the Director
+        <em class="rfc2119">must</em> still satisfy the requirements to advance to Candidate Recommendation despite the absence
+        of a Working Group.</p>
```
chaals: Why do we need to say this?

index.html
```diff
@@ -2156,10 +2165,11 @@ <h3 id="rec-publication">6.6 W3C Recommendation</h3>
             Committee</a>, other W3C groups and to the public.</li>
       </ul>
       <p>Possible next steps:</p>
-      <p>A W3C Recommendation normally retains its status indefinitely. However it</p>
+      <p>A W3C Recommendation normally retains its status indefinitely. However it <em class="rfc2119">may</em></p>
```
chaals: ... be

vfournier17:
Any modifications need to fall under the Patent Policy.  If a Rec is going to be maintained, the WG should stay open.  If there's not enough interest to keep the WG open, there should be no maintenance on the spec.  Members can't have their cake and eat it too.

chaals:
We produce specs not just for members but for the Web at large. Where major revision is needed, there should indeed be a Working Group, but where a minor change reflects reality better, W3C fulfils its mission better by being able to produce specs that don't perpetuate errors.

dwsinger:
By the way, if the fix is contributed from someone other than the team, and we're outside a WG's commitments, we will need some sort of IPR commitment from the contributor. See #67, which is obviously related.

dwsinger:
Do we need to say that 
   a) a Contribution IPR grant will be sought when appropriate?
   b) that Amended Recs have a warning that the W3C’s IPR grant is formally for the Recommendation on which they are based? (and that the document should be published in such a way as the reader can work out where they stand)?
   Or are these a matter of practice?

chaals:
IMHO they are a mater of practise.

michaelchampion:
@chaals wrote
> IMHO they are a matter of practise.
But putting such language in the Process might make it an easier sell to legal staffs....

chaals:
It might but I am actually skeptical that we gain by making the document look as if the editor was paid by the word - even if there are short-term benefits like fewer meetings with your legal team.

vfournier17:
Chaals - I'm not sure what your comment relates to. Would you please clarify?

Virginia Fournier
Sent from my iPhone
Please excuse iTypos

PR #65, opened 2017-08-22
s/only contains/contains only/

wseltzer:
Clarify restriction. 

A Candidate Recommendation is a document that satisfies the technical requirements from the Working Group or *contains only* substantive corrections if there is no Working Group, and has already received wide review. ...
(instead of "only contains")


chaals:
I don't think "contains" is the right term - presumably it contains an entire spec, including corrections of varoious possible types.

dwsinger:
agree with Chaals, this needs clearer phrasing

PR #69, opened 2017-08-28
There is just one vote in elections, which are STV

### LINKED ISSUE #60: Clarify the voting process
The last line in Section 7.3 (about votes) says

'In the case of Advisory Board and TAG elections, "one vote" means "one vote per available seat".'

I think this line is a holdover from previous voting procedures.  We now use STV.  A literal interpretation of this line is that (e.g.) in an AB election with 4 open seats, each AC rep would have 4 votes: i.e. 4 opportunities to use STV.  This is absurd.

I recommend dropping this line.

chaals:
fix #60


michaelchampion:
There's no consensus on this!

________________________________

dwsinger:
I agree; Jeff suggested this edit, but Mike has not agreed to it.

PR #70, opened 2017-08-28
Revert "There is just one vote in elections, which are STV"

chaals:
Reverts w3c/w3process#69

Misunderstanding of the meeting (I had thought we would make the change that made the document consistent, and raise the issue to the AC).


michaelchampion:
Hmm, I might have said I could live with that, if so sorry for reacting from my mobile :-)

But wouldn't it be appropriate to flag it somehow in the text as needing AC discussion?

chaals:
> wouldn't it be appropriate to flag it somehow in the text as needing AC discussion

IMHO No. We don't flag issues within the text in general when we propose the document for AC review, we instead provide a commentary and let them see issues here. Your mileage may vary with editors...

PR #74, opened 2017-08-30
Any charter can be subject to "extended review"

### LINKED ISSUE #33: Re-craft and re-insert the 60-day delay between charter review and Call for Participation
Transferred from https://www.w3.org/community/w3process/track/issues/181
State: Raised

chaals:
Fix #33
Apply to all charters, not just those that continue existing work.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1430,7 +1430,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         <p>The Director's Call for Review of a substantively modified charter <em class="rfc2119">must</em> highlight
           important changes (e.g., regarding deliverables or resource allocation) and include rationale for the changes.</p>
           
-        <p>In an Advisory Committee review of a charter that continues work on a Working Draft (WD) published under any other Charter,
+        <p>In an Advisory Committee review of a charter 
          any Advisory Committee Member <em class="rfc2119">may</em> request an "extended review". If this occurs, the Director
          <em class="rfc2119">must not</em> issue a Call for Participation for the proposed group until at least 60 days
          after announcing the Call for Review.</p>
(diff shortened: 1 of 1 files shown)


dwsinger:
I can't tell but this PR doesn't seem to contain the text proposed by Virginia and accepted by the Process CG.  Can we simply add that text? (Yes, we'll flag it for discussion in TPAC)

chaals:
Sure, just revert #72 (there's a button there to do so), ignore this, and make a different PR.

As editor I didn't copy the text directly because:

1. It needed to change to account for the comment being in AC review rather than email
1. It takes 2 paragraphs to say the same thing as this PR added to #72, repeating the bulky phrtase in each of them
1. It is the general style in the Process to set a minimum time after trigger A for event B, rather than the "between (a) something and (b) something else" style used in Virginia's text.

I assert that my proposed text has the same effect, of requiring the Director to ensure a minimum 60-day gap between charter review starting and call for participation, any time an AC rep requests that as part of a charter review. If not, it does need fixing...

vfournier17:
I can't seem to locate Charles' proposed text.  Would you please post again?  Also, the extended 60-day review is intended to cover both situations; i.e., new charters and rechartering charters (for example) adding new deliverables.  The reason for the proposal is that some members may look at a charter that is in one of their technology sweet spots and say, "Oh, no - this patent review is going to take more than 30 days."  This way, the review period could be extended when a member needs more time for a patent review (for any new deliverables), but would not be imposed in every case.

vfournier17:
This is the most recent language I can parse together:
"With respect to any new or substantively modified Working Group charter, any Advisory Committee representative may request that there be at least 60 days between (a) the Call for Review, and (b) the Call for Participation for such a group (an “Extended Review Period”).

An Extended Review Period request must be submitted as part of a Member's comments submitted during the Call for Review.  Upon receipt of such request, the Director must ensure that there are 60 days between the Call for Review and the Call for Participation. The request for an Extended Review Period must be made prior to the end of the Call for Review period."

chaals:
overtaken by #93 so closing

PR #102, opened 2017-09-21
Minor editorials

dwsinger:
most notably, be consistent in capitalizing Candidate Rec. And take the Brain out of Kardell…


dwsinger:
also:

 "A Candidate Recommendation is a document that satisfies the technical requirements of the Working Group that produced it and their dependencies,   “

I can’t work out what “their dependencies” connects to…who/what is referred to as “their”, and what dependencies?

Ben Wilson — is/was with Digicert, is that the right affiliation?

The document is quite inconsistent about whether it’s “W3C” or “The W3C”. Personally, I prefer the latter, “The world-wide wonder comfortium”. But consistency would be good…

PR #106, opened 2017-09-27
Cleanup

chaals:
Ensure AC review *does* start (at PR)
Update changelog

### DIFF
--- a/index.html
+++ b/index.html
@@ -2159,7 +2159,8 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
       <p>The Director:</p>
       <ul>
         <li><em class="rfc2119">must</em> announce the publication of a Proposed Recommendation to the
-          <a href="#AC">Advisory Committee</a>, and</li>
+		<a href="#AC">Advisory Committee</a> and begin an <a href="#ACReview">Advisory Committee Review</a>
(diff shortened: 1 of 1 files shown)


LJWatson:
I still do not understand why we want to open the ballot at PR? If the intent of the original change was to enable early feedback by opening the ballot at CR and running it through PR, reverting to opening the ballot at PR seems like a step back.

If we open the ballot at CR and run it for a limited period of time (perhaps 4 weeks), the spec should be stable enough for the AC to make a reasoned decision about its readiness to progress. If there are concerns, then it enables the early feedback the original change sought to enable. 

Feedback, especially concern about readiness to progress,  without the formality of a ballot, seems unlikely to have much impact.

dwsinger:
We have to have a formal AC ballot at some point as to whether the spec. should be published as a Rec. We can't do that at CR because there are questions of implementability, features at risk, and so on. So, we notify the AC at CR (so they can raise any general objections) and then  do the formal ballot at PR. Yes, it should be a question of the AC confirming that i's are dotted, t's crossed, and all is hunky-dory, but we have to ask at some point.

LJWatson:
Thanks @dwsinger. Understood. As you were :)

chaals:
I'm assuming @wseltzer will merge this.

chaals:
Oh. This has been overtaken by a patch from Wendy.

PR #119, opened 2017-10-17
careful what you ask for

chaals:
This is speculative...


dwsinger:
Many thanks to both of you

PR #134, opened 2017-11-28
Fix several broken links

xfq:


dwsinger:
oops, my fault both of these.  sorry, thanks

Issue #141, opened 2017-12-07
provide clearer/common wording for transitions to Obsolete/Superseded status


tantek:
The process document should provide clearer wording, and/or re-use common wording from other states for how and with what expected time periods transitions to/from Obsolete/Superseded status happens - in part to address the apparent lack of progress on obsoleting documents as noted: https://www.w3.org/wiki/AB/2017_Priorities#Maintenance and also to provide clear(er) escalation/appeal paths.

jeffjaffe:
Tantek, can you make a text proposal?

nrooney:
Waiting on proposal from @tantek

dwsinger:
will close soon if no action. ping @tantek

cwilso:
I still think this should be done.

dwsinger:
consider closing in the absence of action

PR #145, opened 2017-12-19
clarify withdrawing a Rec

### LINKED ISSUE #138: Editorial improvements to obsolete/superceded
From Andreas in AC review:

#### Change
* Section 6.9
* Description: A Recommendation can be declared superseded as well as obsolete.

### Comment
The process is not clear enough. Example:

> W3C may obsolete a Recommendation, for example if the W3C Community decides that the Recommendation no longer represents best practices,

Who is the W3C Community?

In general, the requirements to start a process that obsoletes (rescinds or supersedes) a recommendation are not defined well enough.

> [dwsinger] Hm, we define four very specific cases. What is missing?

It may be just a clarification that this is the only way to start the process. You first define the option that W3C may obsolete a Recommendation. Then you describe the process by starting with "The director may recommend ....". It is not clear from the reading that the described steps are the only, exclusive process  steps to come to an obsolete Recommendation. So it may just need an editorial edit to clarify this.

chaals:
Fix #138 and #141 

Where should proposals be sent? AC mailing list?

### DIFF
--- a/index.html
+++ b/index.html
@@ -2487,39 +2487,35 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
 
       <p>It is possible that W3C decides that implementing a particular Recommendation is no longer recommended.
         There are three designations for such specifications, chosen depending on the advice W3C wishes to give about
-        further use of the specification.</p>
+        further use of the specification:</p>
+	    
+            <dl>
+	      <dt><a href="RecsSup">Superseded Recommendation</a></dt>
+	        <dd>Used when W3C prefers to recommend a newer specification.</dd>
+	      <dt><a href="RecsObs">Obsolete Recommendation</a></dt>
+	        <dd>Used when W3C no longer recommends implementation of the technology described in a specification, 
+		      e.g. because it appears unlikely to gain sufficient market acceptance to support interoperability.</dd>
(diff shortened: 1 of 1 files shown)


chaals:
If this is the right direction, need to determine how proposals are made to "W3C" before merging.

dwsinger:
Looks plausible but not in the list of 2018 minor post-ballot cleanups

dwsinger:
do we need to re-examine this? it fixes some editorials so I would kinda like to leave it to the editor...

dwsinger:
Over to Natasha to handle, integrate as appropriate (or reject, or reword, etc. as she sees fit)

nrooney:
I actually like the text as it is - short and to the point. The AC can receive an email to the official AC list - that should be sufficient. @chaals if you add this I think we're ok to go.

dwsinger:
yes, I think this cleanup can happen with notification to the AC list. not sure Chaals' input is needed

nrooney:
note to self: review - check if editorial, and bring it in if it is, and if not bring back to CG

nrooney:
On further review, this PR is not ready, please review, edit and re submit

jeffjaffe:
Why in old line 2515 are we taking the Director out of the loop?  Tantek points out in new line 2515 there is an @@/HOWWHERE.

PR #147, opened 2017-12-19
Remove alternative update mechanism

### LINKED ISSUE #146: Can't update process as if it were a Rec
There is no call for implementations, etc, so offering the Rec track as an alternative path makes no sense.

chaals:
fix #146


dwsinger:
I removed the label; does anyone disagree with this PR?

dwsinger:
Editors to merge whenever they feel ready (agreed on the process call)

nrooney:
Thanks!

PR #153, opened 2017-12-24
make dfn elements linkable

chaals:
A few definitions had crept in that were not linkable. This fixes that.


plehegar:
This seems straightforward to me. @dwsinger , can we merge this so that it can make it in Process 2018?

https://github.com/w3c/modern-tooling/issues/71

PR #158, opened 2018-01-16
Only substantive changes do require Director's approval for CR

plehegar:
One minor fix for the document in the changes section, otherwise it is misleading.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3417,8 +3417,6 @@ Superseded is the same as for declaring it Obsolete, below; only the name and ex
         <dt><a href="#WGCharter">Section 5.2.6</a></dt>
          <dd>Charters <em class="rfc2119">must</em> include documentation of any voting procedures additional to those defined 
          in <a href="#Votes">Section 3.4 Voting</a>.</dd>
-        <dt><a href="#revised-cr">Section 6.4.1</a></dt>
-        <dd>All substantive changes to Candidate Recommendations need Director's approval.</dd>
         <dt><a href="#revised-rec">Section 6.7.2</a></dt>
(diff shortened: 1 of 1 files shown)


dwsinger:
Hm, are you sure we don't need Director's assent that proposed changes are not substantive (and hence would need wider review)? Just asking, but is the mental model "any change -> get the Director's OK; substantive change -> get the AC and community OK"?

plehegar:
Section 6.4.1 only talks about substantive changes as far as I can tell so the change log is misleading imho and isn't an actual change from Process 2017 to Process 2018. We do allow Groups to publish new CR with only editorial changes without Director's approval:
 https://www.w3.org/Guide/transitions?profile=CR&cr=editorial
(compare that with https://www.w3.org/Guide/transitions?profile=CR&cr=substantive )
If the Group misjudged the change, the Director will notice it before the approval to Proposed Recommendation (the diff with the document previously approved by the Director will highlight the change) and would force the Group to republish a CR instead (and trigger a call for exclusions).

swickr:
The 2017->2018 changes to 6.4.1 do not appear to read on whether _all_ changes to a CR require Director's review or only substantive changes.   Appendix 14 is not the appropriate place to put a clarification of that point.

As I read it, the _MUST_ in 6.4.1 continues to apply to "substantive changes" earlier in the sentence.

plehegar:
@dwsinger should I revise this PR to simply remove the mention of section 6.4.1 in the change log?

dwsinger:
I am remarkably relaxed about what change logs say since they are only informative (and almost no-one reads them, except to find out when something happened)

plehegar:
ok. removing mention of 6.4.1 in the changelog then.

Issue #172, opened 2018-03-09
Does a spec need to be Rec-ready to be a CR? Align defn with process


nigelmegitt:
https://www.w3.org/2018/Process-20180201/#candidate-rec sets out requirements for advancement to CR. It does not require any statement of reasons why a CR may [update: instating omitted text from original issue filing, with apologies] _not_ advance to Rec. [/update]

However https://www.w3.org/2018/Process-20180201/#RecsCR defines what a CR is and adds informatively:

> Note: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a different next step should include the reasons why the change in expectations comes at so late a stage.

This (first part) is not tested in the requirements for advancing to CR, and in fact it is not unusual for substantive changes to be made after first CR, on the path to Rec. The second part is very reasonable, however it omits an important real world case, which is non-advancement for an unreasonably long period. 

As a minimum, the two sections should be aligned so that the tests for advancement clearly verify the requirements.

My preference for achieving the alignment would be to remove or modify the expectation from the note, since I don't think it reflects reality or is particularly helpful. It is obvious enough that a CR is a step along the path to Rec and that there are extra steps needed if making substantive changes between CR and Rec. Conversely, if the substantive content of a CR is acceptable as a Rec, there are good reasons why we do not just publish it as a Rec: the "acceptability" depends not on the document in itself, but rather on the demonstration that implementers and the AC and Director agree that it is acceptable. So the statement of expectation doesn't seem to be clear enough.

I would propose:

> Note: Candidate Recommendations are expected to be advanced to Recommendation. Announcement of a different next step should include the reasons why the change in expectations comes at so late a stage.

Handling the second problem, omission of what should happen if a document stays in CR for an unreasonably long time, is harder to fix and probably needs some discussion. Automatically revert the CR to a Note? I imagine that would be unacceptable to some people, but it might focus attention on tests and implementation reports... If it is going to be picked up for discussion, probably better for someone to raise a new issue for tracking purposes.

michaelchampion:
The phrase "Candidate Recommendations are expected to be acceptable as Recommendations" was put in the Maturity Levels section during the major revision in 2014, but I don't recall what purpose it was supposed to serve. It's been there for 4 years, what problem is it causing?

[...]

PR #181, opened 2018-04-11
Update note per #172

wseltzer:


dwsinger:
Hm. I think we need to say something. "Candidate recommendations are expected to be good enough that implementation and experience can happen"?

nigelmegitt:
@dwsinger that's proxy text for the bigger question of expectation that it can be advanced, according to whatever criteria advancement requires. Hence my proposal at https://github.com/w3c/w3process/issues/172#issue-303833491 :

> Candidate Recommendations are expected to be advanced to Recommendation. 

Probably better not to try to mix in advancement criteria here, since that creates an editing dependency should those advancement criteria ever change.

frivoal:
@chaals can you clarify which version looks good to you? the original PR from @wseltzer , or the proposed variant by @nigelmegitt or @dwsinger ?

dwsinger:
see conversation in #172. Some unease about going this way...

nrooney:
It may be better to state in 6.4 "CR should be acceptable as a Rec". Comments?

wseltzer:
Close per Sept 12 meeting

dwsinger:
Looks like a more complex pull is needed, see #172. Closing and waiting for a more comprehensive fix.

Issue #83, opened 2017-09-11
Written notification?


LJWatson:
In section 3.6 Resignation from a group, it says "On written notification from an Advisory Committee representative...". Does the notification from SysBot (when an AC rep unhooks someone from the WG/IG) qualify as "written notification"?

dwsinger:
I think in this sense the contrast is with verbal. i.e. telling the team over a coffee doesn't work

LJWatson:
Makes sense. Wonder if "When an AC rep withdraws a participant using the XYZ mechanism, or on other written notification..." (where XYZ is the proper name for that thing) would make it clearer?

chaals:
I think what we mean is archived evidence that the member resigns. I certainly believe that unhooking counts.

I would rather leave this as is myself, but others may differ... Looking for a decision before I make a change.

LJWatson:
I think the text could be clearer (suggested text ^^). Not a showstopper if there isn't much conviction for changing though.

chaals:
I agree we should think about how to make it clearer. I'm not sure I want to use 'through the "patent-policy implementation" form or other written notification' in the text though.

dwsinger:
"On written notification from an Advisory Committee representative or Invited Expert to the team, the Member and their representatives or the Invited Expert will be deemed to have resigned from the relevant group."

Does this sentence add anything? Could we just delete it? Or maybe we should say that they are deemed to have resigned when they are not shown as members of the group, but previously were? Or does that leave us open to possible database errors?

chaals:
I think just asuming the database magically gets it right is a bad idea. And sending a note to W3C saying "We are resigning from group X now" should work, IMHO, as well as doing it through whichever baroque interface is currently working or you happened to find through the website search system...

LJWatson:
Perhaps just "formal notification"?

chaals:
Except that's pretty vague in what counts and doesn't. If I stand up at the coffe table and say it formally, or if I am wearing a suit and tie instead of shorts and bare feet, or if I write it on letterhead...

Yes these are silly examples, but the point is that written gives some clearer hints. I think what we mean is in an archivable form, directed to a person or system that should be empowered to act on the notification. But that's something of a mouthful, and not necessarily clear when people read it.

dwsinger:
deferred to next year

dwsinger:
maybe "recorded" notification?

chaals:
[...]

PR #184, opened 2018-05-03
notifications should be recorded

nrooney:
#83  Adding that resignation notifications should be recorded


dwsinger:
MUST -> <em class="rfc2119">must</em>

chaals:
Requirements should indicate who is responsible for meeting them (in this case, I think the Team). I dealt with this and using the rfc2119 class in #188 but didn't specify the mechanism required.

(Sorry for the mess of 2 PRs. But given that it's around one line, I figure you can pick what  you like and close the other one).

Issue #163, opened 2018-02-01
Representativity on the TAG


ylafon:
Hi,
The Process2018 text says in 2.5.1:
`A Member organization is permitted at most one participant on the TAG except when having more than one participant is caused by a change of affiliation of an existing participant.`
What happens if a non-member organisation manage to have two people elected ?
Striking 'Member' would solve that.

@chaals added in #118 the following:
> Another aspect of that issue is when an organisation nominates an employee one year, and a non-employee the next. This is not uncommon.

dwsinger:
The nomination-of-others question is thorny. What happens if an org. nominates and agrees to support an IE for the good of the community but is otherwise hands-off? I think that asking people to give reasonable transparency to relationships is the best we can do (for example, what happens if one member places contracts that are a major source of funding for another member?).

I'm fine with striking "member"...

jeffjaffe:
+1 to David's comments.

dwsinger:
we probably need to talk about transparency in relationships and so on as well

dwsinger:
Let's strike "member" for now.

dwsinger:
closed via #185

Issue #176, opened 2018-03-28
The term "Affiliation" is used in several places but not strictly defined


cwilso:
Although a "change of affiliation" is a triggering event in several cases, it is not strictly defined what an affiliation is.  Is it employment?  Contract?  Simple statement of "yes this individual speaks for us"?

May be related to issue #9 .

Kometo2:
Well, i get endpoint

chaals:
(This seems tangentially related to #9 but pretty closely linked to #175)

I think the key reference is [Section 3.1.2 Member Representatives](https://www.w3.org/2018/Process-20180201/#member-rep)

**If** we agree that "nominated as a representative by a member" is what affiliation refers to - my untested assumption - moderated by the requirements to notify the team of [Conflicts of Interest](https://www.w3.org/2018/Process-20180201/#coi) and the [Related Member provisions](https://www.w3.org/2018/Process-20180201/#MemberRelated) - perhaps the answer is to link the "change of affiliation" in the [section on triggering elections](https://www.w3.org/2018/Process-20180201/#AB-TAG-vacated) to section 3.1.2?

I note that the term "affiliation" is only used in the document in the sections mentioned above.

nrooney:
Er, I am also going to link #179. Seems like we have a problem here (the only reasonable half-definition is in the [Conflict of Interest Policy](https://w3c.github.io/w3process/#coi)). Let me take a stab at it:

`disclaimer: this is strict. I added everything I could think of. Please feel free to add or suggest dialling it back!`

## Affiliation
Individuals participating within W3C will often have an "affiliation" to one or more organisations (@todo define organisations). "Affiliation" to an organisation will often be through employment (the individual is an "employer" of the organisation) although individuals MAY be "affiliated" through other means. The circumstances in which an individual is "affiliated" with an organisation are described below:

* The individual is an employer of an organisation whose activity is relevant to W3C
* The individual is being paid for consultancy work by an organisation whose activity is relevant to W3C
* The individual's participation in W3C is being supported financially by an organisation whose activity is relevant to W3C (travel, accommodation and associated travel costs)
* The individual is being contracted for work by an organisation whose activity is relevant to W3C
* The individual holds a seat on a board or leadership level group within the organisation whose activity is relevant to W3C
* The individual holds a seat on an advisory group within an organisation whose activity is relevant to W3C
* The individual holds and elected position on a board or group within an organisation whose activity is relevant to W3C
[...]

PR #185, opened 2018-05-03
remove member

nrooney:
remove member from "member organisation" for tag and AB rep restrictions #163

### DIFF
--- a/index.html
+++ b/index.html
@@ -749,11 +749,11 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         W3C Members is represented:</p>
 
       <ul>
-        <li>A Member organization is permitted at most one participant on the TAG except when having more than one participant
+        <li>An organization is permitted at most one participant on the TAG except when having more than one participant
           is caused by a change of affiliation of an existing participant. At the completion of the next regularly scheduled
-          election for the TAG, the Member organization <em class="rfc2119">must</em> have returned to having at most
+          election for the TAG, the organization <em class="rfc2119">must</em> have returned to having at most
           one participant.</li>
-        <li>A Member organization is permitted at most one participant on the AB.</li>
+        <li>An organization is permitted at most one participant on the AB.</li>
         <li>An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.</li>
       </ul>
 


dwsinger:
LGTM

chaals:
I think this needs to be considered alongside who nominated someone and the rest of #176 

Example 1:
Jo works for W3C member Ace Accessibility, but is on the board of Accessibility Braves which certifies accessibility bug hunters and is not a W3C member.
Charlie works for W3C member Cool Access, and is on the board of Accessibility Braves.

If each is nominated by their employer, can Charlie and Jo both be on the TAG?

Can they both be nominated by Cool Access, at successive elections?

Does it matter whether Accessibility Braves is formally incorporated (i.e. a legal entity, or "juridical person") or not (e.g. like W3C)?

Example 2:
Jo works for Ace Accessibility, but as Accessibility Braves became a member, Jo is their AC rep.
Yuan is Ace Accessibility's AC rep.

Can Jo and Yuan each nominate themselves to the AB?

Example 3:
Charlie's employer Cool Access subcontracts Biddable Worx, a W3C member, to represent them in W3C working groups.
Jan works for Biddable Worx, and is working on that subcontract.

Can Charlie and Jan both be on the TAG?
Does it make a difference whether the contract represents 0.8% or 80% of Biddable Worx' income?

dwsinger:
Perhaps we can state that a change of affiliation is either (a) a change of employment or (b) the originally nominating member organization withdraws their support, i.e. would not re-nominate now?

Other conflicts, affiliations, interests, I think can be the part of reasonable disclosure.

cwilso:
That last statement seems more appropriate to #176 .

frivoal:
Not objecting, but I am not sure I see the need for the removal of "Member" here. If an individual is part of an organization that's not a member, then presumably they are an invited expert and represent themselves, rather than that organization. So organizations which are not members cannot have anyone on the TAG or the AB, so limiting them to at most 1 does nothing since they can only have 0 anyway.

chaals:
It is fairly straightforward for two member organisations to nominate people who happen to work for the same non-member organisation. It seems anomalous to allow such a case to provide a non-member with a position that we explicitly deny to members.

frivoal:
I guess so. My working assumption was that the non-member company would be unrelated to W3C, and that the restriction was therefore not particularly useful. But maybe this assumption is not justified. Anyway, even if I am not 100% sure it is useful, I don't see the harm, and it is fewer words in the end.

Issue #175, opened 2018-03-28
Contrast 'immediately' vacated seats and seats vacated at the next election


dwsinger:
in 2.5.3 this section appears to define the four ways a seat gets vacated:
"An Advisory Board or TAG participant's seat is vacated when:

* the participant resigns, or
* an Advisory Board or TAG participant changes affiliations such that the Advisory Board and TAG participation constraints are no longer met, or
* the Director dismisses the participant for failing to meet the criteria for participation in section 3.1, or
* their term ends."

However, this sentence immediately after adds another. 

"If a participant changes affiliation, but the participation constraints are met, that participant's seat becomes vacant at the the next regularly scheduled election for that group."

This should be clarified by inserting "immediately" into the sentence before the bullets:
"An Advisory Board or TAG participant's seat is vacated immediately when:"

chaals:
That would be contrary to long-standing practice, and I think unhelpful. The result would be to force a special election or to reduce the number of people on the relevant body by one until the next regular election has taken place, 

Given that it is a responsibility of TAG and AB members to represent the best interests of the Web and W3C, and **not** their sponsor, and they are therefore elected on that basis, this seems unnecessarily bureaucratic.

It also runs counter to the last change we made to this area, in which TAG participation requirements were loosened specifically to avoid this happening in the common case.

dwsinger:
You're right, bullet point (2) is no longer an immediate vacation.

I'm trying to make sure that the document is clear and readable, not change anything.

An Advisory Board or TAG participant's seat is vacated when:

* the participant resigns, or
* an Advisory Board or TAG participant changes affiliations such that the Advisory Board and TAG participation constraints are no longer met, or
* the Director dismisses the participant for failing to meet the criteria for participation in section 3.1, or
* their term ends, or
* the participant changes affiliation, but the participation constraints are met."

But actually, after a change of affiliation: 
* when the constraints aren't met, we resolve at the next election
* when the constraints are met, the seat is vacated at the next election.

We could simplify the text further.
"An Advisory Board or TAG participant's seat is vacated when:

* the participant resigns, or
* an Advisory Board or TAG participant changes affiliations, or
* the Director dismisses the participant for failing to meet the criteria for participation in section 3.1, or
* their term ends."

and we keep the following sentence, verbatim:
"If a participant changes affiliation, but the participation constraints are met, that participant's seat becomes vacant at the the next regularly scheduled election for that group."

Kometo2:
[...]

PR #186, opened 2018-05-03
Add text on vacating TAG and AB seats

nrooney:
Added some text to help clarify. #175

### DIFF
--- a/index.html
+++ b/index.html
@@ -848,7 +848,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         <li>the end of their term.</li>
       </ul>
 	    
-      <p>An Advisory Board or TAG participant's seat is becomes vacant at the the next regularly scheduled election 
+      <p>An Advisory Board or TAG participant's seat becomes vacant at the the next regularly scheduled election 
 	for that group when the following applies:</p>
       <ul>
         <li>If a participant changes affiliation, but the <a href="#AB-TAG-constraints">participation constraints</a> are met.</li>
--- a/index.html
+++ b/index.html
@@ -834,20 +834,25 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
       </ol>
 
       <h4 id="AB-TAG-vacated">2.5.3 Advisory Board and Technical Architecture Group Vacated Seats</h4>
+	An Advisory Board or TAG participants seat can become vacated immediately or at the next regularly scheduled
+	election depending on the circumstances stated below:
 
-      <p>An Advisory Board or TAG participant's seat is vacated when:</p>
+      <p>An Advisory Board or TAG participant's seat is vacated immediately when one of the following applies:</p>
       <ul>
-        <li>the participant <a href="#resignation">resigns</a>, or</li>
+        <li><a href="#resignation">resignation</a> by the the participant(on the resignation date), or</li>
         <li>an Advisory Board or TAG participant changes affiliations such that the
           <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a>
           are no longer met, or</li>
         <li>the Director dismisses the participant for failing to meet the 
-          <a href="#ParticipationCriteria">criteria for participation in section 3.1</a>, or</li>
-        <li>their term ends.</li>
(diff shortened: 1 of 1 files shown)


dwsinger:
Don't think so; I think we recently changed that the TAG seat is NOT vacated immediately when affiliation change causes duplicate membership, but must be resolved at the next election.

chaals:
@dwsinger not sure which bit you are referring to. The participation constraints explicitly have different rules for AB compared to TAG, to enable the case you outline.

In any event, I don't see anything in the pull request that changes that particular case.

dwsinger:
My read of the PR is that it says
<p>An Advisory Board or TAG participant's seat is vacated immediately when one of the following applies:</p>
…
<li>an Advisory Board or TAG participant changes affiliations such that the
            <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a>
            are no longer met, or</li>


which is not true for the TAG; the seat is NOT vacated immediately; the participation constraints have to be satisfied by the end of the next regularly scheduled election.

What am I missing?

frivoal:
I think another relevant bit worth clarifying is what "change of affiliation" means. Some cases are easy, some are not:

Easy case:
1. A person used to be an employee / director of a member company, now is an employee / director of a different member company

Various less obvious cases:

2. A person used to be self-employed, has now incorporated.
3. A person used to be an IE, now (without change of employment or incorporation), has become a paying member…
    1. … because they just joined as an individual
    2. … because the company they own/control just joined
    3. … because the company that employs them just joined
4. 2 + 3.ii at the same time
5. A person used to be an employee / director of a member company, now is a contractor of the same member company
6. A person used to be a contractor for several member companies, now is an employee / director of one of them
   1. part time, and retains their previous contracting clients
   2. full time, and no longer works for their previous contracting clients.
7. A person used to be an employee or contractor of a member company, now no longer receives any financial compensation from that member company, but the company is still willing to let the person participate in their name
8. A person who used to be employee / director in one member company continues to be, but now is additionally director / employee in one additional company…
   1. … which is a member company
   2. … which is a not member company
9. A Invited Expert used to be self employed, and now is the employee of an company which is not a member
10. A Invited Expert used to be and employee / director of a company which is not a member, and now is the employee / director of an other company which is also not a member

(Plus, any of the above in the opposite direction.)

While I don't think all of the above have necessarily happened to AB/TAG members yet, I can think of concrete examples of individuals active in W3C for pretty much all of the above (hell, I've probably been through half of these myself…)

The above laundry list is certainly not something we should incorporate into the process directly, but some high level principle by which we could judge which case falls on which side would be welcome.

----
My take is that:

2, 3.i, 3.ii, 4, 5, 7, 8.ii, 9, 10 don't seem useful to me to classify as changes in affiliation

3.iii, 6.ii are probably worth counting as changes in affiliation

6.i, 8.i I don't know.

----
I think the underlying principle behind that classification is that, as far as their w3c activity is concerned, if a person switches between being someone's agent to someone else's agent, or from being someone's agent to representing themselves, it is a change of affiliation, and otherwise, it is not, and that you can only be someone's agent in w3c if that someone is a member (otherwise you're representing yourself).

frivoal:
Just realized that my previous comment was more appropriate for a different issue. Deleting and reposing there: #176

dwsinger:
not ok, the TAG affiliation change is resolved at the next election, not immediately, see 2.5.1...

nrooney:
not true for the TAG point - please rewrite.

dwsinger:
much clearer (i think)

frivoal:
> much clearer (i think)

What is? I don't see any update since issues were raised against the initial text.

dwsinger:
still has problems, close without applying?

dwsinger:
@frivoal My mistake. I am concerned we seem more likely to introduce problems here than solve them

dwsinger:
Indeed, this pull Request now appears to say that AB members resolve participation constraints at the next election, like TAG; they don't, AB has to resolve it immediately. Do we really need to make this edit?

frivoal:
I believe that on the last call, we agreed to close this as wontfix. Didn't we?

dwsinger:
I think so.

Issue #171, opened 2018-03-09
href for Candidate Rec link is broken fragment


nigelmegitt:
https://www.w3.org/2018/Process-20180201/#candidate-rec includes a bullet under Possible Next Steps:

> * A revised Candidate Recommendation

however the href in the link for Candidate Recommendation points to `#last-call` which doesn't resolve to anything. I think it should be 

The same issue exists in other places too, like §6.5 Proposed Recommendation where there's the same bullet with broken link. Also §6.6. I'm going to stop hunting now...

PR #187, opened 2018-05-03
Update #last-call to #candidate-rec links

nrooney:
#171

### DIFF
--- a/index.html
+++ b/index.html
@@ -292,7 +292,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
                 <li><a href="#tr-end">6.3.3 Stopping work on a specification</a></li>
               </ul>
             </li>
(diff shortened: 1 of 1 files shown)


dwsinger:
Yes, I think that this is a bad idea, as it break external links (if any)...

chaals:
I'm pretty sure the patent policy still explicitly links to that fragment. Which would be a painful thing to break.

nrooney:
Note to self - add span tag for external ref <span id="last-call"></span>.

jeffjaffe:
@chaals Does this proposed fix when properly done (per Florian) address your concern?

chaals:
@jeffjaffe The concern is that we dont break any link from elsewhere that points to a location within the document using a previously available fragment identifier. Florian describes a couple of ways to do that (there are plenty).

frivoal:
Ok, I've redone that fix, correctly this time. Re-review welcome.

frivoal:
@dwsinger can you have a look, and remove your disapproving review if you agree it is fixed?

PR #188, opened 2018-05-05
written notification of resignation must be recorded

chaals:
* use rfc2119 class for conformance requirements
* say who is responsible for a requirement being met

### DIFF
--- a/index.html
+++ b/index.html
@@ -1154,7 +1154,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
       <h3 id="resignation">3.6 Resignation from a Group</h3>
       <p>A W3C Member or Invited Expert <em class="rfc2119">may</em> resign from a group. On written notification from an Advisory
         Committee representative or Invited Expert to the team, the Member and their representatives or the Invited Expert will
-        be deemed to have resigned from the relevant group. See section 4.2. of the
+        be deemed to have resigned from the relevant group. The written notification MUST be recorded (e.g. by email). See section 4.2. of the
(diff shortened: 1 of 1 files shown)


chaals:
This seems to supersede #184 (thanks to me missing exactly what github was proposing). It was intended to patch that PR to address Dave's comment, and one of my own.

index.html
```diff
@@ -1154,8 +1154,8 @@ <h3 id="WGAppeals">3.5 Appeal of a Chair's Decision</h3>
       <h3 id="resignation">3.6 Resignation from a Group</h3>
       <p>A W3C Member or Invited Expert <em class="rfc2119">may</em> resign from a group. On written notification from an Advisory
         Committee representative or Invited Expert to the team, the Member and their representatives or the Invited Expert will
-        be deemed to have resigned from the relevant group. See section 4.2. of the
-        <a href="https://www.w3.org/Consortium/Patent-Policy" data-ref="PUB33">W3C Patent Policy</a>
+        be deemed to have resigned from the relevant group. The team <em class="rfc2119">must</em> record the notification.
```
frivoal: Should we specify the confidentiality of the notification? Presumably, the resigning member could say very private things in their notification and want it to be treated with various levels of confidentiality. On the other hand, the fact that the member/IE has resigned, once effective, should not itself be more confidential than joining was. I am unsure how much of this needs to be in detailed in the process, but since you're suggesting some normative requirements, we might want to give details.

dwsinger: Yes, you are right, it should be recoded but not necessarily published. "The team must keep a record of the notification, which does not need to be published."

PR #191, opened 2018-05-30
Expand and clarify the 'personal capacity' aspect of AB/TAG participation

kpfleming:
This change separates the 'personal capacity' statements in the AB/TAG Participation section from the statement about delegation of seats, and also spells out the three scenarios for AB/TAG participation via election and appointment and how 'personal capacity' differs from the stated affiliations of participants.

### DIFF
--- a/index.html
+++ b/index.html
@@ -731,13 +731,9 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         are expected to participate regularly and fully. Advisory Board and TAG participants <em class="rfc2119">should</em>
         attend <a href="#ACMeetings">Advisory Committee meetings</a>.</p>
 
-      <p>Individuals elected to the Advisory Board act in their personal capacity on the Board, to serve the needs of the
+      <p>Individuals elected or appointed to the Advisory Board or TAG act in their personal capacity, to serve the needs of the
 	W3C membership as a whole, and the Web community. Whether they are Member representatives or Invited Experts, their
-	activities in those roles are separate and distinct from their activities on the Advisory Board.</p>
-
-      <p>Individuals elected or appointed to the TAG act in their personal capacity in the TAG, to serve the needs of the
-	W3C membership as a whole, and the Web community. If they are Member representatives or Invited Experts, their
-	activities in those roles are separate and distinct from their activities in the TAG.</p>
+	activities in those roles are separate and distinct from their activities on the Advisory Board or TAG.</p>
 
       <p>An individual participates on the Advisory Board or TAG from the moment the individual's term begins
         until the seat is <a href="#AB-TAG-vacated">vacated</a> (e.g. because the term ends). Although Advisory Board and TAG participants do not advocate for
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -731,6 +731,17 @@ <h3 id="AB-TAG-participation">2.5 Advisory Board and Technical Architecture Grou
         are expected to participate regularly and fully. Advisory Board and TAG participants <em class="rfc2119">should</em>
         attend <a href="#ACMeetings">Advisory Committee meetings</a>.</p>
 
+      <p>Individuals elected to the Advisory Board act in their personal capacity on the Board, in spite
+	      of being Member representatives, and in spite of any parallel involvement in other W3C activities as Member
```
chaals: I like the direction of this change. I would prefer wording like "rather than representing the member who nimonated them", and being more explicit about acting in the interests of the Consortium/Web...

chaals:
Actually I think if the text is clear that AB/TAG are not representing an employer or nominating member, but the organisation as a whole, we cover IEs and member employees. Although if you decide it is clearer eplicitly separating the cases, I am not going to argue very hard against it :)

chaals:
This looks good to me. I think you could combine the two paragraphs with "TAG or AB", but I am not sure if it is necessary. (My own preference is to get the document smaller, but that sometimes makes it harder to be clear. I'll leave it to the editor...)

index.html
```diff
@@ -731,6 +731,14 @@ <h3 id="AB-TAG-participation">2.5 Advisory Board and Technical Architecture Grou
...
         attend <a href="#ACMeetings">Advisory Committee meetings</a>.</p>
 
+      <p>Individuals elected to the Advisory Board act in their personal capacity on the Board, to serve the needs of the
+	W3C membership as a whole, and the Web community. Whether they are Member representatives or Invited Experts, their
+	activities in those roles are separate and distinct from their activities on the Advisory Board.</p>
+
+      <p>Individuals elected or appointed to the TAG act in their personal capacity in the TAG, to serve the needs of the
+	W3C membership as a whole, and the Web community. If they are Member representatives or Invited Experts, their
+	activities in those roles are separate and distinct from their activities in the TAG.</p>
+
```
frivoal: These two paragraphs are almost identical. I suggest merging:

> Individuals elected or appointed to the TAG or the Advisory Board act in
> [...]
> distinct from their activities on the TAG or the Advisory Board.

kpfleming:
I tried that, but there are no appointments to the AB, so 'elected or appointed to the TAG or AB' is technically incorrect. To correct that, it would have to be "elected to the AB, or elected or appointed to the TAG", which is cumbersome.

chaals:
@kpfleming 

> there are no appointments to the AB

so "or appointed to' applies to a null set of cases unless we change the way the AB members are determined. However it is still logically true. On that basis I think it's OK to say it...

kpfleming:
Fair enough. Will update the PR later today.

dwsinger:
LGTM

frivoal:
LGTM3

Issue #160, opened 2018-01-29
define "transition request"


samuelweiler:
The term "transition request" should be defined to help those new to W3C. It first appears in section 6.2.2 without any introduction.

nrooney:
Hmmm. Let's have a stab at this:

## Transition Requests
A document progressing through the [W3C Technical Report Development Process](https://w3c.github.io/w3process/#Reports) will need to go through each [maturity level](https://w3c.github.io/w3process/#maturity-levels) on it's way to becoming a W3C Recommendation. To move from one [maturity level](https://w3c.github.io/w3process/#maturity-levels) to the next the Working Group will need to make a **Transition Request**. Each [maturity level](https://w3c.github.io/w3process/#maturity-levels) stage has common and unique **Transition Request** requirements.

jeffjaffe:
Looking at 6.2.2, I believe that Transition request is defined implicitly in the sentence where it is introduced.  Perhaps a small modification (such as using the words "defined as") is a sufficient change.

dwsinger:
agree with Jeff, maybe change
For all Transition Requests, to advance a specification to a new maturity level other than Note,
to
For all requests to transition a specification to a new maturity level other than Note (called "Transition Requests"), 

but this seems moderately pedantic if harmless. we don't have a terms and definitions section.

PR #195, opened 2018-08-08
define "transition request" 

nrooney:
to solve #160

### DIFF
--- a/index.html
+++ b/index.html
@@ -1992,7 +1992,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
 
       <h4 id="transition-reqs">6.2.2 Advancement on the Recommendation Track</h4>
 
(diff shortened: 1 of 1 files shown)


frivoal:
LGTM, although I would think in addition to this, we should convert the whole spec to bikshed and make use of its cross referencing facilities.

jeffjaffe:
The revision takes the word "advance" out of the sentence.  Does that mean it is a transition request even if the WG goes to a lesser level of maturity?

frivoal:
> The revision takes the word "advance" out of the sentence.

That does not seem intentional or desirable. Let's restore it.

dwsinger:
we need to keep 'advance'...

frivoal:
Restored the word advance. I think we're good to (squash and) merge.

This is editorial, so I don't think we need to discuss on the call, and it is OK to merge as soon as there's an approving review.

Issue #86, opened 2017-09-11
Duplication of text


LJWatson:
In 6.1.2 Maturity levels it says "Working Group Note, Interest Group Note (NOTE)". Is the "(Note)" needed?

chaals:
It's the standard shorthand / abbreviation, as CR, REC, etc. Note that "(PR)" is missing, as are formal status for edited, obsolete and superseded Recs. We should fix this one way or another. My preference is to leave them **in** and mark them as `abbr` properly, but I'm not that fussed. Thoughts?

dwsinger:
do we need an acronyms&terms section?

nrooney:
Sent PR. Suggest we close this and open a new issue for acronyms&terms section.

Opened in #198

nrooney:
Going to continue on issue #198

PR #197, opened 2018-08-15
Adds abbr around NOTE in 6.1.2

nrooney:
solves #86

### DIFF
--- a/index.html
+++ b/index.html
@@ -1942,7 +1942,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
            be restored to Recommendation Status. See also clause 10 of the licensing requirements for W3C Recommendations in
           <a href="https://www.w3.org/Consortium/Patent-Policy#sec-Requirements">section 5</a> of the
           <a href="https://www.w3.org/Consortium/Patent-Policy" data-ref="PUB33">W3C Patent Policy</a>.</dd>
-        <dt id="WGNote">Working Group Note, Interest Group Note (NOTE) </dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1942,7 +1942,7 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
            be restored to Recommendation Status. See also clause 10 of the licensing requirements for W3C Recommendations in
           <a href="https://www.w3.org/Consortium/Patent-Policy#sec-Requirements">section 5</a> of the
           <a href="https://www.w3.org/Consortium/Patent-Policy" data-ref="PUB33">W3C Patent Policy</a>.</dd>
-        <dt id="WGNote">Working Group Note, Interest Group Note (NOTE) </dt>
+        <dt id="WGNote">Working Group Note, Interest Group Note (<abbr title="Internet Relay Chat">NOTE</abbr>) </dt>
```
frivoal: `title="Internet Relay Chat"` looks like a copy+paste mistake.

If we want to use abbr, I think we should change the whole line to:
```
<dt id="WGNote"><abbr title="Working Group Note, Interest Group Note">NOTE</abbr>) </dt>
```
and do the same for every other entry in the list. (REC, CR, WD...).

At the same time, leaving this as-is seems fine to me.

What I think we should *really* do is convert the whole document to bikeshed, and use its facilities for defining and cross referencing terms.

dwsinger:
Not accepted, we do this as a general cleanup

nrooney:
Not doing this now - going to continue the full terms and acronyms section discussion in #198

Issue #190, opened 2018-05-14
Increase membership of the AB, possibly TAG


cwilso:
I suggested in the AC meeting in Berlin that we should increase the number of participants in the AB if not also the TAG to account for the increased "diversity" we are garnering with STV.  There seemed to be general support, or at least some support and a lack of antipathy.

I would recommend we increase the number of AB seats by 2 (to 11 elected seats), and the TAG by 1 (to three appointed, 7 elected seats).

nrooney:
I actually support this - there's a lot of work to do and more people will help.

dwsinger:
As long as we feel we can find the candidates, I am OK with this. The downsides to larger groups is the sense that in a small group you can't be a passenger; and finding the candidates.

bkardell:
It feels like the discussion at https://lists.w3.org/Archives/Member/w3c-ac-forum/2018AprJun/0009.html is very relevant to this discussion for two reasons: 1) If any restructuring happens as in there, it seems relevant to reconsider the whole idea of appointed seats... They seem to cease to make any sense in that model.  2) If before that, even, it seems like these bodies need more seats.... well... that seems relevant to how restructuring should happen. 

I feel more the record pretty much like @dwsinger expressed: Making this group too big is going to be counter productive, but I don't know what 'right size' is either - that's a comment I've made a number of times, and a big part of that is related to being able to find good candidates. Just filling slots to fill them is no good, and neither is turning away really great options because there just aren't enough seats.

jeffjaffe:
The Advisory Board discussed this issue on 16 May 2018 and there was a consensus of support for Chris' proposal.

The AB also noted that it is important to get the TAG's input on the TAG part of the proposal.

dwsinger:
If we have continued anxiety about finding candidates, we could say that each body must have at least N and no more than M. With a larger body, we might have to consider what to do if we have too few nominations.

frivoal:
If we do an "at least N, no more than M" system, we need to be clear about what this means with STV (assuming we stick to that). If the number of candidates falls between the two numbers, is everybody elected automatically, or is there some possible result of the election that can reject candidates for not being popular enough (as long as we still have at least N)?

I have been involved in non profit board elections with an "at least N, no more than M" system using approval voting, and in that case, how it works is obvious and non controversial.

michaelchampion:
[...]

PR #202, opened 2018-08-29
Increase the size of the AB.

cwilso:
Fixes Issue #190, as per consensus of the AB on 16 May 2018.

### DIFF
--- a/index.html
+++ b/index.html
@@ -642,16 +642,16 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
 
       <h4 id="ABParticipation">2.3.1 Advisory Board Participation</h4>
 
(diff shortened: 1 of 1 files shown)


dwsinger:
If we agree to "at least M and not more than N", we would replace "eleven" with "at least nine and not more than eleven" and adjust the staggering sentence be "such that half the terms expire" (and leave to the team the management of that).

frivoal:
The proposed patch does match what I understand to be the AB's resolution.

I am a little bit skeptical of an overly large group's efficiency, but probably not skeptical enough to stand against the consensus.

As for I've "at least M and not more than N", I've had good experiences with that on another organization's board (we had M=6 and N=9, also with a 2 year term with staggered sentence to renew half every year). However, I do think it implies a change in voting system, as I don't think meek STV is defined under these conditions.

dwsinger:
approved in the CG call

PR #203, opened 2018-09-11
define affiliation as in #176

dwsinger:


dwsinger:
approved in the CG call

PR #205, opened 2018-09-13
Fix markup issues

frivoal:
This resolves errors and warnings from https://validator.w3.org/nu/

Doing this as a pull request for now, but if this is OK with the chair (hello @dwsinger), going forward I can do this kind of non substantive changes as a direct commit without going through a Pull Request. Or I can continue doing PRs for everything. As the group / chair wants.


LJWatson:
FWIW, I agree that editorial and/or markup improvement changes should be done through direct commit. There should be no need to review spelling corrections and markup corrections.

PR #206, opened 2018-09-13
Update metadata

frivoal:
List of editors, date, copyright years, version of the official process this is
based on…

Also mark as an issue the fact that the Change section is stale.


frivoal:
@dwsinger same question as in #205: do I need to send this sort of change through PRs for review, or can I just commit that directly?

PR #210, opened 2018-09-14
editorial clarification

wseltzer:
remove an extraneous "both" to make the language re Essential Claims make sense


dwsinger:
It's not extraneous. It's a both-and construction (both A and B). And it's important. We would like a grant for BOTH
* your Essential Claims in your own contribution (even if they were essential in the draft already)
AND
* things that become your Essential Claims as a result of incorporating your contribution (this covers things like combination patents).

wseltzer:
Yes, but "both" is misplaced such that the sentence is un-parseable as written.

dwsinger:
Ah, actually it's "those" that is extraneous

Such commitment should cover, at a minimum, all the contributor’s Essential Claims both in the contribution, and those that become the contributor's Essential Claims as a result of incorporating the contribution into the draft that existed at the time of the contribution

should be

Such commitment should cover, at a minimum, all the contributor’s 
Essential Claims both 
in the contribution, and 
that become the contributor's Essential Claims as a result of incorporating the contribution into the draft that existed at the time of the contribution

wseltzer:
ok, PR updated per @dwsinger's suggestion

PR #214, opened 2018-10-02
Clarify what the expectations are for advancing to CR

### LINKED ISSUE #172: Does a spec need to be Rec-ready to be a CR? Align defn with process

frivoal:
Closes #172

### DIFF
--- a/index.html
+++ b/index.html
@@ -1964,7 +1964,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           abandoned without producing a Recommendation.</dd>
       </dl>
 
-      <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
+      <p>Only sufficiently technically mature work should be advanced.
         <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired,
         this can be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
         less stable features to other technical reports.</p>
--- a/index.html
+++ b/index.html
@@ -1917,11 +1917,11 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           </ul>
         </dd>
         <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates either that
-          the technical report is expected to advance to Recommendation without substantive changes,
-          or that no further improvement is deemed possible without additional implementation experience or feedback from the wider community.
-          Announcement of a
-          different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late
-          a stage.</dd>
+          no further improvement is expected without additional implementation experience or feedback from the wider community.
+          A Candidate Recommendation is expected to be of the quality of a Recommendation itself,
+          and acceptable as such if and when the requirements for further advancement are met.
+          Announcement of a different next step <em class="rfc2119">should</em> include
+          the reasons why the change in expectations comes at so late a stage.</dd>
         <dt id="RecsPR">Proposed Recommendation</dt>
         <dd>A Proposed Recommendation is a document that has been accepted by the W3C Director as of sufficient quality to become
           a W3C Recommendation. This phase     triggers formal review by the Advisory Committee, who <em class="rfc2119">may</em> recommend that the document be published as a W3C Recommendation, returned to the Working Group for further work, or abandoned. Substantive changes <em class="rfc2119">must not</em> be made to a Proposed Recommendation
--- a/index.html
+++ b/index.html
@@ -1966,7 +1966,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
 
       <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
         <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired,
-        this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
+        this can be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
         less stable features to other technical reports.</p>
 
 
--- a/index.html
+++ b/index.html
@@ -1917,8 +1917,8 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           </ul>
         </dd>
         <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates that
-          no further improvement is expected without additional implementation experience or feedback from the wider community.
-          A Candidate Recommendation is expected to be of the quality of a Recommendation itself,
+          no further improvement is expected without additional implementation experience and testing.
+          A Candidate Recommendation is expected to be as well-written, detailed, self-consistent, and technically complete as a Recommendation,
           and acceptable as such if and when the requirements for further advancement are met.
           Announcement of a different next step <em class="rfc2119">should</em> include
           the reasons why the change in expectations comes at so late a stage.</dd>
@@ -1965,7 +1965,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
       </dl>
 
       <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
-        <strong>Note:</strong> should faster advancement to meet scheduling considerations be desired,
+        <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired,
         this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
         less stable features to other technical reports.</p>
 
--- a/index.html
+++ b/index.html
@@ -1916,7 +1916,10 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates either that
+          the technical report is expected to advance to Recommendation without substantive changes,
+          or that no further improvement is deemed possible without additional implementation experience or feedback from the wider community.
+          Announcement of a
           different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late
           a stage.</dd>
         <dt id="RecsPR">Proposed Recommendation</dt>
@@ -1960,6 +1963,12 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           a stable reference for a useful document that is not intended to be a formal standard, or to document work that was
           abandoned without producing a Recommendation.</dd>
       </dl>
+
+      <p>Advancement to a new maturity level must be made based on the technical maturity of the work
+        rather than on scheduling considerations.
+        If faster advancement is desired,
+        this may be achieved by reducing the scope of the technical report to the more stable subset
+        and by postponing less stable features to separate technical reports.</p>
 
       <p>Working Groups and Interest Groups <em class="rfc2119">may</em> make available "Editor's drafts". Editor's drafts have
         no official standing whatsoever, and do not necessarily imply consensus of a Working Group or Interest Group, nor are
--- a/index.html
+++ b/index.html
@@ -1964,11 +1964,12 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           abandoned without producing a Recommendation.</dd>
       </dl>
 
-      <p>Advancement to a new maturity level must be made based on the technical maturity of the work
-        rather than on scheduling considerations.
-        If faster advancement is desired,
-        this may be achieved by reducing the scope of the technical report to the more stable subset
-        and by postponing less stable features to separate technical reports.</p>
+      <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
+        <strong>Note:</strong> should faster advancement to meet scheduling considerations be desired,
+        this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
+        less stable features to other technical reports.</p>
+
+
 
       <p>Working Groups and Interest Groups <em class="rfc2119">may</em> make available "Editor's drafts". Editor's drafts have
         no official standing whatsoever, and do not necessarily imply consensus of a Working Group or Interest Group, nor are
--- a/index.html
+++ b/index.html
@@ -1916,7 +1916,7 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates either that
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates that
           no further improvement is expected without additional implementation experience or feedback from the wider community.
           A Candidate Recommendation is expected to be of the quality of a Recommendation itself,
           and acceptable as such if and when the requirements for further advancement are met.


index.html
```diff
@@ -1960,6 +1963,12 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
           abandoned without producing a Recommendation.</dd>
       </dl>
 
+      <p>Advancement to a new maturity level must be made based on the technical maturity of the work
```
nigelmegitt: Seems like we're picking on "scheduling considerations" to the exclusion of other potential motivations for pressing ahead. Suggest rewording to:

> The decision to advance to a new maturity level must be made solely based on the technical maturity of the work. 
> Note: should faster advancement to meet scheduling considerations be desired, this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
> less stable features to other technical reports.

frivoal: I like your suggestion. Applying it.

index.html
```diff
@@ -1915,7 +1915,10 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates either that
+          the technical report is expected to advance to Recommendation without substantive changes,
+          or that no further improvement is deemed possible without additional implementation experience or feedback from the wider community.
```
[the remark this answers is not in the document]
frivoal: Hmmm. I'd like to leave the door open to "let's leave this undefined in this level", but not to "this is known to be wrong, but whatever, let's ship anyway".

I think my phrasing may be a bit too strict, and yours a bit too loose.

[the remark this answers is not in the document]
nigelmegitt: I agree with allowing "let's leave this undefined" being a reasonable deliberate choice that should not in itself block transition (but could be a blocker if there is unresolved feedback that raises it as an issue, for example).

Happy to review other wording proposals to try to hit the sweet spot.

dwsinger:
I'd rather not lose the sense of "Candidate Recommendations are expected to be acceptable as Recommendations", and I wonder if we should have "The text of a Candidate Recommendation is expected to be of the quality of, and acceptable as, the text of a Recommendation itself." or the like?

frivoal:
I understand what you're trying to achieve with that phrasing, but I am not sure that it works.
* We do not actually have a requirement that the prose of RECs be high quality prose, separately from the technical requirements. I don't know how to require than other than observing after wide review + an AC vote that nobody is complaining that the spec is crap. And I think we already have that. So maybe we don't need anything.
* Saying it's about "the text" sounds it might exclude non textual things like figures, which is certainly not what you intend.

Assuming you don't want to just drop this, how about that:
> A Candidate Recommendation is expected to be of the quality of a Recommendation itself, and acceptable as such if and when the requirements for further advancement are met.

nigelmegitt:
> "The text of a Candidate Recommendation is expected to be of the quality of, and acceptable as, the text of a Recommendation itself."

@dwsinger the test for this is that it passes the Director's, and the AC's review, right, because those are the reviews that happen to get from CR to PR and then Rec? The Director reviews the transition to CR, but the AC isn't likely to comment until PR. We either need to close that loop so that your goal can be verified before transitioning to CR, or open up the very real possibility of getting feedback along the lines "this should never have been a CR" _after_ the spec is in CR and has already met its CR exit criteria. The process needs to allow for the right feedback to happen at the right time, or accept that such feedback cannot be meaningfully processed. What would you do here?

frivoal:
To get accepted as a CR, the spec needs not only to pass the Director's personal review, but also to demonstrate wide review. That seems like a perfectly appropriate time for anyone to point out if a spec is such a mess that it shouldn't be allowed to go forward.

nigelmegitt:
@frivoal right, so in that case does adding this statement of expectation help anyone? It seems like it's more an instruction to reviewers on the level of quality they should be aiming for, rather than a part of the process per se.

dwsinger:
I think we need, at least, to set an expectation, that the team and others can point at: if, magically, implementation experience and interop and all the things that happen post-CR go amazingly swimmingly, you should basically have the PR/Rec. text in hand. The spec needs to be of specification-quality, basically, or as close as it can get without the CR-period experience. And we need to say that that is the expectation.

dwsinger:
> @frivoal right, so in that case does adding this statement of expectation help anyone? It seems like it's more an instruction to reviewers on the level of quality they should be aiming for, rather than a part of the process per se.

In a sense, yes. reviewers should be able to say "nope, your document is not spec.-ready and it's required, and you knew it was required because it's written in the process". I think if it's unstated we'll get cases trying to slip by, and slip by they will.

frivoal:
@dwsinger your comments seem to be in support of the phrasing I proposed in https://github.com/w3c/w3process/pull/214#issuecomment-426866521, but you don't say so explicitly. Does that work for you, or should we try some more?

frivoal:
@dwsinger , @nigelmegitt I've updated the PR with a blend of the phrases we have been discussing. I hope the latest version covers your expectations.

To make it easier to read than looking at a diff, here are the two key paragraphs (do look at the diff if you want to see them in context)

>  <strong>Note</strong>: Advancing to Candidate Recommendation indicates either that no further improvement is expected without additional implementation experience or feedback from the wider community. A Candidate Recommendation is expected to be of the quality of a Recommendation itself, and acceptable as such if and when the requirements for further advancement are met. Announcement of a different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late a stage.

> The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
<strong>Note:</strong> should faster advancement to meet scheduling considerations be desired, this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring less stable features to other technical reports.

dwsinger:
This is fine.  Thanks for persisting

frivoal:
@dwsinger Should I just merge? Do you want to issue a CfC?

index.html
```diff
@@ -1915,9 +1915,12 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a
-          different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late
-          a stage.</dd>
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates either that
```
nigelmegitt: The word "either" needs to be removed so this sentence makes grammatical sense.

frivoal:
> Looks good to me after the grammar tweak I've proposed is made.

Done.

index.html
frivoal: oops. That's a leftover from earlier phrasings. Deleting.

fantasai:
Wrt
```
A Candidate Recommendation is expected to be of the quality of a Recommendation itself,
and acceptable as such if and when the requirements for further advancement are met.
```

I really liked @dwsinger’s phrasing in [this comment](https://github.com/w3c/w3process/issues/172#issuecomment-420732732):
> The point of the CR transition is to ask for implementations, implementer feedback, testing, and so on. In the miraculous case that this all goes 100% perfectly, the text of the document should be good enough to publish as a Recommendation. That's what I think that the Process requires, and should be required. I think it irresponsible to ask people to implement and give feedback on something that is not well enough written, consistent, or technically complete enough to publish (and hence, not enough to implement thoroughly either). So, I think the sentence should stand.

So maybe something like “A Candidate Recommendation is expected to be as well-written, detailed, self-consistent, and technically complete as a Recommendation, and acceptable as such ...”?

index.html
```diff
@@ -1915,9 +1915,12 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a
-          different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late
-          a stage.</dd>
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates that
+          no further improvement is expected without additional implementation experience or feedback from the wider community.
```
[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

index.html
```diff
@@ -1960,6 +1963,13 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
           abandoned without producing a Recommendation.</dd>
       </dl>
 
+      <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
+        <strong>Note:</strong> should faster advancement to meet scheduling considerations be desired,
```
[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

index.html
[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
nigelmegitt: My reading of the Process is that wide review feedback is welcome during CR, and is not restricted to WD prior to CR; it is true that wide review must be demonstrated to enter CR, but that doesn't restrict the opportunity for providing review comments.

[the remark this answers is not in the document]
fantasai: Right. But this is talking about why we transition to CR. We don't do it to get wide review, we do it because we need implementation experience.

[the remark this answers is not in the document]
nigelmegitt: OK, but feedback may need to be addressed whether it is implementation-related or not. For an example, if since entering CR some law were passed that had implications for a feature in some country, the WG should probably have a discussion about that.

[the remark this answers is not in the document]
fantasai: Of course. But this sentence is explaining what distinguishes a CR from a WD, not explaining things that are common to both states.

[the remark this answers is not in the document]
frivoal: @fantasai I took most of your suggested improvements, except for `s/is expected/can be expected/`, as I believe this risks forbidding leaving things intentionally undefined.

index.html
```diff
@@ -1916,9 +1916,12 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
               corresponds to the "Last Call Working Draft" discussed in the Patent Policy.</li>
           </ul>
         </dd>
-        <dd><strong>Note</strong>: Candidate Recommendations are expected to be acceptable as Recommendations. Announcement of a
-          different next step <em class="rfc2119">should</em> include the reasons why the change in expectations comes at so late
-          a stage.</dd>
+        <dd><strong>Note</strong>: Advancing to Candidate Recommendation indicates that
+          no further improvement is expected without additional implementation experience and testing.
```
nigelmegitt: No further _substantive_ improvement?

WGs typically re-read their specs and make editorial improvements, and that should fall within the range of what is expected.

frivoal: Right, I was wondering about something like this.

The flip side is that just because editorial improvements are allowed doesn't mean that they are expected, and since we expect CRs to be acceptable as RECs, we should mostly not enter CR when things are unreadable and need massive editorial improvements.

Then again, if we adopt the suggestion you propose, combining it with the rest of the text, it means that we're supposed to enter CR with high quality prose that would be acceptable as a REC, but that this is not incompatible with having further plans for editorial improvements, which sounds reasonable to me.

I'll ponder that for a short white, but unless someone pushes back, I'll probably make this change.

dwsinger: I think the key word here is *expected*. No, we don't want CRs that have known editorial issues either. On the other hand, this sentence is clearly about the technical content. No technical improvement is expected without...?

index.html
```diff
@@ -1961,6 +1964,13 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
           abandoned without producing a Recommendation.</dd>
       </dl>
 
+      <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
+        <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired,
+        this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
+        less stable features to other technical reports.</p>
+
+
+
```
wseltzer: Disagree with this addition, as it reflects hope over realpolitik. A group may rationally and sensibly opt to bundle features and hold back "technically mature" features in order to offer a synchronized package of mature features.

wseltzer: NB: that disagreement refers to the whole insertion, " <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work. 
        <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired, this may be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring less stable features to other technical reports.</p>"

dwsinger: I don't understand what you're saying. (a) you think groups should push immature features to CR? or (b) they should hold back and wait until all they stuff they wants is mature (which doesn't conflict with this statement) or (c) something else?

nigelmegitt: I am also confused: I don't understand how your point @wseltzer differs from the proposed text. It seems that we are all in agreement that, in the case that quicker advancement is desired, since advancement is based solely on the technical maturity, then insufficiently mature features would need to be dropped to achieve that desire. If alternatively the desire is to ship with those features, then the converse is that the group must wait until they are sufficiently mature. Scope and timing are both within the control of the working group, but may work against each other. Does the wording need changing to make it even clearer that this is what the note says?

wseltzer: Correct, @nigelmegitt , and in the case the group decides that bundling is more important than speed, they're making decision (not) to advance based on factors other than solely technical maturity.

dwsinger: We don't say anything about why a group might NOT advance; it might be based on chair incompetence, sheer stupidity, utter laziness, or a host of factors. We only set requirements on the decision TO advance.

wseltzer: Nonetheless, I find the addition unduly constraining.

dwsinger:
Ok. Since it is only a NOTE let’s remove it in the interests of making the progress we can

index.html
frivoal: I just don't get it:
* It is a note, so it is not constraining
* it says that the group has a choice between either waiting for everything to be ready and advancing together, or advancing some parts first and the rest later if fast advancement is desired and not everything is ready. Disagreeing with that seems to be calling for advancing things that are not ready, which I strongly object to.

(I don't object to removing the note for the sake of making progress, since it is indeed just a note, unless the intent is to allow advancing things that are not ready).

frivoal: Maybe I am just dense, but what alternative is there other than:
1. advancing parts that are ready first, other parts later
2. waiting until everything is ready, then advancing everything
3. advancing things that are not ready.

The note says: ”if you don't want to do 2, you can do 1“. Unless you're claiming that 3 is a legitimate thing to do, I just don't understand the pushback.

fantasai: Suggest s/may/can/ in the note, but otherwise agreed with frivoal/nigel/dsinger : unless you're arguing for #3 (advancing things that are not ready in pursuit of preconceived deadlines) I don't understand the objection.

frivoal: > Suggest s/may/can/ in the note

Good suggestion, which I don't expect to change the opinion of anyone already in favor of this, so I've updated the text accordingly: https://github.com/w3c/w3process/pull/214/commits/f5e96b21e43d97cb176974c88dc3cfdb14dcc3f5

swickr:
@dwsinger commented a while back

> reviewers should be able to say "nope, your document is not spec.-ready and it's required, and you knew it was required because it's written in the process". I think if it's unstated we'll get cases trying to slip by, and slip by they will.

This suggests a pandora's box of subjectiveness that I'm loathe to open in the text of the Process.   By the time a WG feels their spec is sufficiently mature to become a CR all reviewers with concerns should be expected to propose -- or at least outline -- changes that would resolve their concern.   "Not spec-ready" is only a useful comment when accompanied by sufficient text for the WG to understand what changes are being asked of it.   [6.2.2](https://w3c.github.io/w3process/#transition-reqs) and [6.4](https://w3c.github.io/w3process/#candidate-rec) hold the normative prose for reviewers to make such comments and have them considered during transition deliberations.

[6.1.2](https://w3c.github.io/w3process/#maturity-levels) is the wrong place to be providing advice to Working Groups, their editors, and reviewers.

I don't support the proposed additions to 6.1.2; I only support dropping the [Note as an historic artifact](https://github.com/w3c/w3process/issues/172#issuecomment-438511401) as it apparently now causes confusion.    Additional informative advice such as that proposed could usefully be incorporated into [Working Group Effectiveness](https://github.com/w3c/wg-effectiveness).

index.html
```diff
@@ -1961,6 +1964,13 @@ <h4 id="maturity-levels">6.1.2 Maturity Levels</h4>
           abandoned without producing a Recommendation.</dd>
       </dl>
 
+      <p>The decision to advance to a new maturity level must be made solely based on the technical maturity of the work.
+        <strong>Note:</strong> Should faster advancement to meet scheduling considerations be desired,
```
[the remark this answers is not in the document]
frivoal: I'm relatively indifferent between the existing phrasing and the alternative one proposed by @nigelmegitt . As for the one suggested by @chaals, would not object to it, but I like it less. Sure, it's shorted, but it hides the fact that advancing faster is a matter of choice.

nigelmegitt:
Removing the note altogether was one of the proposals I suggested in #172, so it would satisfy the issue as far as I'm concerned. I actually think the modification as proposed in this pull request is better than that, but complete removal would also be acceptable to me.

dwsinger:
> 
> This suggests a pandora's box of subjectiveness that I'm loathe to open in the text of the Process. By the time a WG feels their spec is sufficiently mature to become a CR all reviewers with concerns should be expected to propose -- or at least outline -- changes that would resolve their concern. 

Agreed.  A claim that something is not spec.-ready would need to be supported by evidence (for example, the existence of a TBD in the document, or editorial notes indicating uncertainty, and so on).

>"Not spec-ready" is only a useful comment when accompanied by sufficient text for the WG to understand what changes are being asked of it. [6.2.2](https://w3c.github.io/w3process/#transition-reqs) and [6.4](https://w3c.github.io/w3process/#candidate-rec) hold the normative prose for reviewers to make such comments and have them considered during transition deliberations.

If you want to move the requirements and advice to 6.4, which is also about Candidate Recommendation, I am fine with that.

> I don't support the proposed additions to 6.1.2; I only support dropping the [Note as an historic artifact](https://github.com/w3c/w3process/issues/172#issuecomment-438511401) as it apparently now causes confusion. Additional informative advice such as that proposed could usefully be incorporated into [Working Group Effectiveness](https://github.com/w3c/wg-effectiveness).

So, you think that we should return to the old days where groups issued Last Calls (now CRs) on documents that could not be considered as publishable? I don't, and I don't understand why you would.

swickr:
> If you want to move the requirements and advice to 6.4, which is also about Candidate Recommendation, I am fine with that.

I want any new proposed normative requirements to be in 6.4 and I recommend that advice be in a separate document.  Statements such as "The decision to advance to a new maturity level must be made solely based on the technical maturity of the work" are claimed in this thread to not be constraining but have the same type of Process consistency issue that opened #172.

> So, you think that we should return to the old days where groups issued Last Calls (now CRs) on documents that could not be considered as publishable?

No, certainly not.

dwsinger:
> > "The decision to advance to a new maturity level must be made solely based on the technical maturity of the work."
> 
> is over-constraining as @wseltzer previously objected.

You seem to understand something from Wendy here that no-one else does, so can you explain? You (plurally) want to be able to advance technically *immature* work?

wseltzer:
No, the decision to -- or not to -- advance may be made on other considerations. For example, the group may decide *not* to advance work that is technically mature, in order to synchronize it with other work that is not yet mature.

wseltzer:
"Only sufficiently technically mature work should be advanced," is different in meaning from "The decision to advance to a new maturity level must be made solely based on the technical maturity of the work." I'd find the first acceptable, not the second.

dwsinger:
The original statement says nothing at all about why you make a decision *not to* advance, only about the criteria *to* advance. Nonetheless, if you can re-phrase to something which has the same intended meaning, I expect to get consensus.

wseltzer:
@dwsinger clearly our interpretations of the phrase and what's said by implication differ.  I suggest "Only sufficiently technically mature work should be advanced." Otherwise, I suggest we leave it out entirely.

dwsinger:
try again. 

@swickr @wseltzer  OK, 

retain:

"A Candidate Recommendation is expected to be as well-written, detailed, self-consistent, and technically complete as a Recommendation, and acceptable as such if and when the requirements for further advancement are met."

replace 
"The decision to advance to a new maturity level must be made solely based on the technical maturity of the work." 
with
 "Only sufficiently technically mature work should be advanced."

Move to guide:

"Note: Should faster advancement to meet scheduling considerations be desired, this can be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring less stable features to other technical reports." is advice that states the somewhat-obvious and could be in the guide.

Omit or move to guide, or leave as note with the retentions:

"Note: Advancing to Candidate Recommendation indicates that no further improvement is expected without additional implementation experience and testing." 

The retentions could be in 6.4, which is about CR, or left in the section that the original phrase is in.

frivoal:
@dwsinger 's proposal in https://github.com/w3c/w3process/pull/214#issuecomment-438871242 is acceptable to me. I'd prefer keeping the notes in instead of moving then the the guide or dropping, as I think such notes improve the understandability of a document more than terseness does, but if there is an actual objection to having them, and no objection if they're removed, I can live with that.

wseltzer:
+1 @dwsinger's https://github.com/w3c/w3process/pull/214#issuecomment-438871242

swickr:
The [rephrasing](https://github.com/w3c/w3process/pull/214#issuecomment-438871242) addresses my core objection and with that change I accede to a consensus for including this advisory material in the Process rather than in /Guide.

fantasai:
Looks good to me, too.

frivoal:

PR #215, opened 2018-10-02
Clarify maturity requirements for TR updates at the same maturity

### LINKED ISSUE #207: What are the requirements to update a CR?
cloned from #172  https://github.com/w3c/w3process/issues/172#issuecomment-413196028

Do we need to describe what conditions need to met to *update* a CR with a 'second CR'? One school says "same as entering CR", but another points out that maybe the CR has more than one problem, identified since CR entry, and perhaps it should be legitimate to update the CR to fix some but not all.

Or maybe that's what WG working drafts are for?

frivoal:
Closes #207

This PR is is build on top of https://github.com/w3c/w3process/pull/214, which is expected to be merged first.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1994,7 +1994,13 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         this can be achieved by reducing the scope of the technical report to a subset that is adequately mature and deferring
         less stable features to other technical reports.</p>
 
-
+      <p>When publishing an updated version of an existing Candidate Commendation or Recommendation,
+        technical reports are expected to meet the same maturity criteria as when they are first published under that status.
+        However, in the interest of replacing stale documents with improved ones in a timely manner,
+        if flaws have been discovered in the technical report after its initial publication as a CR or REC
+        that would have been severe enough to reject that publication had they be known in time,
+        it is also permissible to publish an updated CR or REC following the usual process,
+        even if only some of these flaws have been satisfactorily addressed.</p>
 
       <p>Working Groups and Interest Groups <em class="rfc2119">may</em> make available "Editor's drafts". Editor's drafts have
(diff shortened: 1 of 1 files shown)


dwsinger:
hm. We lost "Candidate Recommendations are expected to be acceptable as Recommendations", and I wonder if we should have "The text of a Candidate Recommendation is expected to be of the quality of, and acceptable as, the text of a Recommendation itself." or the like?

frivoal:
@dwsinger I think you comment above is meant for #214, rather than this PR which merely builds on top of it. you may want to delete it and repost it there. (I'll remove this comment as well if you do so)

dwsinger:
> @dwsinger I think you comment above is meant for #214, rather than this PR which merely builds on top of it. you may want to delete it and repost it there. (I'll remove this comment as well if you do so)

Happy to move it, but the comment was specifically about the loss of the sense behind "Candidate Recommendations are expected to be acceptable as Recommendations."

frivoal:
yes, but that sentence was removed by #214, not by this patch, which just starts from that branch and then adds more.

dwsinger:
OK, this PR adds only the advice about updating at the same maturity level, with the rest of the changes handled in #214 

"When publishing an updated version of an existing Candidate Commendation or Recommendation, technical reports are expected to meet the same maturity criteria as when they are first published under that status. However, in the interest of replacing stale documents with improved ones in a timely manner,  if flaws have been discovered in the technical report after its initial publication as a CR or REC that would have been severe enough to reject that publication had they be known in time,  it is also permissible to publish an updated CR or REC following the usual process, even if only some of these flaws have been satisfactorily addressed."

frivoal:
Rebased on top of the updated changes of #214.

frivoal:

PR #216, opened 2018-10-02
The number of elected AB member may vary between 9 and 11

### LINKED ISSUE #204: Should the AB and maybe TAG have lower/upper bounds for membership?
At the moment, the TAG and AB have defined participation numbers. (See #190 for the most recent change).

In the past, we had unopposed elections when the numbers were smaller. With an increased size, we run the risk of (in quieter times, if they ever happen again) of having trouble filling the larger committee. This can be alleviated by defining that these committees have "at least N and no more than M". In a sense, today, N=M.

I believe that the handling of this is unaffected by the voting system, though some have expressed concern to the contrary. 

I believe that the following apply under any voting system, and whether M=N or M>N.

1. At any election, if there are fewer nominations than needed to maintain the committee at at least N, then continue to do whatever we do now (I think our process is that we delay until enough nominations are received).
2. If the election system does not permit rejecting candidates, and the number of nominations is less than the number of free seats, then the candidates are elected unopposed (the resulting committee has <=M occupied seats).
3. If the number of nominations is greater than the number of free seats, then an election is held to appoint to fill exactly M seats.
4. If the election system were to permit rejecting candidates, then an election is always held. If the result is that fewer than N seats become occupied (too many rejected candidates), then the introduction of that voting system will have to describe what happens (the election is void and nominations re-opened? we open a second round for the unfilled seats?). We do not have, and have not had in the past, such an election system, so this should be dealt with if we ever introduce such.

Effectively at each election, we look at the number of seats not open for re-election, and subtract that from N and M (e.g. if N=9 and M=11 and 5 seats are not open for re-election, then we say we are looking to elect between 4 and 6 people).

frivoal:
Closes #204

### DIFF
--- a/index.html
+++ b/index.html
@@ -661,7 +661,9 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
 
       <p>The number of seats open for election each year is determined as follows:
       <ul>
-        <li><var>c</var> is the number of elected Advisory Board members whose term has not yet ended and will continue to serve on the AB.
+        <li><var>c</var> is the number of elected Advisory Board members whose term has not yet ended,
+          including both those who will continue to serve on the AB to the completion of their term
+          and seats vacated before the end of the term.
         <li><var>i</var> is the number of incomplete terms to be filled due to vacated seats.
         <li><var>m</var> is the result of <code>9 &minus; <var>c</var></code>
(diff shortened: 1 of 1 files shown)


frivoal:
Closing in favor of https://github.com/w3c/w3process/pull/224

Issue #28, opened 2017-04-21
Substantive changes undefined for Charter, Process reviews, and things other than specifications


dwsinger:
Transferred from https://www.w3.org/community/w3process/track/issues/176
State: Raised

dwsinger:
we need text proposals, please. editors to check that links correctly identify where terms are defined

dwsinger:
Perhaps this needs defining in the possible Terms and Definitions section?

frivoal:
I think this issue has actually several sub parts:
1. "Substantive change" and "editorial change" are only defined for specs, not for charters
2. The usage of the term "editorial change" is inconsistent, and some place contrast Substantive changes to "minor changes" instead of "editorial changes".
3. Daniel objects to step 2 of of 7.1.2 which allows the Director to make Substantive changes to charters or specs without returning the proposal for additional work.

I have made a pull request to solve 1 and 2: https://github.com/w3c/w3process/pull/217. I expect we might quibble on the details, but probably not on the principle.

As for 3, I first thought that this was best addressed as part of #182, but actually, there is a much simpler solution, which I have proposed in a separate pull request (https://github.com/w3c/w3process/pull/218): simply delete step 2 of 7.1.2. Making any substantive change to a document after the AC has voted on it pretty much invalidates the vote, as there is not reliable way of knowing how people would have voted had that change been included originally. We can (and should) go into further details as part of #182, but for now, just removing this possibility seems a good idea.

frivoal:
update: I've completed the https://github.com/w3c/w3process/pull/217 PR to define editorial vs substantive for the process in addition to for charters.

dwsinger:
I agree that changes to scope or deliverables are the very essence of substantive.  Dates are almost certainly so as well.

Is there room here for "if anyone thinks it's substantive, it's substantive"? That is, for things in the grey areas, the Director writes "I propose doing X and I think it's not substantive; if anyone disagrees, say so, and we'll put the revised text to formal review".

As an aside, I'd like to keep the text as short as possible, of course, and give us as much latitude as possible...

frivoal:
I'm starting to think that splitting this PR into two may not have been a good idea after all, as the question of what is or isn't substantive is not that detached from what the Director is or isn't allowed to change by himself.

[...]

PR #217, opened 2018-10-02
Defines substantive and editorial changes for charters

frivoal:
These were previously only defined for specs. Also, be consistent in the
use of editorial vs minor changes.

Part of #28

### DIFF
--- a/index.html
+++ b/index.html
@@ -2099,6 +2099,13 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         considered <dfn id="editorial-change">editorial changes</dfn>, the latter two
         <dfn id="substantive-change">substantive changes</dfn>.</p>
 
+      <p>When discussing changes to the W3C process,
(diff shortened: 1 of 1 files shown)


frivoal:
@wseltzer you mean that the director should be allowed to add a deliverable to a charter, after teh AC has voted, without saying why he is doing that (which is allowed for editorial comments but not for substantive ones)?

wseltzer:
For some charters, the scope of activity is defined substantively, and the charter permits a group to add any in-scope deliverable at any time. In such a case, if a commenter pointed out an in-scope document that had group input but hadn't been listed as a deliverable, I'd support adding it. To be clear, I also support sending every change back to all the reviewers to verify that their interpretation matches the Director's proposal -- but not baking that into Process.

frivoal:
This pull request predates that.

I could rebase it, but I intend to send a different pull request later to address this issue anyway, so closing instead.

PR #218, opened 2018-10-02
Require a return for additional work for substantive changes 

frivoal:
Part of #28


frivoal:
This pull request predates that.

I could rebase it, but I intend to send a different pull request later to address this issue anyway, so closing instead.

PR #220, opened 2018-10-03
Clarify that "at most one" participant in AB/TAG is only about elected / appointed seats

### LINKED ISSUE #208: Removal of "Member" from 2.5.1: unintended consequences
The removal of "Member" from 2.5.1 bullets appears to prohibit W3C from having a chair/participant and a Team Contact: "An organization is permitted at most one participant on the AB" (and TAG, with term consideration). Should the text read "An organization other than W3C..."?

frivoal:
Closes #208

### DIFF
--- a/index.html
+++ b/index.html
@@ -757,11 +757,11 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         W3C Members is represented:</p>
 
       <ul>
-        <li>An organization is permitted at most one participant on the TAG except when having more than one participant
+        <li>An organization is permitted to hold at most one seat on the TAG except when having more than one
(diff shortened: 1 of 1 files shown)


dwsinger:
LGTM

chaals:
(To be clear, I disagree with the proposal to make the limitation apply to appointed seats on the TAG - although I won't block CG consensus to put it in our proposal to the AC)

frivoal:
@chaals that's not really a change though, this was already the case prior to the change that this PR is cleaning up: https://github.com/w3c/w3process/commit/e1788267a674a33dedcb9e9b951edca0e2f05ad4

If you want to change that, how about a new issue?

PR #221, opened 2018-10-03
Refer to the separate CG an BG process document.

### LINKED ISSUE #17: Should the Process define Business and Community Groups?

### LINKED ISSUE #180: Document how to move work from a Community Group into a WG Rec Track document
This is undefined, and some wonder if FSAs apply, some wonder about Member Submission

frivoal:
Closes #17
Closes #180

### DIFF
--- a/index.html
+++ b/index.html
@@ -404,6 +404,9 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
           <a href="#AC">W3C Advisory Committee</a>), the Team, and the general public.</li>
       </ol>
 
+      <p>The W3C also operates Community and Business Groups,
+      which are separately described in <a href="https://www.w3.org/community/about/agreements/">their own process document</a>.
+
       <h2 id="Organization">2 Members, Advisory Committee, Team, Advisory Board, Technical Architecture Group</h2>
 
       <p>W3C's mission is to lead the Web to its full potential. W3C <a href="#Members">Member</a> organizations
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -404,6 +404,9 @@ <h2 id="Intro">1 Introduction</h2>
           <a href="#AC">W3C Advisory Committee</a>), the Team, and the general public.</li>
       </ol>
 
+      <p>The W3C also operates Community and Business Groups,
+      which are separately described in <a href="https://www.w3.org/community/about/agreements/">their own process document</a>.
```
chaals: Should that be incorporated by reference, making it a normative part of the process?

frivoal:
> Should that be incorporated by reference, making it a normative part of the process?

I believe the implication is that this would give the AC jurisdiction over it, and the AB the responsibility to watch over it. Right? It's an interesting question.

index.html
wseltzer: No, I believe it should not be. Additional incorporations would trigger even more up-front legal review, making more of a barrier to entry for even those not intending to do CG/BG work.

dwsinger:
The W3C Process describes how we run Member processes, and CGs and BGs are open to the public.

nigelmegitt:
> The W3C Process describes how we run Member processes, and CGs and BGs are open to the public.

That's such an important distinction it should probably be in the title of the document, like "W3C  Process for Member Activities", say.

frivoal:
@chaals 

We merged this based on the discussion during the process CG call despite your comment saying that:
> Should that be incorporated by reference, making it a normative part of the process?

The rationale was not to necessarily disagree with the comment, but to think that we can go in steps. Mentioning the existence of CGs and BGs can easily be done without incorporating by reference, and was judged useful as such. It also does not conflict with later incorporating by reference, and it felt like that question needed more discussion (see for example https://github.com/w3c/w3process/pull/221#discussion_r222320063 or https://github.com/w3c/w3process/pull/221#issuecomment-426595476).

If you'd like to pursue that, I suggest opening a separate issue.

PR #225, opened 2018-11-05
Let the AB choose its chair(s) and make the CEO an ex officio member

### LINKED ISSUE #223: Allow AB to choose its own chair
https://www.w3.org/2018/Process-20180201/#ABParticipation
> The Advisory Board consists of nine elected participants and a Chair. The Team appoints the Chair of the Advisory Board, who is generally the CEO.

It seems to me that having the CEO chair the AB meeting isn't the best organization. It would make more sense for the AB to appoint someone else as the chair, to optimize for best chairing rather than best CEOing, and let the CEO be free to participate as a regular opinionated participant without having to balance the group.

My suggested wording:

> The Advisory Board consists of nine elected participants, the CEO, and a Chair, who is generally one of the participants of the AB. The AB by consensus (or by vote if that fails) appoints the Chair of the Advisory Board, who sets the agenda with the advice and consent of the AB and the CEO.

This means technically that the AB can choose a chair who is not an elected member of the AB, but I think that's a feature not a bug. :) It might sometimes be useful.

frivoal:
Closes #223

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -596,17 +596,19 @@ Advisory Board (AB)</h3>
 <h4 id="ABParticipation">
 Advisory Board Participation</h4>
 
-	The [=Advisory Board=] consists of nine to eleven elected participants and a Chair.
-	The [=Team=] appoints the Chair of the <a href="#AB">Advisory Board</a>,
-	who is generally the CEO.
+	The [=Advisory Board=] consists of nine to eleven elected participants,
+	the CEO,
+	and a Chair(s), who are generally participants of the AB.
+	The elected members of the AB and the CEO by consensus (or by vote if that fails)
+	appoint the Chair(s) of the Advisory Board,
+	who sets the agenda with the advice and consent of the AB and the CEO.
 	The team also appoints a <a href="https://www.w3.org/Guide/staff-contact">Team Contact</a> for the [=AB=],
 	as described in <a href="#ReqsAllGroups"></a>.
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -647,15 +647,17 @@ <h3 id="AB">2.3 Advisory Board (AB)</h3>
       <h4 id="ABParticipation">2.3.1 Advisory Board Participation</h4>
 
       <p>The Advisory Board consists of eleven elected participants
-        and a Chair.
-        The Team appoints the Chair of the <a href="#AB">Advisory Board</a>, who is generally the CEO.
+        the CEO,
+        and Chair(s), who are generally participants of the AB.
+        The elected members of the AB and the CEO by consensus (or by vote if that fails) appoint the Chair(s) of the Advisory Board,
```
cwilso: This doesn't say anything about how frequently the AB revisits the choice of chair, which is probably a good idea in both directions - you don't want to change chairs every month, ideally, but you also should be forced to ask the question "is the current chair the optimal one" on so often a cycle that it doesn't seem epochal to choose a new chair.

cwilso:
Left a comment.  I should note that I still don't think this is a good idea, fwiw.

LJWatson:
Has #223 been agreed? I see some consensus on the thread, but would expect more discussion/formal consensus before it is translated into a Process change.

frivoal:
@LJWatson I'm not intending to land this PR until/unless we confirm actual consensus. This is just so that we get concrete text to look at while talking about it.

frivoal:
This pull request predates that. Rebasing.

PR #227, opened 2018-11-14
Avoid implying that elected members of the AB/TAG represent an organisation

frivoal:
Phrasing improvement suggested by @swickr in https://github.com/w3c/w3process/issues/208#issuecomment-438494615


swickr:
s/thisis/this is/

dwsinger:
this leaves "the organization must have returned to having at most one seat." but I think the sense is nonetheless clear. OK by me.

frivoal:
> OK by me.

I read that as a personal +1. If you meant it as a chair's declaration of it being good to merge, I'd appreciate an explicit statement to that effect.

frivoal:

PR #228, opened 2018-11-14
Revert to an earlier phrasing

frivoal:
The new phrasing is narrower than the original one, and excludes cases
that shouldn't be excluded.


frivoal:
The [CfC about the 2019](https://lists.w3.org/Archives/Public/public-w3process/2018Nov/0001.html) process is closed and the chair has concluded we do have consensus. Merging.

PR #234, opened 2018-12-20
Replace RFC2777 with RFC3797

### LINKED ISSUE #233: RFC2777 is obsolete
The Process invokes the [RFC2777](https://tools.ietf.org/html/rfc2777) to define the [Verifiable Random Selection Procedure](https://www.w3.org/2018/Process-20180201/#random). RFC2777 is obsolete, and replaced by [RFC3797](https://tools.ietf.org/html/rfc3797) (using W3C vocabulary, we would say that RFC2777 is superseded by RFC3797).

[The Change section](https://tools.ietf.org/html/rfc3797#page-16) does not seem to indicate any difference that would be problematic for us.

We should update accordingly.

frivoal:
Closes #233


frivoal:
AB decision:
> RESOLVED: Update the RFC reference in Process 2019

PR #237, opened 2019-01-31
Spelling fixes and one typo fix.

dbaron:
I'm hoping these are noncontroversial.  All but one ("seat" to "seats") were caught by `aspell` in US English mode, although noticing the "avaiable" was what made me run `aspell`.

I made them against `gh-pages`, but I'm not entirely sure that's the right place.

### DIFF
--- a/index.html
+++ b/index.html
@@ -784,11 +784,11 @@ Advisory Board, TAG, Working Group, Interest Group, Workshop, charter, Working D
         <em class="rfc2119">must</em> stabilize it no later than the Call for Votes. The Director <em class="rfc2119">should</em>
         announce appointments no later than the start of a nomination period as part of the Call for Nominations.</p>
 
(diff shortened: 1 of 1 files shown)


frivoal:
> I made them against gh-pages, but I'm not entirely sure that's the right place.

It is for now. We're going to switch to bikeshed soon, and then changes will be made on the master branch and auto-published to gh-pages, but that hasn't happened yet.

> I'm hoping these are noncontroversial.

I certainly have no issue with any of these.

The only thing I'm wondering is whether we should fix them in the 2019 process (whose approval cycle ends in 1 day, with no other change suggested) or the 2020 process. By my reading of the process (section 7.1.2), this kind of minor change can be integrated without needing to trigger any re-vote or any other process churn, so I'd be inclined to fold them in asap. Chair (@dwsinger), what do you think?

dwsinger:
I think spelling fixes are fine. we should push this before publication.

PR #240, opened 2019-02-07
[editorial] Grammar fix: add missing auxiliary

frivoal:


frivoal:
Agenda+ for quick approval. This is small enough that I'm also happy to get an asynchronous chair decision.

frivoal:
I'm not opposed to switching to "may", but this seems to be more of a description of capability rather than a normative requirement on its own. The list that follows does cover the normative requirements of how to do that.

I think "can" reads better, but if others would like a "may" here, I'll oblige.

chaals:
Anybody *can* change the confidentiality by putting the information somewhere more people can see it. I think it is a normative requirement. The team and delegates may, all other parties to the information must not. (Which means without team consent, since that would amount to being authorised / delegated)

frivoal:
Ok, you two convinced me. Updating.

PR #247, opened 2019-02-07
Remove extraneous phrasing

frivoal:
There's no need to be overly polite, nor to suggest that this is a note, which is normally used for informative text, while this definition is meant to be normative.


frivoal:
Agenda+ for quick approval. This is small enough that I'm also happy to get an asynchronous chair decision.

cwilso:
LGTM.  I'd suggest being more descriptive in the title in the future, though - "Remove excessive politeness in defining terms."

PR #254, opened 2019-02-19
[director-free] Gratuitous mentions of the Director

frivoal:
This pull request removes a number of gratuitous / decorative / redundant mentions of the Director. It changes nothing about the decision power of the Director, instead it focuses on two things (in separate commits):
* Redundant mentions of the Director,  such as things like “the Director’s decision to FOO” when FOO is already defined to be something only the Director can decide.
* Cases where the team must announce something and where no decision is involved. In such cases claiming that the announcement is made in the name of the Director isn't very helpful. Since the Team does these announcements in practice, just replace “the Director must announce…” by “the Team must announce…”.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -258,7 +258,7 @@ Members</h3>
 					reviews plans for W3C at each <a href="#ACMeetings">Advisory Committee meeting</a>;
 
 				<li>
-					reviews formal proposals from the W3C Director:
+					reviews formal proposals from the W3C:
 					<a href="#CharterReview">Charter Proposals</a>,
 					<a href="#RecsPR">Proposed Recommendations</a>,
 					and <a href="#GAProcess">Proposed Process Documents</a>.
@@ -1273,7 +1273,7 @@ Recording and Reporting Formal Objections</h4>
 	and propose changes that would remove the [=Formal Objection=];
 	these proposals <em class="rfc2119">may</em> be vague or incomplete.
 	[=Formal Objections=] that do not provide substantive arguments
-	or rationale are unlikely to receive serious consideration by the Director.
+	or rationale are unlikely to receive serious consideration.
 
 	A record of each [=Formal Objection=] <em class="rfc2119">must</em> be <a href="#confidentiality-change">publicly available</a>.
 	A Call for Review (of a document) to the Advisory Committee <em class="rfc2119">must</em> identify any [=Formal Objections=].
@@ -1472,7 +1472,7 @@ Dissemination Policies</h2>
 
 	<ul>
 		<li>
-			<a href="#Reports">W3C technical reports</a> whose publication has been approved by the Director.
+			<a href="#Reports">W3C technical reports</a> whose publication has been approved.
 			Per the Membership Agreement,
 			W3C technical reports (and software) are available free of charge to the general public;
 			(refer to the <a href="https://www.w3.org/Consortium/Legal/copyright-documents">W3C Document License</a> [[DOC-LICENSE]]).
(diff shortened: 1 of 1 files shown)


jeffjaffe:
While I agree with Florian's logic, since we IAC plan to relook at the Role of the Director, I would prefer not to make this change at this time.  Making the change will require me to find the Director to ensure that he agrees, which seems like an unnecessary intrusion on his time under these circumstances.

index.bs
```diff
@@ -1626,8 +1626,7 @@ Requirements for All Working and Interest Groups</h3>
 	The Director appoints (and re-appoints) Chairs for all groups.
 	The Chair is a <a href="#member-rep">Member representative</a>,
 	a <a href="#Team">Team representative</a>,
-	or an <a href="#invited-expert-wg">Invited Expert</a>
-	(invited by the Director).
```
cwilso: I don't think this one is necessarily gratuitous.  Invited Experts generally need to be invited by the Chair, do they not?  But this is to INVITE the chair, so it needs to be clear who has that authority.

frivoal: You are correct. I got confused because the Director can rescind the invitation of an IE, and that somehow made me think that IE's invitation always came from the Director. But they do not, so this change is incorrect. Will revert.

frivoal:
@jeffjaffe For any change to regarding the Director that actually changes what the Director does, I'd agree that we should not do that without involving him. I feel more comfortable with this edit because because it is effectively a stylistic edit, that doesn't change anything other than how many time the word "Director" is mentioned.

I think the substantive discussions about the role of the Director, which ever way they go, will be simplified if we can look at the process that mentions the Director when he actually has a role, so that we don't have to deal with the mental overhead (and the git conflicts) or remembering that half the time he's mentioned it doesn't actually do anything.

jeffjaffe:
@frivoal fwiw, the reason that I feel that the Director needs to be involved even for an apparently stylistic edit is because people don't always agree on what is stylistic: cf. IE Chair example found by @cwilso

dwsinger:
OK.  Let's do all the "role of the Director" edits on a branch, with reasonably distinct pull requests for each 'layer', roughly:

* remove cross-referential, gratuitous non-defining, mentions of the Director
* change the simple cases where it’s described as the Director but actually the team (announces, mostly)
* change the cases where it’s the CEO (discipline and management mostly)
* change the cases where we remove the Director 
— TAG membership, chair, and ability to appoint,
— hearing formal objections and chair appeals

I think there was some thought that maybe we could do the first two on main branch, but on reflection, it's probably better to do all this in a sandbox.

index.bs
```diff
@@ -2924,12 +2924,12 @@ Advancement on the Recommendation Track</h4>
...
 	there is usually a formal review meeting
-	to ensure the requirements have been met before [=Director=]'s approval is given.
+	to ensure the requirements have been met before approval is given.
 
 	[=Transition Requests=] to [=First Public Working Draft=]
 	or [=Candidate Recommendation=]
 	will not normally be approved
-	while a [=Working Group=]'s [=charter=] is undergoing or awaiting a [=Director=]'s decision
+	while a [=Working Group=]'s [=charter=] is undergoing or awaiting a decision
 	on an [=Advisory Committee Review=].</p>
```
wseltzer, line 2933: s/on an AC Review/after an AC Review/ for clarity

frivoal: I think I prefer it the way it is, but I don't feel strongly. Any third opinion?

wseltzer: agree to leave it unchanged

PR #255, opened 2019-02-19
Retire the “Team Submission” Mechanism

### LINKED ISSUE #246: Does the “Team Submission“ mechanism still serve a purpose?
The [team submission Process](https://w3c.github.io/w3process/#TeamSubmission) has been used once in the last 10 years. The Team certainly has useful things to say, and shouldn't be prevented from doing so, but it is not clear that this specific mechanism is still relevant or needed to enable that.

Given that the Team Submission process offers no Patent Policy coverage, it seems that between a the Team's Blog for opinion pieces, CGs (or WGs) for technical work, /Guide, https://w3c.github.io/..., or and the fact that the Team maintains the W3C website at large anyway, there's enough means for Team members to express themselves.

I wonder if we couldn't simplify the W3C Process a little by doing away with Team Submissions entirely (existing content can of course stay).

frivoal:
This is a historic mechanism that does not really get used nowadays, given the broad availability of alternative ways for the Team to make it's opinions known.

Closes #246

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -539,23 +539,6 @@ The W3C Team</h3>
 	Within W3C, the Host institutions are governed by hosting agreements;
 	the [=Hosts=] themselves are not W3C Members.
 
-<h4 id="TeamSubmission">
-Team Submissions</h4>
-
-	Team members <em class="rfc2119">may</em> request that the Director publish information at the W3C Web site.
-	At the Director's discretion,
-	these documents are published as “Team Submissions”.
-	These documents are analogous to <a href="#Submission">Member Submissions</a>
-	(e.g., in <a href="#SubmissionScope">expected scope</a>).
-	However, there is no additional Team comment.
-	The <a href="#DocumentStatus">document status section</a> of a Team Submission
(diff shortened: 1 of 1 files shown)


frivoal:
@dwsinger, given the thumbs up from @plehegar, as well as the discussion in the issue and on the call, are you looking for any further feedback? If yes, please characterize it and I'll take an action to go get that feedback. If no, I'd appreciate an asynchronous decision (after a CfC if you think it's appropriate), as our calls are rather infrequent (and likely to be consumed by topics with stronger needs for discussion, like Evergreen Standards), and I'd like to move faster when possible.

dwsinger:
If the team is OK, I am ok.  It's easily put back if we change our mind.

michaelchampion:
I don't have any remaining concerns about this, so it can go for broader review to see if the AC sees an ongoing purpose for the Member/Team submission mechanism.

frivoal:
> If the team is OK, I am ok.

Is that a personal opinion, or a ProcessCG Chair approval to merge? (or both)?

dwsinger:
I don't expect further feedback. I think you can delete team subs and add the change to the log, so it gets highlighted.

PR #257, opened 2019-02-19
Rephrase incorrect uses of the specialized term “publish”

### LINKED ISSUE #245: Incorrect use of the specialized term "publish"
[Section 6.1](https://w3c.github.io/w3process/#rec-advance) says:
> Publishing as used in this document refers to producing a version which is listed as a W3C Technical Report on its Technical Reports page https://www.w3.org/TR

However, [Section 10 on member submissions](https://w3c.github.io/w3process/#Submission)
>After review, the Team may **publish** the material at the W3C Web site.

Given that member submissions are not included in /TR (as correctly pointed out in a note further down in the same section), publish is the wrong verb to use in that sentence. I suggest replacing with "may make the material available at the W3C website".

frivoal:
Closes #245

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4157,11 +4157,11 @@ Member Submission Process</h2>
 	or other ideas
 	for consideration by the [=Team=].
 	After review,
-	the [=Team=] <em class="rfc2119">may</em> publish the material at the W3C Web site.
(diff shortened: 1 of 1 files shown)


DAYNAxyz:
Dayna Rae Hurtado
 Github
     WORLDWIDE

PR #270, opened 2019-05-14
Define and differentiate between Group Decisions and Chair Decisions

### LINKED ISSUE #249: Define "Working Group Decision"
There is a large section on Consensus and decision making, but it fails to actually define what constitutes a "Working Group Decision", despite that term being important to various fundamental stages of the process.

There should be a clear definition of how such a decision comes to exist and the grounds and procedures under which it can be challenged, or we should clearly require that it is defined in charters.

frivoal:
Closes #249

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1382,7 +1382,7 @@ Votes</h3>
 	Procedures for <a href="#ACVotes">Advisory Committee votes</a> are described separately.
 
 <h3 id="WGAppeals">
-Appeal of a Chair's Decision or of a Group Decision</h3>
+Appeal of [=Chair Decisions=] and [=Group Decisions=]</h3>
 
 	Groups resolve issues through dialog.
 	Individuals who disagree strongly with a decision
--- a/index.bs
+++ b/index.bs
@@ -1397,7 +1397,7 @@ Appeal of [=Chair Decisions=] and [=Group Decisions=]</h3>
 	decisions taken by the [=Chair=] of a [=Working Group=] or [=Interest Group=]
 	on the basis of having assessed the [=consensus=] of the group
 	or following a vote (see [[#Votes]])
-	are called <dfn export>Group Decision</dfn>.
+	are called <dfn export>Group Decisions</dfn>.
 
 	When group participants believe that their concerns are not being duly considered by the group or the [=Chair=],
 	they <em class="rfc2119">may</em> ask the <a href="#def-Director">Director</a>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1382,18 +1382,28 @@ Votes</h3>
...
 	(e.g., to a decision made as the result of a <a href="#Votes">vote</a>).
 
-	When group participants believe that their concerns are not being duly considered by the group,
+	As detailed in other parts of this document,
+	the [=Chair=] of a [=Working Group=] or [=Interest Group=] has the prerogative
+	to make certain decisions based on their own judgement.
+	Such decisions are called <dfn export>Chair Decisions</dfn>.
+	In contrast,
+	decisions taken by the [=Chair=] of a [=Working Group=] or [=Interest Group=]
+	on the basis of having assessed the [=consensus=] of the group
```
nigelmegitt: A Group Decision might be made as a result of a vote as well as consensus, so add a new line beneath this such as:

```
or following a [=vote=]
```

assuming that decisions arrived at by vote can also be appealed.

index.bs
```diff
@@ -1382,18 +1382,28 @@ Votes</h3>
 	Procedures for <a href="#ACVotes">Advisory Committee votes</a> are described separately.
 
 <h3 id="WGAppeals">
-Appeal of a Chair's Decision</h3>
+Appeal of a Chair's Decision or of a Group Decision</h3>
```
nrooney: Maybe separate these sections out? It's a little complex being both "chairs decision" and "group decision". It may be best for one section as "Appeal of a Chair's Decision" and "Appeal of a Group Decision".

index.bs
nigelmegitt: Discussed at https://www.w3.org/2019/05/22-w3process-minutes.html#item01 and there seems to be consensus that appeals of group decisions arrived at by vote should be permitted.

index.bs
frivoal: Separating in two sections would mean duplicating most of the text, and would make it harder to explain the different between the two types of decisions. However, I do agree that the title was clumsy, so I've rephrased.

frivoal:
Hi all, I think this is ready to merge. If anyone disagrees, please say so. Otherwise, @dwsinger, can you declare a resolution to merge?

dwsinger:
I'm OK except for the mismatch between singular and plural in "In contrast, decisions taken by the [=Chair=] of a [=Working Group=] or [=Interest Group=] on the basis of having assessed the [=consensus=] of the group or following a vote (see [[#Votes]]) are called <dfn export>Group Decision</dfn>"

either "are called Group Decisions" (plural) or if you want to retain a singular definition, "a decision taken by the chair…is called a Group Decision".

frivoal:
> I'm OK except for the mismatch between singular and plural

Thanks. Fixed.

dwsinger:
agreed on the process call July 24th

PR #274, opened 2019-05-20
Remove reference to stale document

frivoal:
Section 2.1.1 includes this:

> Refer to the [Guidelines for Disciplinary Action](https://www.w3.org/2002/09/discipline).

This links to a document that is not publicly viewable. We briefly discussed in the AB whether it should be made public, but the conclusion is that while there's nothing secret there, it is too outdated to be useful.

This pull request therefore proposes to remove references to it.

If we eventually come around to make an up to date version of it, that could be added back to the process, but until then, this is just cruft.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -322,7 +322,6 @@ Rights of Members</h4>
 	and repeated attempts to address these violations have not resolved the situation,
 	the [=Director=] <em class="rfc2119">may</em> take disciplinary action.
 	Arbitration in the case of further disagreement is governed by paragraph 19 of the Membership Agreement [[MEMBER-AGREEMENT]].
-	Refer to the <a href="https://www.w3.org/2002/09/discipline">Guidelines for Disciplinary Action</a> [[DISCIPLINARY-GL]].
 
 <h4 id="RelatedAndConsortiumMembers">
 Member Consortia and Related Members</h4>
@@ -4998,11 +4997,6 @@ and the <a href="https://www.w3.org/2018/Process-20180201/">1 February 2018 Proc
 	"MEMBER-SUB": {
 		"href": "https://www.w3.org/2000/09/submission",
 		"title": "How to send a Submission request",
-		"publisher": "W3C"
-	},
-	"DISCIPLINARY-GL": {
-		"href": "https://www.w3.org/2002/09/discipline",
-		"title": "Guidelines for Disciplinary Action",
 		"publisher": "W3C"
 	},
 	"ELECTION-HOWTO": {


frivoal:
@dwsinger can we get an asynchronous resolution to merge for this?

nigelmegitt:
In my view it would be better to leave the link in, add informative text noting that it is visible only to members, and as a completely independent activity update the document as needed.

frivoal:
@nigelmegitt what good does it do to keep a reference to a document that isn't visible by most people, and that is too outdated to be useful for the people who can see it?

People regularly complain that the process is too long. I'd rather remove the bits that don't do anything useful.

nigelmegitt:
@frivoal it is visible to the people to whom it might apply. Even if it is outdated it is actually still somewhat useful and it is important to have _something_ rather than _nothing_. 

I have had enough conversations that go along the lines:

> "This behaviour is not good"
>
> "There's no reference documentation that proscribes it or sets out any consequences so I'm free to do what I like in this space"

to know that having _something_ written down can make a huge difference. Which conclusion do you prefer to the above hypothetical conversation?

A:

> "Hmm that's strange I thought there was some documentation but I can not find it, I'll raise an issue"
> -> problem gets kicked into the long grass, nothing _actually_ happens, repeat unwanted behaviour occurs.

B:

> "Yes there is, [document] says so, and outlines the possible consequences"
>
> "Okay I guess I shouldn't do that in future then"

Unfortunately these conversations _do_ happen and as a Chair I would very much not like to have the tools available to me removed, sending down path A, when path B could easily remain available.

dwsinger:
@nigelmegitt The code of ethics and professional conduct is explicitly linked and made a normative requirement of participants in W3C. So this is additional...and maybe superfluous

frivoal:
My initial request was that this document be made public, because nothing in it is secret, and as you say, it may be useful. The AB resolve that we should not make it public, as it is too outdated to be useful (and the CEPC was added later, as well as disciplinary powers for the Director).

Either the document is useful, and it should be public, or it is not, and it should be removed.

nigelmegitt:
> Either the document is useful, and it should be public, or it is not, and it should be removed.

Those aren't logically mutually exclusive options. If the AB thinks it should be non-public, that does not in itself render it not useful (@frivoal I take your point that the reason they wanted to keep it non-public is because it is not useful; it happens that I would disagree with that assessment, for what it's worth, but that's not really important right now). Hence my suggestion to add informative text that it is not public.

frivoal:
> @frivoal I take your point that the reason they wanted to keep it non-public is because it is not useful; it happens that I would disagree with that assessment, for what it's worth, but that's not really important right now

I think it is important. Public or not, if this text is not useful, it should be removed. I am not expressing a strong* personal judgment on whether it is useful, I'm merely taking the AB resolution that it is not.

If we insist that it is useful and that we should keep it on that basis, I will insist that it becomes public. There is no reason why useful information on how abusive situations are handled should be restricted to a certain class of participants.

frivoal:
@nigelmegitt Based on my last comment, what do you think?

In addition, I'll concur with the AB assessment that the content of this document is not merely old, it's wrong. It says:
> There is precedent for disciplinary action in the Process Document: "bad standing". [...] Rather than attempt to enumerate additional violations and corresponding actions in the Process Document, [...]

Bad standing is no longer something that the Process defines. However, the Process does not merely establish precedent for disciplinary action, and explicitly now grants the Director that the power.
The document also ignores the existence of ombuds, and of the CEPC (which the process does refer to), and associated procedures.

I continue to think we should remove that reference, which adds nothing to those who cannot read it, and may confuse those that can.

I co

jeffjaffe:
@nigelmegitt If CEPC is updated at the same time that we next update the W3C Process, does that remove your objection to dropping the stale reference?

nigelmegitt:
> If CEPC is updated at the same time that we next update the W3C Process, does that remove your objection to dropping the stale reference?

@jeffjaffe sorry, no, CEPC violations are only one of a number of potential process violations that are listed in the Guidelines for Disciplinary Action. The others are important too.

dwsinger:
Perhaps we should put this guideline document in a manageable place (Github version) and simply update it, instead. Thoughts?

nigelmegitt:
@dwsinger Yes, that's exactly what we should do.

dwsinger:
I raised an issue with the AB (who created this document in 2004) that it needs updating.

frivoal:
Agenda+ to decide if we should address as part of Process 2020 or defer to a subsequent cycle.

(I think we should accept the pull request, won't object if we don't)

chaals:
I think we should not accept the PR.

I note that in private AB space I made a proposal to update. Hopefully I will be able to get that more publicly visible soon.

dwsinger:
left reference in for the 2020 cycle, please fix the document

dwsinger:
I think we should move it somewhere where it can be updated; update it; and as soon as it's moved, change the link in the process (and make sure it's referenced as merely informative).

koalie:
While I'm perusing the information on this PR, I'd like to note the relation with recent issue #418 (I was not aware of this PR when I raised it 3 weeks ago).

frivoal:
This is subsumed by https://github.com/w3c/w3process/pull/432

PR #286, opened 2019-05-23
Clarify rule on "currently vacant" seats

### LINKED ISSUE #238: new "currently vacant" rule seems to preclude resignation in advance that is coincident with an AB/TAG election
While reviewing the proposed process document, I'm not entirely happy about the use of "currently vacant" in the new wording in [section 2.5.2](https://w3c.github.io/w3process/#AB-TAG-elections) that comes from #204.  This is because I'd like it to be clear that a vacancy can be created by a resignation that becomes effective when the election results are in.  This sort of future vacancy could be due to a change of affiliation (the rule in [section 2.5.3](https://w3c.github.io/w3process/#AB-TAG-vacated) that "that participant's seat becomes vacant at the the next regularly scheduled election for that group") or just due to a voluntary resignation that a member of the TAG or AB would like to time with the election to reduce process overhead.

It could be improved, I think, by changing both occurrences of "plus any seat that is currently vacant from the prior year" to something like "plus any vacancies in seats expiring in the alternate year".  But more explicit wording (as there is for the special elections case in 2.5.3, which explicitly says "The group Chair may request an election, and the election may begin, as soon as a current member gives notice of a resignation, including a resignation effective as of a given date in the future." from #62) may also be desirable.

frivoal:
Closes #238

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -780,12 +780,12 @@ Advisory Board and Technical Architecture Group Elections</h4>
 	In the case of regularly scheduled elections of the [=TAG=],
 	the minimum and maximum number of avaiable seats are the same:
 	the 3 seats of the terms expiring that year,
-	plus any seat that is currently vacant from the prior year.
+	plus the number of other seats that are vacant or will be vacant by the time the newly elected members take their seats.
 
 	In the case of regularly scheduled elections of the [=AB=],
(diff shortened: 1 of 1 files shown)


cwilso:
LGTM.

PR #299, opened 2019-07-21
Make it clear than any decision can be objected to

### LINKED ISSUE #281: Chair appointment community involvement, transparency, enabling objections and handling them
When team appointments of Chairs are done as part of a new Charter, AC Reps can Formally Object as part of the Charter review. In case of a Charter Extension, which may at the same time change the Chair, AC Reps can initiate an Advisory Committee Appeal. That seems enough to provide an opportunity for oversight, but there’s currently no way to Object to a change of Chair at other times. Maybe we should allow Formal Objections to Chair changes, or some other process, such as ratification by the AB.

frivoal:
Closes #281

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/299.html" title="Last updated on Jul 15, 2021, 12:32 PM UTC (64e5bfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/299/0e74dbd...frivoal:64e5bfa.html" title="Last updated on Jul 15, 2021, 12:32 PM UTC (64e5bfa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1241,6 +1241,19 @@ Recording and Reporting Formal Objections</h4>
 	Formal Objections receive Director consideration.
 	The word “objection” used alone has ordinary English connotations.
 
+	[=Formal Objections=] <em class=rfc2119>may</em> be registered against any
+	[=Chair Decision=],
+	[=Group decision=],
+	[=W3C Decision=],
+	or decision of the [=Director=], [=CEO=], or [=Team=]
+	within the scope of this Process,
+	except those for which an alternative specific means of appeal
+	(such as [=Advisory Committee Appeal=] or <a href="#SubmissionNo">Submission Appeal</a>)
+	is indicated.
+	However, individuals <em rfc=2119>should</em> first seek to resolve their issues
+	through dialogue,
+	and attempt to find <a lt=consensus>consensual</a> solutions.
+
 	An individual who registers a [=Formal Objection=] <em class="rfc2119">should</em> cite technical arguments
 	and propose changes that would remove the [=Formal Objection=];
 	these proposals <em class="rfc2119">may</em> be vague or incomplete.


dwsinger:
are we 100% sure that the list is comprehensive? If not, I'd suggest "Formal Objections can be registered and against formal decision, including those of…"

frivoal:
That's what I start with, then realized that "formal decision" wasn't defined, so I preferred to expand the definition in place. We should indeed double check, and if anything is missing, add it there. But I think it's correct.

dwsinger:
CEO decisions might include things that are nothing to do with the membership ("I decided to install a water cooler in the offices"). Hm.

frivoal:
punting on this, as we're working on formal objections as part of the director-free process, and that Tackling FOs from multiple angles at the same time isn't the best approach.

frivoal:
@jeffjaffe If I understood you correctly, during the call you said you thought that only the things that are explicitely listed in the process as being possible to formally object to can indeed be formally objected to.

As far as I can tell, hardly anything in the process is explicitly listed as something you can formally object to. Various parts of the process detail what happens **if** someone Formally Object and gives various time when various people must indicate whether there were Formal Objections. These certainly give more than a hint about certain things being possible to formally object to, but specific things are largely not called out as "you can formally object to that". 

3.3.2 does say:
> In the W3C process, an individual may register a Formal Objection to a decision.

I believe the implication is that if a decision, any decision, irks you sufficiently, you can appeal to the Director about it by making your objection formal. This pull request is attempting to make that interpretation explicit, not to introduce new process or new rights.

Several people in the ProcessCG call, who are AB members, expressed support for merging that pull request. So this interpretation of mine is not an isolated view.

You seem to believe that interpretation is wrong. I'd like to hear how you understand the current process: what can or cannot be formally objected to, and how do you tell them apart, if not by the means detailed in this pull request, 



As to your point about various circumstances during which people might be unhappy about chair assignments, I think it's an interesting, but orthogonal point. Even though chair appointments are at the discretion of the Director, and could be changed any time, there are some circumstances (changing affiliation) which the process considers to be important enough to require that a an explicit decision be made whether to maintain or replace the chair. Maybe that is not a helpful criteria, maybe it is. Maybe there should also be a criteria that a decision must be taken when the chair's employer discontinues their browser engine (or some generalization of that concept). Maybe there shouldn't be. But if the Process calls something out specifically as needing a decision, then that is what makes it possible to object to the decision when it is taken.

jeffjaffe:
@frivoal , rather than address your entire post which sometimes characterizes my position and sometimes doesn't, I would prefer to address what I think is the substantive piece:

> You seem to believe that interpretation is wrong. I'd like to hear how you understand the current process: what can or cannot be formally objected to, and how do you tell them apart, if not by the means detailed in this pull request.

We have decades of actual experience where people formally object.  Web stakeholders may formally object to Working Group decisions about what goes into a spec.  AC reps may formally object to charter proposals or document reviews.  I do not see the need to make any changes at this time to our _practice_, hence I don't think we need a _process_ change at this time.  If the community wanted to make this more formal, my suggestion would be to formalize what we have actually been doing.

As I mentioned above, we had reached a consensus when discussing #281 that it would be useful to open this "box of the process document" as part of the director-free exercise.  For example, that would give the W3C Council the opportunity to weigh in - whether they want to increase their workload of formal objections.  I continue to think that we have ample time to consider that issue at that time.

mnot:
> We have decades of actual experience where people formally object. Web stakeholders may formally object to Working Group decisions about what goes into a spec. AC reps may formally object to charter proposals or document reviews. I do not see the need to make any changes at this time to our practice, hence I don't think we need a process change at this time. If the community wanted to make this more formal, my suggestion would be to formalize what we have actually been doing.

I find this characterisation really disturbing. You're conflating "stakeholders _typically_ formally object..." with "stakeholders _may_ formally object..."

Just because a right is rarely exercised does not mean the right doesn't exist.

If the interpretation of the current process is so different between the CEO and the rest of the community, that needs to be clarified as a matter of haste, not put off to the indefinite future.

dwsinger:
I'm concerned; does the CEO (or team) feel that there are formal decisions for which there is no avenue of formal objection? If so, I'd like to know what they are. If not, then clarifying in the process that formally, any decision can be objected to (but you should explore less confrontational avenues first), seems the right thing to do.

mnot:
> We'll need to ensure in other ways that objections to chair appointments are always done professionally.

Very much agreed - I think the W3C is already on the right track here, regarding fostering a professional environment.

frivoal:
@tantek duplication is not the goal, and that's certainly to be avoided, so maybe a rephrasing is in order. However, I suggest you have a look at the 4 comments that start at https://github.com/w3c/w3process/pull/299#issuecomment-545502561 for why I believe a clarification is desirable.

PR #309, opened 2019-08-08
Define how Governing Documents other than the Process are revised

### LINKED ISSUE #305: The process does not include any disposition for updating the Patent Policy
The process does not include any disposition for updating the Patent Policy. Effectively, I suppose this means it can (and can only) be done by the director's fiat, but that's probably not ideal. An alternative understanding would be that since the Patent Policy is normatively referenced by the Process, and since the AB is in charge of the Process, then it is in charge of the Patent Policy as well.

I'd suggest explicitly stating that the Patent Policy can be updated in the same way as the Process can: prepared by the AB, then proposed to an AC Review (which implies a director's approval as well, as long as there is a Director). We might also want to say that PSIG ought to be involved, but then again, PSIG's existence is not mandated by the Process, so I'm not sure. Even if we don't mention them explicitly, so long as PSIG exists in its current form, I'd expect the AB to consult with them, and for the Team / AC / Director to push back if they haven't.

I don't expect updating the Patent Policy will be a frequent occurrence, but that doesn't mean we shouldn't know how to update it when we want/need to.

frivoal:
Closes #305

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4478,12 +4478,22 @@ Rejection of a Submission Request, and Submission Appeals</h3>
 	  that ensures the appropriate <a href="#confidentiality-levels">level of confidentiality</a>.
 
 <h2 id="GAProcess">
-Process Evolution</h2>
+Evolution of the Governing Documents</h2>
 
-	The <span id="ProcessDoc">W3C Process Document</span> undergoes similar consensus-building processes as technical reports,
-	with the [=Advisory Board=] acting as the sponsoring Working Group.
+	The <dfn>W3C Governing Documents</dfn> comprise:
+	* the W3C Process (this document)
+	* the W3C Patent Policy [[!PATENT-POLICY]]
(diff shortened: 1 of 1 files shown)


jeffjaffe:
@frivoal unless I missed it, I don't see where the Team is explicitly assigned a role (other than polling the AC) in this revision.

frivoal:
@jeffjaffe 

> To help assure fitness for purpose,
> the group should solicit input from the W3C community,
> and in particular the Team.

The intent is to make clear that feedback from the team is expected and should be sought for (and once provided, needs to be addressed  as all feedback needs to be). I have tried various ways of phrasing it, but other attempts tended to give the Team veto power, or a dedicated review phase, which isn't what we're trying to do here. The point, as I understand it, is that the Team needs to be in the loop and their feedback taken seriously.

jeffjaffe:
Thanks.  I suggest closing this issue.

Issue #322, opened 2019-08-27
What are the rules for where/how W3C publications are developed?


dwsinger:
This question arose from the Registries work. It used to be that we could assume everyone used a W3C-hosted CVS repository. Now we're mostly using Github. What are the rules that establish what you _can_ use for development?

I suspect that we need at least:
* the ability to file bug reports (aka issues)
* history tracking (so we can see what was changed when, by whom, etc.)
* the W3C maintains a backup (so if there is a disc crash or the service goes down, we don't lose)
* the ability to register for notifications (so people and groups can 'watch' the development; summary reports are a nice to have version of this)
* the development tool/site needs to be accessible (a) to those needing accessible access and (b) to our community internationally

What else?

LJWatson:
It should be somewhere that is recognisably W3C, so that people can 
trust that whatever it is they're looking at is really a W3C activity.

frivoal:
> the W3C maintains a backup

Yes, but I'd be more specific: a back of of the history of the document AND a backup of the issues filed and associated discussions (AND a backup of general discussions not tagged as issues, if such things exist)

frivoal:
I'd also add a SHOULD, unfortunately not met by github:
The W3C should control the domain under which this is hosted (even if it doesn't control the underlying service) so that if the underlying service goes down, we can not only restore the content from backups, but also restore it as the original URLs, to avoid breaking incoming links.

dwsinger:
Yes, that's what I meant by backup. I see your anxiety about the URLs. We could have redirect pointers from w3.org URLs to the repos, of course, and publish those as the official locations. That means we could redirect to a different GIT repo if needed.

So far we have:

The management system used to develop a formal W3C deliverable (technical report, in the process sense):

* must support the ability to file bug reports (aka issues)
* must have history tracking (so we can see what was changed when, by whom, etc.)
* the W3C must a full backup (not only the latest state, but also history, and all conversation),  (so if there is a disc crash or the service goes down, we don't lose)
* must support the ability to register for notifications (so people and groups can 'watch' the development; summary reports are a nice to have version of this)
* the development tool/site should be accessible (a) to those needing accessible access and (b) to our community internationally
* may be publicly viewable, but if restricted, anyone with a W3C account (team, member, or IE) has access ((not restricted to a WG, for example)
* should be clearly marked as an official W3C repository (to distinguish it from clones)
[...]

PR #335, opened 2019-10-08
Registries

frivoal:
Add process for Registries based on https://github.com/w3c/w3process/issues/168 and TPAC discussions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/pull/335.html" title="Last updated on Mar 10, 2021, 4:06 PM UTC (ebd4c99)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/335/3ae9f7c...ebd4c99.html" title="Last updated on Mar 10, 2021, 4:06 PM UTC (ebd4c99)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2259,7 +2259,9 @@ W3C Technical Reports</h3>
 	on its <a href="https://www.w3.org/TR/">Technical Reports page https://www.w3.org/TR</a> [[TR]].
 
 	This chapter describes the formal requirements
-	for [=publishing=] and maintaining a [=W3C Recommendation=] or [=Note=].
+	for [=publishing=] and maintaining a [=W3C Recommendation=],
+	[=Note=],
+	or [=Registry Report=].
 
 	<dl>
 		<dt>Recommendations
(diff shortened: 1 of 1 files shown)


frivoal:
Notes:
* This includes (in two separate commits) the variant where the rules and the tables are together in one document, and the one where everything is inside the section of a REC. It does not include the variant where the rules and the tables could live in separate documents: this is more complicated to define, and there wasn't much appetite for it at TPAC.
* An explainer and/or material for `/guide` will also need to be developed, but this is just the process part. Let's take one thing at a time.
* This does not include the rules / guidance on where and how editors drafts should be developed. We need that too, but process-wise, it's a separate edit/issue (see https://github.com/w3c/w3process/issues/322)

dwsinger:
There was a spirited discussion at TPAC that the tables should not be required even to be in /TR, let alone required to be in the same document. Were we even in the same meeting?

fantasai:
There was a lot of confusion about that part, concerns that it would be the same as trying to publish a new FPWD or CR or something like that, or that machine-readable formats wouldn't be allowed, and so there was a lot of pushback at the start of the discussion portion. But once we explained that you can publish the tables in a machine-readable format, and that the entire process can be automated from accepting a GH pull request to posting on /TR using Echidna, people seemed to be on board afaict.

index.bs
```diff
@@ -3909,6 +3930,342 @@ Working Group and Interest Group Notes</h3>
...
+
+		<li>
+			Define the [=rules of the registry=] for each of its [=registry tables=].
+
+		<li>
+			Include the entire contents of each [=registry table=],
+			either inline in the report
+			(e.g. formatted as a table, or list, or other appropriate representation),
+			or in a machine-readable file published as part of the [=technical report=],
+			or (preferably) both.
```
wseltzer: or as a link to a document or documents published elsewhere.

Eiygy1992 (approved):

PR #357, opened 2019-12-10
spelling fix: judgment

wseltzer:


dwsinger:
see <https://www.dailywritingtips.com/judgement-or-judgment> ;-)

wseltzer:
we should be consistent. elsewhere and previously, it's judgment.

fantasai:
Can I propose the alternative of fixing the existing occurence to 'judgement' instead? :)

nigelmegitt:
Can I propose sidestepping this and changing it to "assessment" instead? 🙂

dwsinger:
I'm going for "ordered, adjudged and decreed"

<https://en.wikipedia.org/wiki/Legal_doublet>

dwsinger:
2.5 has an existing spelling of "judgment". By the powers invested in me, I decree that we shall be consistent and change 3.5 and 6.2.4.1 to the same. This year.

PR #359, opened 2019-12-19
Avoid using the word contributor to limit conflicts with the Patent Policy

frivoal:
The Patent policy is going to need to define the word "Contributor",
and that definition is expected to be different from the one here.
Changing the terminology here is easy, and avoids conflict or confusion.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2547,17 +2547,17 @@ Proposed Changes</h4>
 	<dfn lt="proposed change">proposed changes</dfn>.
 
 <h4 id="contributor-license">
-Contributor License Grants</h4>
+License Grants from Non-Participants</h4>
 
-	When a contributor who is not already obligated under the Patent Policy
+	When a party who is not already obligated under the Patent Policy
 	offers a change in class 3 or 4
 	(as described in [[#correction-classes]]) to a technical report under this process
-	the [=Team=] <em class=rfc2119>must</em> request from the contributor
+	the [=Team=] <em class=rfc2119>must</em> request from them
 	a recorded royalty-free patent commitment;
 	for a change in class 4, the Team <em class=rfc2119>must</em> secure such commitment.
 	Such commitment <em class=rfc2119>should</em> cover,
 	at a minimum,
-	all the contributor's Essential Claims both in the contribution,
+	all the party's Essential Claims both in the contribution,
 	and that become Essential Claims as a result of incorporating the contribution into the draft
 	that existed at the time of the contribution,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2547,17 +2547,17 @@ Proposed Changes</h4>
 	<dfn lt="proposed change">proposed changes</dfn>.
 
 <h4 id="contributor-license">
-Contributor License Grants</h4>
+License Grants from Non-Participants</h4>
 
-	When a contributor who is not already obligated under the Patent Policy
+	When a party who is not already obligated under the Patent Policy
```
chaals: nit, s/obligated/obliged/ Or s/is not already obligated under/has not yet incurred disclosure obligations under/

chaals: s/a party/somebody/ ?

index.bs
```diff
@@ -2547,17 +2547,17 @@ Proposed Changes</h4>
...
 	offers a change in class 3 or 4
 	(as described in [[#correction-classes]]) to a technical report under this process
-	the [=Team=] <em class=rfc2119>must</em> request from the contributor
+	the [=Team=] <em class=rfc2119>must</em> request from them
```
chaals: s/from them// ?

index.bs
```diff
@@ -2547,17 +2547,17 @@ Proposed Changes</h4>
...
 	a recorded royalty-free patent commitment;
 	for a change in class 4, the Team <em class=rfc2119>must</em> secure such commitment.
 	Such commitment <em class=rfc2119>should</em> cover,
 	at a minimum,
-	all the contributor's Essential Claims both in the contribution,
+	all the party's Essential Claims both in the contribution,
```
chaals: s/the party's// ?

index.bs
fantasai: Judging by the definitions, I think we should keep the wording as-is:
https://www.bing.com/search?q=define+oblige
https://www.bing.com/search?q=define+obligate

fantasai: No, because the Process is currently vague on whether "contributor" refers to an individual or the company they work for, and this needs to stay ambiguous.

index.bs
fantasai: wfm

index.bs
fantasai: No, because we can't obligate them for essential claims they don't have. :)

frivoal:
@chaals, for the reasons given by @fantasai, I've accepted https://github.com/w3c/w3process/pull/359#discussion_r359880400, and reject the rest of your comments. Since you said you'd be happy if this were merged as is, I suppose that's OK by you.

dwsinger:
approved to merge on the process CG call Jan 8 2020

PR #361, opened 2020-01-06
Propose removing CR reference to Last Call

wseltzer:


fantasai:
Hi Wendy! I don't think we can adopt this change until we have approval for the new Patent Policy. But if/when that happens, yes, we should definitely clean up these references. :)

chaals:
As far as I can tell, these changes are made feasible by the 1 August 2017 Patent Policy (which was the purpose of that update).

Making the change now seems sensible to me. It increases slightly the gentle pressure to update the Patent Policy again, and has no harmful effects I can see.

frivoal:
@wseltzer The everblue branch has been landed on master, and no longer exists, and this apparently caused github to close this pull request, as it was against the everblue branch. Somehow I cannot just reopen it, so I've created https://github.com/w3c/w3process/pull/363, with the same content.

PR #363, opened 2020-01-09
Propose removing CR reference to Last Call

frivoal:
Recreation of https://github.com/w3c/w3process/pull/361 against the master branch, now that the everblue branch (against which it was based) has been merged.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2765,8 +2765,8 @@ The W3C Recommendation Track</h3>
 		</svg>
 	</div>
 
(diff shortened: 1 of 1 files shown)


dwsinger:
Though the 2017 version comments that Last Call is now called simply Candidate Rec., it's the defined term used in the patent policy (7 times according to a text search).

dwsinger:
Let's consider handling this as a final patent-policy/process consistency pass?

frivoal:
This was replaced by #395.

PR #364, opened 2020-01-09
Drop the "Edited" Recommendation moniker

### LINKED ISSUE #362: "Edited Recommendations" do not serve a useful purpose in an everblue/teal world
The current process has "Edited Recommendations" as a sub-kind of Recommendations.
It is a little conflicted about what the exactly means:
- 6.1.2 claims it's what you get when you make editorial changes to a REC
- 6.7.2 claims it's what you get when you make class 2 or class 3 changes to a REC

I don't believe we've ever issued any document that uses that as its official status, and w3.org/tr doesn't let you search for such things.

Either way, Everblue/teal gives us a new path for revising a REC, and calls the result a REC. This effectively leaves the "Edited REC" label only for RECs revised with editorial-only changes when the revision is done using the same path as previously available. That doesn't sound like a useful distinction.

I suggest dropping the "Edited Recommendations" moniker, and merely acknowledging, in case any such document does exist:
> Previous version of this Process also defined something called **Edited Recommendations**. Going forward, documents with that maturity are considered equivalent in all respects with Recommendations, and new ones will no longer be produced.

frivoal:
Closes #362

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2591,7 +2591,7 @@ The W3C Recommendation Track</h3>
 		<li>Publication of one or more [=Candidate Recommendations=].
 		<li>Publication of a [=Proposed Recommendation=].
 		<li>Publication as a [=W3C Recommendation=].
-		<li>Possibly, publication as an [=Edited Recommendation|Edited=] or [=Amended Recommendation=].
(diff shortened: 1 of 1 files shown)


dwsinger:
agreed to merge on 29th Jan call

PR #376, opened 2020-02-17
Editorial pass

koalie:
s/the W3C/W3C/ (except in e.g. "the W3C process" or "the W3C Patent Policy", etc.)
s/website/Web site/ (I don't feel strongly one way or another, but apparently "Web site" wins)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -228,7 +228,7 @@ Introduction</h2>
 			and the general public.
 	</ol>
 
-	The W3C also operates Community and Business Groups,
+	W3C also operates Community and Business Groups,
 	which are separately described in <a href="https://www.w3.org/community/about/agreements/">their own process document</a> [[BG-CG]].
 
 <h2 id="Organization">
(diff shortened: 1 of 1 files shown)


dwsinger:
weird, this reads funny to me. "The world-wide-web consortium..." reads better than the missing artcile...

koalie:
> weird, this reads funny to me. "The world-wide-web consortium..." reads better than the missing artcile...

In the text recently added to the process document, there were a few occurences of for example "that the W3C endorses" while the rest of the document uses "that W3C endorses" (for example). This pass makes consistent use of "W3C" as a proper noun.

dwsinger:
Reading some reference material:
* acronyms (letter sets that are pronounced as words, like NATO or UNESCO) don't get an article; but that's not us, we have to say double-you-three-see
* initialisms do if their expansion would…maybe? The british broadcasting corporation…the BBC. But then we get CNN that doesn't take an article.

frivoal:
Since this is editorial, not in full agreement, and applying to existing text that isn't new in Process 2020, I'm marking this as deferred to the next cycle.

dwsinger:
Though I am not sure whether we should use "W3C does" or "The W3C does", I have a higher need for consistency than I do for one or the other, and I think we have a head of comms for a reason, and she should get to decide correct usage, for better or worse.

koalie:
> Though I am not sure whether we should use "W3C does" or "The W3C does", I have a higher need for consistency than I do for one or the other, and I think we have a head of comms for a reason, and she should get to decide correct usage, for better or worse.

FWIW, I stand by the corrections I performed in patch-7 and described above 23 days ago.

PR #386, opened 2020-02-18
singular

koalie:


dwsinger:
I think it's old, and fine as a plural.

wseltzer:
editorial; indifferent.

dwsinger:
to sound right, I think it has to change to something like

>Rejection of a Submission Request, and Appeal of a Submission Rejection

(or stay as it is)

frivoal:
Minutes on this topic form the July 15 meeting are here: https://www.w3.org/2020/07/15-w3process-minutes.html#t15

Notably:
> Resolution: Defer to editor

----

This is purely editorial, there isn't 100% agreement about what phrasing is best, so this could go any direction without major issue.

That said, this is a section title, and cross references from other documents sometimes use the section title (rather than its number) to be specific about what they're referring to, so stability has some value.

As the resolution makes this my call, I chose to close it with no action.

@koalie, I hope you can live with this, but if not, you're of course free to reopen.

koalie:
Oh yeah, I can live with it :)

PR #395, opened 2020-04-03
Link to the new Patent Policy

frivoal:


frivoal:
Merging, as approved during the 2020-04-08 call.

PR #413, opened 2020-06-10
Rename proposed changes to candidate changes, and use proposed changes for those under AC Review

### LINKED ISSUE #408: Name for Proposed Changes under patent review
Process 2020 provides for an amendment process for RECs:
1. Changes are "proposed" by marking them into the spec so that they can be reviewed / evaluated / tested.
2. At some point a "last call for review of proposed changes" is issued identifying specific proposed changes for review by the AC and the lawyers.
3. Provided the changes are then accepted, the proposed changes are folded inline as in an updated REC.

We don't have a name for the identified changes, which can be a subset of all the proposed changes in a spec. It might be nice to have a name for them, to help us identify them as being the changes under review.

frivoal:
Closes #408

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2507,14 +2507,14 @@ Errata Management</h4>
 	alongside the affected [=technical report=] text
 	or at the start or end of the most relevant section(s).
 
-<h4 id=proposed-changes>
-Proposed Changes</h4>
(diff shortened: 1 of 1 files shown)


dwsinger:
agreed on the June 24th call

PR #420, opened 2020-07-13
Remove mentions of when various groups were created

frivoal:
We're trying to make the process simpler and shorter and more understandable. The removals proposed in this PR are tiny, but I don't think these pieces of text brought any value at all to the process.

Groups are welcome to document their history, but the Process isn't the place for that.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -553,8 +553,7 @@ The W3C Team</h3>
 <h3 id="AB">
 Advisory Board (AB)</h3>
 
-	Created in March 1998,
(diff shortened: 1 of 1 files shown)


dwsinger:
what am I missing? the Git diff seems to show no difference...

frivoal:
> what am I missing? the Git diff seems to show no difference...

It shows a difference here: lines 556 and 614 are deleted (and the first word in the subsequent line is capitalized)

Issue #137, opened 2017-12-05
Concerns about Amended Recommendations


dwsinger:
From Andreas in AC review:

### Change
* Section 6.7.2
* URL: https://www.w3.org/2017/11/Process-20171102#revised-rec
* Description: W3C can produce an Amended Recommendation to update a Recommendation, without adding new features analogous to the way a Working Group produces an Edited Recommendation.

### Comment
Although this is not marked as blocking comment (to not disrupt the process), there are strong concerns regarding this change. The intent, to bug fix a recommendation when the responsible working group does not exist anymore, is understandable. But the current form leaves a lot of questions open and possibly also gives to much freedom to the W3C team to change a recommendation. The context for an Amended Recommendation need to be better defined. It should be defined which requirements could lead to an Amended Recommendation. What attempts have been made to re-charter a Working Group or at least what have been the reasons to not start such an attempt. The W3C team may not always have the subject matter experts. How do they decide to adopt a change or not? Can an Amended Specification supersede an existing recommendation or make it obsolete?

dbaron:
I wouldn't want to make things too hard, since I feel that good maintenance of specifications is important, and the desire for maintenance is what I think led to the creation of Amended Recommendation.

I would think the goal would be for changes (e.g., incorporation of errata) that are small enough that chartering a new working group wouldn't be appropriate.  Maybe that should be stated explicitly if it isn't already?

Maybe some additional protection might be appropriate, such as:
* explicit sign-off on the changes by a technical reviewer (or two) who was not involved in writing them? (to help address the common problem of "everybody thinks these changes are good at a high level, but nobody actually read them")
* a slightly longer minimum duration for the AC review?

dwsinger:
I think we're going to run with the formal process for the next year, and see how well we do in clearing backlogged maintenance without causing more problems than we're solving. I encourage everyone to watch the process and see whether things slip through cracks.

frivoal:
I share @dwsinger 's concerns about Amended RECs. But also, as far as I know, since the idea was introduced, the W3C has never actually issued an Amended REC. As there have been complaints about the Process being too long, a controversial area that isn't getting used seems to be a good candidate for removal to me.

It was for a time a trendy idea that working groups should be shutdown once they've have accomplished their main goal. I don't believe this is a popular view any more, and that we now generally prefer to keep working groups around when the technology they've specified is in active use, so that they can do maintenance. A number of changes in Process 2020 should make that easier too.

[...]

PR #428, opened 2020-07-20
Retire Amended RECs

frivoal:
This is a possible solution to https://github.com/w3c/w3process/issues/137

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3490,6 +3490,12 @@ Revising a Recommendation: Substantive Changes</h5>
 	and advance the specification from that state.
 	(See <a href="#correction-classes">class 3 changes</a>.)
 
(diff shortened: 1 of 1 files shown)


dwsinger:
do we need a Note instead pointing out our inability to revise in teh absence of an owning WG?

frivoal:
@dwsinger,
> do we need a Note

Seems reasonable to me. I've updated the pull request with such a note, and a bit of editorial massaging of surrounding text to make the Team's ability to do class 1 &amp; 2 changes more readable.

frivoal:
In response to @plehegar 's concern that using candidate corrections instead of amended RECs might lower expectations in terms of the rigor with which the team does maintenance-without-a-WG, I've added some extra requirements.

Re-review welcome.

plehegar (approved):

nigelmegitt (approved):

Issue #422, opened 2020-07-13
What's the point of what the Process has to say about Liasons?


frivoal:
The first two paragraphs and the last paragraph of [Section 9](https://w3c.github.io/w3process/#Liaisons) speak about Liaisons.

Given that Liaisons are defined to cover a broad range of possible relationships with other organizations, and that no particular kind of relationship with other organizations has to be called Liaisons, it's not clear to me that "All liaisons must be coordinated by the Team" (which is the only normative statement) has any teeth.

I suspect we could delete these three paragraphs from the Process and not notice that anything has changed. Given the complains about the process being too long, maybe we should indeed remove them.

A page like https://www.w3.org/2001/11/StdLiaison does a better job of explaing what Liasons are, which one currently exist, and how to set up new ones. It probably would benefit from more visibility on the W3C website, but the w3c website redesign is probably a better means of surfacing it than advertising it in the Process.

jeffjaffe:
@frivoal I agree that we should simplify Section 9.  The first two and last paragraph can probably be collapsed into one paragraph - the current second paragraph is the one with normative importance.

W3C is accustomed to working in an informal fashion which might hide the benefit of the second paragraph, but we work with a vast array of other organizations.  For those who are more formal, the linkage in the process is significant.

We should also explore whether we need an entire section on Liaisons, or whether this better fits (e.g.) as a part of 2.2.

nigelmegitt:
Seems like a sensible area to look at. I believe W3C has an agreement with ISO/IEC about the standing of Recs as international standards. We should check that changing or removing the liaisons section has no impact on that agreement.

dwsinger:
Um, W3C isn't very liaison oriented, true, but they mean a lot to other organizations and maybe therefore to us.

A liaison relationship usually gives reciprocal rights to join groups and attend meetings, outside normal membership rules. It also enables each organization to provide documents (I know, W3C is not contribution-document-based in its workflow) into the other group.

So, you'll have people attending meetings not as an IE, or as a member rep., but as a liaison rep. from a liaison organization; and we could (and probably should in some cases) appoint a liaison rep. the other way, someone who can attend their meetings on behalf of the W3C.

frivoal:
@dwsinger no question that Liaisons are important. What I'm questioning is the existing text about them: seems a lot of words with very limited normative effects.

[...]

PR #433, opened 2020-07-25
Simplify the text on liaisons

frivoal:
Related to #422

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -482,7 +482,7 @@ Advisory Committee Meetings</h5>
 					in particular those started or terminated since the previous Advisory Committee meeting.
 
 				<li>
-					The allocation of resources to pursuing <a href="#Liaisons">liaisons</a> with other organizations.
+					The allocation of resources to pursuing [=liaisons=] with other organizations.
 			</ul>
 		</dd>
 	</dl>
@@ -549,6 +549,18 @@ The W3C Team</h3>
 	and the Massachusetts Institute of Technology (<abbr>MIT</abbr>).
 	Within W3C, the Host institutions are governed by hosting agreements;
 	the [=Hosts=] themselves are not W3C Members.
+
+	All <dfn export>liaisons</dfn> with external organizations
+	<em class="rfc2119">must</em> be coordinated by the [=Team=]
+	due to requirements for public communication;
+	patent,
+	copyright,
+	and other IPR policies;
+	confidentiality agreements;
+	and mutual membership agreements.
+	Refer to <a href="https://www.w3.org/2001/11/StdLiaison">W3C liaisons with other organizations</a> [[LIAISON]]
+	for information about existing liaisons
+	and the guidelines W3C follows when creating a liaison.
 
 <h3 id="AB">
 Advisory Board (AB)</h3>
@@ -1820,7 +1832,7 @@ Invited Expert in a Working Group</h5>
 	The [=Chair=] <em class="rfc2119">may</em> invite an individual with a particular expertise
 	to participate in a Working Group.
 	This individual <em class="rfc2119">may</em> represent an organization in the group
-	(e.g., if acting as a liaison with another organization).
+	(e.g., if acting as a [=liaison=] with another organization).
 
 	An individual is an <dfn export>Invited Expert in a Working Group</dfn> if all of the following conditions are satisfied:
 
@@ -2379,7 +2391,7 @@ Wide Review</h5>
 	the record of requests to and responses from reviewers,
 	especially <a href="https://www.w3.org/Guide/process/charter.html#horizontal-review">W3C Horizontal Groups</a> [[CHARTER]]
 	and groups identified as dependencies in the charter
-	or identified as <a href="https://www.w3.org/2001/11/StdLiaison.html">liaisons</a> [[LIAISON]],
+	or identified as [=liaisons=],
 	and seek evidence of clear communication to the general public
 	about appropriate times and which content to review
 	and whether such reviews actually occurred.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4078,30 +4090,12 @@ Workshops and Symposia</h2>
...
-	and other IPR policies;
-	confidentiality agreements;
-	and mutual membership agreements.
+<h2 id=mou oldids="Liaisons">
+Memorandum of Understanding</h2>
 
 	The [=W3C Director=] <em class="rfc2119">may</em> negotiate
 	a Memorandum of Understanding with another organization.
-	For the purposes of the W3C Process a <dfn id="mou" export lt="Memorandum of Understanding | Memoranda of Understanding | MoU">Memorandum of Understanding</dfn> (<abbr>MoU</abbr>)
+	For the purposes of the W3C Process a <dfn export lt="Memorandum of Understanding | Memoranda of Understanding | MoU">Memorandum of Understanding</dfn> (<abbr>MoU</abbr>)
```
fantasai, line 4098: Don't break links. Keep the ID here.

index.bs
```diff
@@ -550,6 +550,18 @@ The W3C Team</h3>
 	Within W3C, the Host institutions are governed by hosting agreements;
 	the [=Hosts=] themselves are not W3C Members.
 
+	All <dfn export>liaisons</dfn> with external organizations
+	<em class="rfc2119">must</em> be coordinated by the [=Team=]
```
fantasai, line 554: 1. This defining instance doesn't actually define the term. It only defines requirements related to it.
2. Given the broad definition of liaison in W3C's context, and that the issue mentioned how such formality would not be required by the more informal variants of it, do we really want to attach all these requirements to all variants of liaison?

index.bs
frivoal: I have not deleted that id, but moved it to the section title, which only speaks of MoUs now. I can restore it to what it was, but that felt reasonable to me.

index.bs
frivoal: The older text had more commentary about what liaisons were, but that was still not terribly constraining.

Personally, I'd be happy to drop all text about Liaisons from the process, since I don't think it's terribly constraining in the first place.

dwsinger: I think it's important we engage in the usual practice of allowing a liaison representative to attend meetings as if a member of a WG.

index.bs
```diff
@@ -4078,30 +4090,12 @@ Workshops and Symposia</h2>
...
-	to refer to coordination of activities with a variety of organizations,
-	through a number of mechanisms
-	ranging from very informal
-	(e.g., an individual from another organization participates in a W3C Working Group,
-	or just follows its work)
-	to mutual membership,
-	to even more formal agreements.
-	Liaisons are not meant to substitute for W3C membership.
-
-	All liaisons <em class="rfc2119">must</em> be coordinated by the [=Team=]
```
chrisn, line 4094: I think we need to keep this part. W3C does receive formal liaisons from other SDOs, and these sometimes do have restrictions relating to confidentiality. I agree this isn't necessarily "all liaisons", but these steps do apply so some of them.

jeffjaffe:
> E.g. the WHATWG MOU etc was more political handholding than actually necessary for any participation/work in practice.

This is a pretty disappointing assessment.  W3C and WHATWG had struggled for years to find a collaboration model and the MoU has been highly successful in repairing a fractured community.

frivoal:
I think the question of whether past MoUs have accomplished a lot and the question of whether the text in the current process is useful are reasonably separate.

Regardless of whether you think Liaisons and MoUs are useful (I think they can be, but again I think that's besides the point), I believe:
* The current text on liaisons doesn't really mandate nor ban anything. It does say that Liaisons must be coordinated by the Team, but without an constraining definition of what a Liaison is, that is not terribly actionable. Therefore from a normative point of view, I think it is decently close to a no-op, and would like to remove it altogether. Jeff thinks there's some value in that statement, and would like to preserve that bit.
* The current text on MoUs requires that they go through AC ballots first, makes the result of an AC appeal binding, and bases this on a somewhat high level, but still decently clear definition of what MoUs are. This is may or may not be a good idea, but it is not a no-op. For now, I don't suggest changing it.

chrisn:
W3C does receive Liaison statements from other standards bodies from time to time, and there is a process for dealing with them (i.e., coordination by the Team, typically by notifying the relevant WGs and/or IGs). Should the Process document not cover this? If we're lacking a definition, I'm sure we can come up with one.

dwsinger:
I'm less concerned about liaison _documents_ than I am about preserving the ability to have liaison _reprseentatives_ attend meetings (each way).

frivoal:
@dwsinger That's an interesting point, but I am far from convinced that the current text clearly provides for that. How about adding an explicit statement to that effect in https://w3c.github.io/w3process/#group-participation instead?

dwsinger:
@frivoal isn't this text deleted? it's the place where memberships and specifically mutual memberships are mentioned:

> from very informal
> (e.g., an individual from another organization participates in a W3C Working Group,
> or just follows its work)
> to mutual membership,
> to even more formal agreements.
> Liaisons are not meant to substitute for W3C membership.

frivoal:
It's there, but only as an example of a very informal relation. If that just says that we may use invited experts (or even have other organizations become members and send reps) as the vehicle for liaisons, we'd continue to be able to do that without that sentence. Maybe it's trying to say that the team can invite liaison reps to wg meetings separately from them being WG participants under normal rules, but if that's the case, it's not doing it very well: it's only framed as an example, and says nothing about patent policy implications. And then we have "Liaisons are not meant to substitute for W3C membership", which further points to it being the former meaning.

So again, I don't think this text *does* anything. I support the existence of Liaisons, and if we want to formalize who can or must do what with regards to them, we should write that down, but that's separate from trimming no-op text.

dwsinger:
No argument from me that the old text is confusing, with apparent expectations in examples that are not reflected in the normative text, and so on.

Is this a 'high nail' for this year?

frivoal:
> Is this a 'high nail' for this year?

Well, there's a stated goal to reduce complexity and page count, to make the hole thing more digestible. This is an isolated chunk, so it doesn't do much for complexity, but if we do want to get the page count down, looking into removing sections that seem to do nothing seems worth a try.

Alvinchongry:
Editorial

Alvinchongry:
Editorial improvement

PR #434, opened 2020-07-29
Meetings of new groups are 8 weeks from charter approval

### LINKED ISSUE #251: First meeting of a WG allowed too soon
In section [5.2.6](https://w3c.github.io/w3process/#WGCharter)
> If known, the date of the first face-to-face meeting. The date of the first face-to-face meeting of a proposed group must not be sooner than eight weeks after the date of the proposal. 

This does not leave enough time for review, AC ballot, call for participation, and so on.

This means people must make travel plans before knowing whether or when the charter is accepted, possibly despite objecting to it in case their objection is overruled.

Also, if there are indeed objections and the charter is still under negotiation when that planned meeting occurs, then we're having a meeting of a group whose scope, deliverables, work mode, chairs, etc, aren't agreed upon yet. It's unclear what the group can do in such cases. It's also unclear how we can schedule a meeting without trying to get to some degree of consensus about the time and place among the participants, and before a group is launched, it's not clear who the participants are yet. Of course there's already a core group of people driving the charter who are known, but if they meet among themselves without giving other W3C members a chance to join, that's doing a disservice to the broader Membership, and to the new WG as well by making participation hard.

If we're talking about groups who's charter is being renewed, I think that's probably allowable, as we generally know what's going to happen more or less, and a little bit of uncertainty is probably better than the disruption of cancelling / rescheduling a regular meeting.

But for new groups, it feels like this creates an incentive for "approve the charter NOW, not matter the standing objections, because we're meeting anyway". I think that's bad.

I'd delete this clause, and instead add (probably in [5.2.4](https://w3c.github.io/w3process/#cfp) something like:
> New working groups must not schedule a face-to-face meeting less than 12 weeks after the call for participation.

8 weeks is the minimum for any f2f meeting (see [3.2](https://w3c.github.io/w3process/#GeneralMeetings)), so I'm putting a bit more here, 12 weeks, to give some time after the cfp for new members to join and discuss the meeting.

frivoal:
Closes #251. See that issue for the discussion leading to this proposal.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/434.html" title="Last updated on Jul 15, 2021, 12:32 PM UTC (1efa491)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/434/092bbea...frivoal:1efa491.html" title="Last updated on Jul 15, 2021, 12:32 PM UTC (1efa491)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1150,7 +1150,7 @@ Meetings</h3>
 	For new proposed [=Working Groups=] or [=Interest Groups=]
 	that have not yet been chartered,
 	the minimum announcement period
-	is counted from the date of the <a href=#cfp>Call for Participation</a>,
+	starts on the date of the <a href=#cfp>Call for Participation</a>,
 	and meetings <em class=rfc2119>must not</em> be announced before the call has been issued.
 	This is to ensure a fair opportunity to participate for all W3C members,
 	so informal meetings <em class=rfc2119>should not</em> be planned to get around this timing restriction.
--- a/index.bs
+++ b/index.bs
@@ -1147,6 +1147,13 @@ Meetings</h3>
 	about the date and location of a meeting.
 	Shorter notice for a meeting is allowed
 	provided that there are no objections from group participants.
+	For new proposed [=Working Groups=] or [=Interest Groups=]
+	that have not yet been chartered,
+	the minimum announcement period
+	is counted from the date of the <a href=#cfp>Call for Participation</a>,
+	and meetings <em class=rfc2119>must not</em> be announced before the call has been issued.
+	This is to ensure a fair opportunity to participate for all W3C members,
+	so informal meetings <em class=rfc2119>should not</em> be planned to get around this timing restriction.
 
 <h3 id="Consensus">
 Consensus</h3>
@@ -2078,13 +2085,6 @@ Working Group and Interest Group Charters</h4>
 
 		<li>
 			Meeting mechanisms and expected frequency;
-
-		<li>
-			If known,
-			the date of the first <a href="#ftf-meeting">face-to-face meeting</a>.
-			The date of the first face-to-face meeting of a proposed group
-			<em class="rfc2119">must not</em> be sooner than <span class="time-interval">eight weeks</span>
-			after the date of the proposal.
 
 		<li>
 			Communication mechanisms to be employed within the group,


michaelchampion:
Maybe I'm missing something, but the "face to face" qualifier seems to have been removed in the PR?  

I wonder if this has been overtaken by events now that WG meetings are at least temporarily virtual, not F2F.  Is there any point in constraining when a WG can have a virtual meeting?  If so, I suspect it is far less than 8 weeks.

frivoal:
@michaelchampion, the "8 weeks for a f2f" requirements is still there, unchanged. (so is 1 week for a teleconf). What the PR does is 2 things:
- say that the minimum delay counts from the Call for Participation, without changing how long the delay is for various types of meetings
- remove mentions of the first f2f meeting from charters, since it can only be planed after the charter is adopted (and a fortiori, after it's written)

samuelweiler:
Given how specs are commonly incubated, I can imagine a group of people (in a CG?) working on a spec saying "we want to meet about this".  They could.  They do.

If they have a charter pending - maybe one that's been held for an indefinite period while an objection is dealt with - absent this text, they could still and meet, not knowing if the group will be chartered yet or not.

What does this text do to them?  Is seems to lock them into not meeting at all while the charter is pending.  I don't want to see that result.

jwrosewell:
Doesn't this relate to the authority of the group chair(s) to make decisions? I would have thought group chair(s) would be aware of the issues associated with meeting arrangements and ensuring different views are represented.

jeffjaffe:
> Doesn't this relate to the authority of the group chair(s) to make decisions? I would have thought group chair(s) would be aware of the issues associated with meeting arrangements and ensuring different views are represented.

Yes, but the process document is providing a rule for the minimum amount of time before the first F2F meeting.

jwrosewell:
As I understand the W3C Process all charters go to all members for approval (or objection). Therefore members will have plenty of notice about new groups. Chairs can be sanctioned if they discriminate against specific participants. We want to empower chair(s) to run groups in the most appropriate way possible given each group has different dynamics. Therefore perhaps this is best dealt with as a guideline to provide 8 weeks notice for all F2F meetings, and that meetings should not be arranged until after the charter is approved.

michaelchampion:
@wseltzer said 
> This seems overly prescriptive.

Agree, as do other commenters.

The community around a spec being incubated, with consensus determined by a chair, is the best judge of how and when to proceed in a particular situation.  The Process should not be an ever-growing set of "rules"; it should but a guide to *principles* for groups and chairs to find a way to align with and problems to look out for. 

The principle here is something like "groups should make sure that likely meeting attendees have sufficient advance notice to make plans to attend."  The historical problems in this area are that people who should be or want to be in a meeting feel excluded by meeting invitations they don't see or don't have time to plan for.  There should be consequences for not meeting the principles or ignoring actual problems but that's what  AC ballots to create a WG or advance its specs are supposed to enable.  Automating that judgment with a bunch of "rules",  with a large Team to enforce them, doesn't help W3C's  reputation as being an excessively bureaucratic venue to work in.

TzviyaSiegman:
Thanks @wseltzer et al for making these comments. I agree with @michaelchampion that the Process is getting to look like a long list of rules. Why is it necessary to include this level of detail? Different groups might have different needs, and we should allow for that flexibility. Perhaps Mike's wording "groups should make sure that likely meeting attendees have sufficient advance notice to make plans to attend" is sufficient.

dwsinger:
the question was raised because, alas, there was a disagreement and a problem (as is typically the case). sometimes members want to be able to say to chairs "you really can't do that" (and have something that they can point at), and it's all too easy to imagine hearing "you've had plenty of time, we've been talking about a september meeting for months now" while not admitting that the scope/charter of the group were in serious flux over those months, and hence people couldn't be sure they would/could join and/or attend.

samuelweiler:
> ... the scope/charter of the group were in serious flux over those months, and hence people couldn't be sure they would/could join and/or attend.

Why the bright line of uncertainty synchronized with WG formation?

I'm used to seeing work progressing (in "incubation") well before WGs are formed. It seems like "people" would have already joined those efforts or not.

Indeed, I would more worry that people who were involved in incubation - who would be welcome at a meeting of incubating group - would be suddenly unwelcome at a formal WG meeting if they're not affiliated with a W3C member.

[@dwsinger apparently I edited your post above while trying to post this.  sorry if I mangled your text.  I tried to fix it.]

dwsinger:
A meeting of the CG or IG is a meeting of the CG or IG, not an informal meeting of the WG. And that's fine.

frivoal:
If you don't have consensus on your charter yet, you don't have consensus on your charter yet, and your top priority should be to try and get consensus, not to move along as if you had it already, assuming everybody else will just fall in line eventually and can catch up then.

Otherwise, what we're operating under is momentum, not consensus.

samuelweiler:
> If you don't have consensus on your charter yet, you don't have consensus on your charter yet, and your top priority should be to try and get consensus, not to move along as if you had it already, assuming everybody else will just fall in line eventually and can catch up then.
> 
> Otherwise, what we're operating under is momentum, not consensus.

I like tantek's phrase "stop-energy", and I propose the following scenario for your consideration.

Assume for a moment that you thought you had consensus.  Your spec was moving merrily along in WICG or similar.  No objections.  There's lots of interest.  You have several spec editors, a veritable army of people writing code, and even multiple plausible WG chair candidates who are, themselves, writing neither spec nor code.  You schedule a working meeting, mostly so these people writing code can decide that ASN.1 BER is problematic - if they're going to swallow the ASN.1 pill, they need DER - or maybe they'd be happier with XML, all except for a lone holdout.  JSON hasn't even been brought up yet.  They people _need_ to be within shouting distance of each other for a few hours, ideally with a free-flowing supply of beer.

In parallel with arguing about the finer points of data serialization schemes, you started the process of chartering a WG group, only to get a last-day-of-AC review objection.  Maybe the objector thinks your WG should have an additional work item within its scope - they don't see anything wrong with the spec you're working on (well, other than that cursed ASN.1 issue, which they don't care about) - they just think this WG would be a good place to _also_ do their pet project.

It's possible that two or three proponents for the work (and, heck, in this example, even the objector is supportive of the base spec) could work through the objection, maybe by recruiting some additional participants to take on the extra item.  Or maybe by convincing the objector there is a better path.

Should the whole group be held hostage while that is worked out?  Do we really need to tell the people writing code they can't get together in the meantime?

Let's not create failure modes that will inspire people to (further) hesitate before seeking a WG charter.

jwrosewell:
As a newbie who has just submitted a draft charter for the [Decentralized Web Interest Group](https://github.com/w3c/charter-drafts/pull/276) I'm very interested in this issue and the comments. My thoughts are.

1. The process should focus on agreeing the charter. If someone objects then the group can't start until that objection is resolved. If that means people can't write code or meet under the banner of the W3C then so be it. This focused proposers on resolving the objections by gaining consensus, or ultimately making the case to the Director. If people meet outside the W3C then there is little the W3C can do about that.

2. There should be a baseline for all groups that state what the chair can and can't do. The charter can amend that baseline if needed. Therefore this issue can be resolved either by amending the process as proposed, or removing it. Both work in practice.

As simpler is better my view would be to remove this amendment and when there is a specific case of shenanigans during the charter approval process then call it out there.

More pressing issues I have revolve around defining when incubation finishes and working groups start, and implementers being encouraged to achieve a defined W3C consensus threshold before making changes or features widely available.

michaelchampion:
> If that means people can't write code or meet under the banner of the W3C then so be it.

As a W3C 'oldbie" I suspect you have a very different understanding of what the organization is or can do than the way it has operated for last 25 years or so.  Perhaps W3C should evolve to be the organization you think it should be, or some new organization will evolve to replace it, that's a question for some new generation to address.  But W3C as it is today:
- ... is mostly a *venue for collaboration* among people with diverse, sometimes conflicting, objectives and business models, but who also understand the benefits of cooperative development of a common platform.  Its essence is https://en.wikipedia.org/wiki/Coopetition 

- ... develops *voluntary* standards whose only enforcement is via the marketplace.  For example, browsers that don't fully interoperate with each other tend to get rejected by consumers if their favorite websites don't look or act right. Since it does NOT certify compliance with its standards, so it leaves even the determination of whether a product conforms to its standards up to the marketplace (which tends not to distinguish de facto standards from "real" standards) . 

- ... generally endorses what has evolved in incubation or prototyping efforts.  This Process CG was created largely to drive Process changes that reflect the reality that web standards today are driven more by the open-source implementers than by the traditional spec-first standards process.  The challenge now is to find a reasonable balance between the concerns of implementers and the concerns of users; going back to the spec-first model is almost certainly impossible.

- ... gets ignored when it tries to apply "[stop energy](http://www.userland.com/whatIsStopEnergy)" to an idea that has a critical mass of supporters and implementers.  "[HTML5](https://en.wikipedia.org/wiki/HTML5)" is the obvious example of what happens when W3C tries to push something like the  [XHTML/XForms](https://www.techrepublic.com/article/the-future-of-web-forms-from-html-to-xhtml-to-xforms/)  approach that made a lot of theoretical sense but got little developer support.  The consequence was HTML and DOM moving out of W3C's control.

This reality is certainly problematic, notably because the unfettered marketplace for web technologies has become controlled by semi-monopolies that leverage profits from one domain to dominate adjacent domains, setting de facto standards along the way.  But W3C can do little about the underlying realities such as economies of scale and Metcalfe's Law that enable the behemoths to get their way (until they over-reach and consumers rebel or governments intervene).   

For better or worse, W3C is effectively the somewhat neutral venue where behemoths seek common ground and the smaller players have some ability to join the conversation.  Things work reasonably well when essentially everyone shares a common goal, such as making the web interoperable, and universally accessible to people with different abilities and languages.  Things tend to get stalemated when conflicting cultural values or business models make it difficult to find common ground.  

So, I respectfully suggest that you won't get far insisting that "implementers are encouraged to achieve a defined W3C consensus threshold before making changes or features widely available.  And while its true  "If people meet outside the W3C then there is little the W3C can do about that", doing things that encourage people to move conversations outside W3C is generally not a good idea.

index.bs
```diff
@@ -1147,6 +1147,13 @@ Meetings</h3>
 	about the date and location of a meeting.
 	Shorter notice for a meeting is allowed
 	provided that there are no objections from group participants.
+	For new proposed [=Working Groups=] or [=Interest Groups=]
+	that have not yet been chartered,
+	the minimum announcement period
+	starts on the date of the <a href=#cfp>Call for Participation</a>,
+	and meetings <em class=rfc2119>must not</em> be announced before the call has been issued.
```
dontcallmedom, line 1154: not sure if that rule is still being pursued, but an impact of this is that it's likely to create a 3 to 4 months period of the year that is bad to launch new groups due to TPAC scheduling: assuming an end-of October physical TPAC, groups needs to be launched before mid-August to be part of TPAC (a much thought-after goal for new groups), right at the end of one of the worst period to run an AC review (mid-July to mid-August due to northern hemisphere Summer vacation time). Given how challenging launching a group can be, I'm not sure this is an additional barrier that should be added

dwsinger: ah. I wonder if we can be more relaxed about TPAC because, after all, everyone comes anyway, right?

frivoal:
oops, closed accidentally. Reopening.

samuelweiler:
This has been sitting with no evident consensus for over two years.  I propose to close both this and #251.

frivoal:
As I stated in https://github.com/w3c/process/issues/251#issuecomment-1653556553, I still think this problem is worth solving, but the particular solution proposed in this PR isn't getting traction, so I am retracting my proposal and closing.

Issue #435, opened 2020-08-06
Tooling Requirements


fantasai:
Tooling across WGs is fragmenting severely, and is sometimes resulting in inaccessible WG operations. We should include in our Process some of the core requirements, e.g. archived for posterity, accessible to participants with disabilities and in different countries, documented so that newcomers and observers can find discussions, etc. (And put in a separate policy document details / best practices / Team support status / etc. that can be maintained by the systeam to reflect the current state of the world.)

Some initial thoughts on this topic have been drafted into https://www.w3.org/wiki/W3C_Tooling_Policy https://www.w3.org/wiki/Tooling_Considerations https://lists.w3.org/Archives/Member/w3c-ac-forum/2020AprJun/0137.html and https://github.com/w3c/AB-memberonly/issues/30

fantasai:
Agenda+ against https://github.com/frivoal/w3process/commit/31d42c9f1f0fc7ef6f378498dc4a9c72b3ecad8c

samuelweiler:
> accessible to participants with disabilities and in different countries

I'm fine with requiring accessibility for disabilities.

I don't want to see our groups' tooling choices held hostage by any nation's censorship choices. Censorship violates the "web-for-all" principle cited in the draft statement in the wiki.  Even when it's just a single tool being blocked, we should not let censorship keep groups from using the best tools for the job.

Unfortunately, it's not a single tool being blocked.  This weekend we saw reports of [China blocking all traffic using certain recommended security mechanisms (namely TLS 1.3 with ESNI)](https://www.zdnet.com/article/china-is-now-blocking-all-encrypted-https-traffic-using-tls-1-3-and-esni/).

I'm tempted to suggest we should not complicate process with this at all.  The Process is already too complicated.  But if we're going to do it, this should not be a piece of the whole.

fantasai:
> I don't want to see our groups' tooling choices held hostage by any nation's censorship choices.

I don't want to see participants excluded because of what country they live in.

> Censorship violates the "web-for-all" principle cited in the draft statement in the wiki.

The Web can't be for "all" if we exclude the participation of entire populations based on their residency. Being inclusive to all participants is more important than being convenient to all participants. A tool simply cannot be the best tool for a Working Group to do its job if some of its members are unable to use it and are thereby excluded from participation.

dwsinger:
[...]

PR #436, opened 2020-08-06
New section on tooling

frivoal:
Taking a first stab at solving #435. This is derived from https://www.w3.org/wiki/W3C_Tooling_Policy. It includes the abstract version of the most critical requirements listed there. The intent is that more details about how to do that in practice, as well as the things that are more best practices than strict rules, should be detailed in a (to be written) companion policy document.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/436.html" title="Last updated on Mar 16, 2021, 4:49 AM UTC (546f01f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/436/60a9548...frivoal:546f01f.html" title="Last updated on Mar 16, 2021, 4:49 AM UTC (546f01f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1557,7 +1557,7 @@ Tooling</h3>
 
 	</ul>
 
-	The Team is responsible for ensuring adherence to these rules
+	The [=Team=] is responsible for ensuring adherence to these rules
 	and for bringing any group not in compliance into compliance.
 
 	Issue: Link to policy document with detailed recommendations / best practices / additional info.
--- a/index.bs
+++ b/index.bs
@@ -1496,7 +1496,7 @@ Tooling</h3>
 	and across time.
 	Thus in order to allow all would-be participants to effectively participate,
 	to allow future participants and observers to understand the rationale and origins of current decisions,
-	and to guarantee long lived access to its publications,
+	and to guarantee long-lived access to its publications,
 	W3C requires that:
 
 	<ul>
@@ -1539,7 +1539,7 @@ Tooling</h3>
 			<em class=rfc2119>must</em> be archived by W3C for future reference.
 			This includes discussions conducted over email lists
 			or in issue-tracking services
-			or any equivalent fora.
+			or any equivalent forum.
 			For fora whose archive exists on a domain separate from its active operations,
 			the archive <em class=rfc2119>must</em> be available at the same level of confidentiality.
 
--- a/index.bs
+++ b/index.bs
@@ -1535,8 +1535,8 @@ Tooling</h3>
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>must</em> be usable
-			without cost (direct or indirect)
-			by all active or would-be active group participants
+			without additional cost
+			by all who wish to participate,
 			to allow their effective participation
 			regardless of disability
 			or geographical location.
--- a/index.bs
+++ b/index.bs
@@ -1114,7 +1114,7 @@ Meetings</h3>
 	A [=Chair=] <em class="rfc2119">may</em> invite an individual with a particular expertise
 	to attend a meeting on an exceptional basis.
 	This person is a meeting guest,
-	not a group <a href="#participant">participant</a>.
+	not a group [=participant=].
 	Meeting guests do not have <a href="#Votes">voting rights</a>.
 	It is the responsibility of the Chair to ensure
 	that all meeting guests respect the chartered <a href="#confidentiality-levels">level of confidentiality</a>
@@ -1387,7 +1387,7 @@ Votes</h3>
 	</ul>
 
 	In order to vote to resolve a substantive issue,
-	an individual <em class="rfc2119">must</em> be a group <a href="#participant">participant</a>.
+	an individual <em class="rfc2119">must</em> be a group [=participant=].
 	Each organization represented in the group <em class="rfc2119">must</em> have at most one vote,
 	even when the organization is represented by several participants in the group
 	(including Invited Experts).
@@ -1486,6 +1486,76 @@ Resignation from a Group</h3>
 	The team <em class="rfc2119">must</em> record the notification.
 	See “Exclusion and Resignation from the Working Group” in the W3C Patent Policy [[PATENT-POLICY]]
 	for information about obligations remaining after resignation from certain groups.
+
+<h3 id="tooling">
+Tooling</h3>
+
+	For [=W3C Groups=] operating under this Process,
+	a core operating principle is to allow access across disabilities,
+	across country borders,
+	and across time.
+	Thus in order to allow all would-be participants to effectively participate,
+	to allow future participants and observers to understand the rationale and origins of current decisions,
+	and to guarantee long lived access to its publications,
+	W3C requires that:
+
+	<ul>
+		<li>
+			Any tooling used by the group
+			for producing its documentation and deliverables
+			or for official group discussions
+			<em class=rfc2119>must</em> be usable
+			without cost
+			by all active or would-be active group participants
+			to allow their equal participation
+			regardless of disability
+			or geographical location.
+
+			Note: If a new participant joins who cannot use the tool,
+			this can require the [=Working Group=] to change its tooling
+			or operate some workaround.
+
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			(i.e. intended for use or reference outside its own membership)
+			<em class=rfc2119>must</em> be published on a W3C-controlled domain
+			and backed up by W3C systems
+			such that if the underlying service is discontinued,
+			W3C can continue to serve such content without breaking links
+			or other key functionality.
+
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			<em class=rfc2119>must</em> be accessible worldwide
+			and to persons with disabilities.
+
+		<li>
+			Persistent text-based discussions sponsored by the group
+			pertaining to their work
+			and intended to be referenceable by all group members,
+			as well as all official meeting minutes,
+			<em class=rfc2119>must</em> be archived by W3C for future reference.
+			This includes discussions conducted over email lists
+			or in issue-tracking services
+			or any equivalent fora.
+			For fora whose archive exists on a domain separate from its active operations,
+			the archive <em class=rfc2119>must</em> be available at the same level of confidentiality.
+
+		<li>
+			All tools and archives used by the group
+			for its discussions and recordkeeping
+			<em class=rfc2119>must</em> be documented
+			such that new participants and observers
+			can easily find the group’s tools and records.
+
+			Note: Community Groups producing material
+			intended to be adopted by a Working Group for standardization
+			should also adhere to these policies.
+	</ul>
+
+	Issue: Link to policy document with detailed recommendations / best practices / additional info.
 
 <h2 id="dissemination"> <span id="chapterDissemination"></span>
 Dissemination Policies</h2>
--- a/index.bs
+++ b/index.bs
@@ -1554,10 +1554,6 @@ Tooling</h3>
 
 	</ul>
 
-	Note: Community Groups producing material
-	intended to be adopted by a Working Group for standardization
-	should also adhere to these policies.
-
 	Issue: Link to policy document with detailed recommendations / best practices / additional info.
 
 <h2 id="dissemination"> <span id="chapterDissemination"></span>
--- a/index.bs
+++ b/index.bs
@@ -1501,13 +1501,43 @@ Tooling</h3>
 
 	<ul>
 		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			(i.e. intended for use or reference outside its own membership)
+			<em class=rfc2119>must</em> be published and promoted at a W3C-controlled URL,
+			and backed up by W3C systems
+			such that if the underlying service is discontinued,
+			W3C can continue to serve such content without breaking incoming links
+			or other key functionality.
+
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			<em class=rfc2119>must</em> follow best practices for accessibility worldwide
+			and for persons with disabilities.
+			Network access to w3.org itself may be assumed.
+
+		<li>
+			Persistent text-based discussions sponsored by the group
+			pertaining to their work
+			and intended to be referenceable by all group members,
+			including official meeting minutes and especially records of decisions made,
+			<em class=rfc2119>must</em> be archived by W3C for future reference.
+			This includes discussions conducted over email lists
+			or in issue-tracking services
+			or any equivalent fora.
+			Any discussion system whose archive is separate from its active interactive operation
+			<em class=rfc2119>must</em> make that archive available
+			at the same level of confidentiality.
+
+		<li>
 			Any tooling used by the group
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>must</em> be usable
-			without cost
+			without cost (direct or indirect)
 			by all active or would-be active group participants
-			to allow their equal participation
+			to allow their effective participation
 			regardless of disability
 			or geographical location.
 
@@ -1516,45 +1546,17 @@ Tooling</h3>
 			or operate some workaround.
 
 		<li>
-			All reports, publications, or other deliverables
-			produced by the group for public consumption
-			(i.e. intended for use or reference outside its own membership)
-			<em class=rfc2119>must</em> be published on a W3C-controlled domain
-			and backed up by W3C systems
-			such that if the underlying service is discontinued,
-			W3C can continue to serve such content without breaking links
-			or other key functionality.
-
-		<li>
-			All reports, publications, or other deliverables
-			produced by the group for public consumption
-			<em class=rfc2119>must</em> be accessible worldwide
-			and to persons with disabilities.
-			Network access to w3.org itself may be assumed.
-
-		<li>
-			Persistent text-based discussions sponsored by the group
-			pertaining to their work
-			and intended to be referenceable by all group members,
-			as well as all official meeting minutes,
-			<em class=rfc2119>must</em> be archived by W3C for future reference.
-			This includes discussions conducted over email lists
-			or in issue-tracking services
-			or any equivalent forum.
-			For fora whose archive exists on a domain separate from its active operations,
-			the archive <em class=rfc2119>must</em> be available at the same level of confidentiality.
-
-		<li>
 			All tools and archives used by the group
 			for its discussions and recordkeeping
 			<em class=rfc2119>must</em> be documented
 			such that new participants and observers
 			can easily find the group’s tools and records.
 
-			Note: Community Groups producing material
-			intended to be adopted by a Working Group for standardization
-			should also adhere to these policies.
 	</ul>
+
+	Note: Community Groups producing material
+	intended to be adopted by a Working Group for standardization
+	should also adhere to these policies.
 
 	Issue: Link to policy document with detailed recommendations / best practices / additional info.
 
--- a/index.bs
+++ b/index.bs
@@ -1530,6 +1530,9 @@ Tooling</h3>
 			<em class=rfc2119>must</em> make that archive available
 			at the same level of confidentiality.
 
+			Note: The lack, or loss, of such archives does not by itself
+			invalidate an otherwise-valid decision.
+
 		<li>
 			Any tooling used by the group
 			for producing its documentation and deliverables
@@ -1553,6 +1556,9 @@ Tooling</h3>
 			can easily find the group’s tools and records.
 
 	</ul>
+
+	The Team is responsible for ensuring adherence to these rules
+	and for bringing any group not in compliance into compliance.
 
 	Issue: Link to policy document with detailed recommendations / best practices / additional info.
 
--- a/index.bs
+++ b/index.bs
@@ -1504,7 +1504,7 @@ Tooling</h3>
 			All reports, publications, or other deliverables
 			produced by the group for public consumption
 			(i.e. intended for use or reference outside its own membership)
-			<em class=rfc2119>must</em> be published and promoted at a W3C-controlled URL,
+			<em class=rfc2119>should</em> be published and promoted at a W3C-controlled URL,
 			and backed up by W3C systems
 			such that if the underlying service is discontinued,
 			W3C can continue to serve such content without breaking incoming links
@@ -1513,16 +1513,16 @@ Tooling</h3>
 		<li>
 			All reports, publications, or other deliverables
 			produced by the group for public consumption
-			<em class=rfc2119>must</em> follow best practices for accessibility worldwide
+			<em class=rfc2119>should</em> follow best practices for accessibility worldwide
 			and for persons with disabilities.
 			Network access to w3.org itself may be assumed.
 
 		<li>
 			Persistent text-based discussions sponsored by the group
 			pertaining to their work
-			and intended to be referenceable by all group members,
-			including official meeting minutes and especially records of decisions made,
-			<em class=rfc2119>must</em> be archived by W3C for future reference.
+			and intended to be referenceable by all group members <em class=rfc2119>should</em>--
+			and official meeting minutes and especially records of decisions made <em class=rfc2119>must</em>--
+			be archived by W3C for future reference.
 			This includes discussions conducted over email lists
 			or in issue-tracking services
 			or any equivalent fora.
@@ -1534,7 +1534,7 @@ Tooling</h3>
 			Any tooling used by the group
 			for producing its documentation and deliverables
 			or for official group discussions
-			<em class=rfc2119>must</em> be usable
+			<em class=rfc2119>should</em> be usable
 			without additional cost
 			by all who wish to participate,
 			to allow their effective participation
@@ -1548,7 +1548,7 @@ Tooling</h3>
 		<li>
 			All tools and archives used by the group
 			for its discussions and recordkeeping
-			<em class=rfc2119>must</em> be documented
+			<em class=rfc2119>should</em> be documented
 			such that new participants and observers
 			can easily find the group’s tools and records.
 
--- a/index.bs
+++ b/index.bs
@@ -1518,11 +1518,13 @@ Tooling</h3>
 			Network access to w3.org itself may be assumed.
 
 		<li>
-			Persistent text-based discussions sponsored by the group
+			Official meeting minutes and especially records of decisions made <em class=rfc2119>must</em>
+			be archived by W3C for future reference;
+			and other persistent text-based discussions
+			sponsored by the group,
 			pertaining to their work
-			and intended to be referenceable by all group members <em class=rfc2119>should</em>--
-			and official meeting minutes and especially records of decisions made <em class=rfc2119>must</em>--
-			be archived by W3C for future reference.
+			and intended to be referenceable by all group members
+			<em class=rfc2119>should</em> be.
 			This includes discussions conducted over email lists
 			or in issue-tracking services
 			or any equivalent fora.
--- a/index.bs
+++ b/index.bs
@@ -1530,6 +1530,7 @@ Tooling</h3>
 			produced by the group for public consumption
 			<em class=rfc2119>must</em> be accessible worldwide
 			and to persons with disabilities.
+			Network access to w3.org itself may be assumed.
 
 		<li>
 			Persistent text-based discussions sponsored by the group


dwsinger:
I'm tempted to generalize what Tantek says: that in general we can presume that the world-wide-web is a world-wide-tool, not just w3.org, though it's worth calling w3.org out explicitly. (This is closer to Sam's position, I think).

I'm trying, until we get through AB/team, review, to keep the big picture in one place on the Wiki, so I took the liberty of adding Tantek's point to the [General](https://www.w3.org/wiki/W3C_Tooling_Policy#General_Policy_.28all_tools.29) section. Feel free to adjust.

plehegar:
one clarification question: Does "<em class=rfc2119>must</em> be published on a W3C-controlled domain" exclude and w3c.github.io (and related)? If yes, does it mean that all editor's drafts must be replicated. Since we don't check the content, we'd probably have to isolate them from www.w3.org to avoid same-origin leaks) ? Same question for drafts.csswg.org . cc @vivienlacourba .

dwsinger:
> one clarification question: Does "_must_ be published on a W3C-controlled domain" exclude and w3c.github.io (and related)? If yes, does it mean that all editor's drafts must be replicated. Since we don't check the content, we'd probably have to isolate them from [www.w3.org](http://www.w3.org) to avoid same-origin leaks) ? Same question for drafts.csswg.org . cc @vivienlacourba .

No, because the clause applies to final deliverables only: "All reports, publications, or other deliverables produced by the group for public consumption" not work-in-progress.  

Also, the phrase is ambiguous; I think the requirement is that they be addressed by w3.org URLs and backed up by the W3C, so that we could, if needed, host them elsewhere (e.g. natively) and retain the addressing, if the selected service goes away or is no longer appropriate for use.

Finally, this doesn't appear this way in the wiki, which tries to keep the big picture in one place. This PR is head of the game.

fantasai:
> one clarification question: Does "must be published on a W3C-controlled domain" exclude and w3c.github.io (and related)? ... Same question for drafts.csswg.org . cc @vivienlacourba 

@plehegar Yes, it excludes w3c.github.io and it would exclude csswg.org unless someone at W3C has edit access to the DNS records such that W3C can take over hosting if the current provider shuts down.

>  If yes, does it mean that all editor's drafts must be replicated. Since we don't check the content, we'd probably have to isolate them from www.w3.org to avoid same-origin leaks) ?

I think that depends on whether the editors' drafts are being used as internal scratch space (equivalent to comments in an issue tracker) or if they are being widely referenced. The Process 2020 changes were intended to help us shift us back to using w3.org/TR for referencing of our specs, by making it possible to keep those documents in sync with the ED sources. But e.g. the TAG and i18n have a number of other documents hosted on github.io and are using those URLs to refer people outside the group to those documents. These would be non-compliant.

I know that right now it seems like github.io will never shut down... but I'm sure the OpenType folks at ISO thought Yahoo would keep its groups hosting infrastructure up forever, and they just shut down in January; we had 30 days to rescue the ML archives, and several of them were corrupted. It is irresponsible for us to rely on external services without the ability to handle their demise.

fantasai:
@tantek

* Wrt tracking/cryptomining, I agree that's an unacceptable cost, but I think that gets a bit too detailed and should go into an accompanying policy document.
* Wrt exempting country-blocks against w3.org itself, yes, I agree we should assume access to w3.org.
* Wrt exempting github.io, I disagree, see my reply to plh in https://github.com/w3c/w3process/pull/436#issuecomment-683105398 If we need to set up some additional infrastructure at w3.org to easily proxy and back up content from github.io, that's fine, but using github.io URLs for broadcast content does not provide the level of future-proofing that we should expect of our output.

frivoal:
About the ban on crypto-mining, tracking, etc, I think it would be good to find a brief and high level way to state it in the Process itself. I feel that details about how that principle would speficically apply to crypto-mining, tracking, embedded advertisements, etc, do belong in a companion/explainer/policy document that's not as formal as the Process, but the abstract principle feels it would belong, as I don't think this is currently implied by any of the things listed in this PR.

dwsinger:
I added "must not cause members any direct costs (pay for use, etc.); and should not incur cost, tracking, or other client computation requirements unrelated to the direct use of the tool;"

to the [Wiki](https://www.w3.org/wiki/W3C_Tooling_Policy)

chaals:
How does this proposed policy apply to ReCaptcha, which uses people's work to build a particular company's competitive advantage over rivals?

frivoal:
@chaals I'd say something like reCaptcha would not be acceptable on specs and other deliverables, archives, or any kind of output we refer people to. As for whether it is OK on interactive tools we use while working (as opposed to outputs), I'd leave that outside of the bounds of the formal process, to Chairs discretion, possibly advised by a non-binding policy document. (Personal feeling on what that policy should be: it is a tolerable negative, better avoided if all other things are equal, but not that strong a concern).

TzviyaSiegman:
I recommend offering some guidance around accessibility instead saying must be accessible to people with disabilities. Perhaps instead recommend WCAG AA compliance or something similar. This allows for accessibility for those who might not technically have a disability but who have situational or temporary disabilities (e.g. broken arm, scratched cornea).

The phrase  "Note: If a new participant joins who cannot use the tool, this can require the Working Group to change its tooling or operate some workaround." puts the burden of shifting to a new tool on a new member who is unlikely to voice their needs. This is not acceptable.

chaals:
@TzviyaSiegman wrote:

> The phrase "Note: If a new participant joins who cannot use the tool, this can require the Working Group to change its tooling or operate some workaround." puts the burden of shifting to a new tool on a new member who is unlikely to voice their needs. This is not acceptable.

This is a vexed problem I think. 

I agree that expecting a newcomer to a Group to say "by the way, you use a tool that I cannot use effectively nor learn to use for some reason intrinsic to the tool" is a recipe for having people instead silently decide not to participate.

If there were more obviously great tools around, groups could simply choose tools that actually work for people and won't raise issues. But my observation of reality is that the available tools tend to have some advantages for a given set of users at the expense of not working for others. So we cannot expect Working Groups to pick a tool that won't cause problems, because there aren't any.

Making clear that Working Group tooling is temporary, chosen as a current answer to competing requirements and constraints, and MUST be reconsidered whenever a new participant joins may bring pushback from existing participants, because it potentially imposes a notable burden on them.

I think that the burden imposed is the inevitable price of ensuring a welcoming and inclusive environment for the diverse range of stakeholders whom W3C **wants** as participants.

If this solution, or something like it, is what we decide we want, it might be easier to successfully implement by holding substantial discussions among Working Groups and gaining broad acceptance of the idea before we consider making it something the Process imposes. 

That discussion is also a burden on Groups. I think asking them to consider what inclusion and diversity requires of them in practice is an imposition that Working Groups are likely to accept reasonably readily. If they don't think about it, and believe the Process is adding "bureaucratic busywork" without a good reason, they are more likely to simply work around or ignore what they consider a barrier to effective work.

frivoal:
> The phrase "Note: If a new participant joins who cannot use the tool, this can require the Working Group to change its tooling or operate some workaround." puts the burden of shifting to a new tool on a new member who is unlikely to voice their needs. This is not acceptable.

We may need to improve the phrasing, but putting the burden on the new members is not the intent of the text. The point isn't that we expect newcomers to need to ask for accommodations. Groups should pick tools that are newcomer friendly, without needing to be asked. Precisely what it means to pick good tools is a complex trade off between various dimensions of usability and accessibility, and seems more suited to guidelines than mandatory rules.

But if someone does ask, then the group must accommodate. Telling someone to go away because "we like it this way, too bad you cannot deal with it" is not acceptable. That's not sufficient for a good tooling policy, but it seems like a minimum requirement that can be elevated to MUST status.

dwsinger:
I agree with @frivoal . I don't think we can require groups to anticipate every need, or to avoid every possible pitfall — that's impractical. But I also see agree with @TzviyaSiegman ; imagine we had a participant who has a need that they consider somewhat private (e.g. "I have severe dyslexia and this tool doesn't permit my text-entry assistant to work"); this sentence asks them to break privacy. We can't require all groups to be aware of all tools and the populations that might not be able to use them. This does suggest, however, that maintaining a set of modern, approved, tools, that groups are encouraged to use, makes sense. How can we improve?

nigelmegitt:
> How can we improve?

The intro email sent to the new member when joining a Group can encourage the new member to signal to the Chairs/Team if they have any accessibility requirements, and begin an initially private conversation about whether the Group's tools meet their needs. This would put the onus on the Chairs/Team to understand the needs of the participants and to propose changes if needed, without new members feeling that they have to share all their requirements with the whole Group.

This review of needs could even be a private WBS poll (or similar) with limited read access to the results, and a _really_ neat thing would be if Chairs and Team have visibility of the results collated across the whole Group.

Members should be able to signal (privately, again) changes to their accessibility needs over time.

There should also be a way for prospective participants to check in also.

There should _not_ be a strong requirement to meet all the needs of all the members in every respect, because it may not be practical to do so. But there should be a commitment to get as close as possible.

This does leave room for more improvement though. It still feels not quite right, somehow: W3C members who are not members of a particular WG but who may want to join a meeting for some reason, for time to time, may not have their needs met. For example care would need to be taken in joint WG meetings.

dwsinger:
@nigelmegitt yes, I think we need to make it clear that the onus of making sure that the rules are followed is on the team contact and chair. Whether we need to get formal about it, or simply state it and allow it to be quoted when needed, I am not sure.

tantek (approved):
This is an improvement

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+		<li>
+			Any tooling used by the group
+			for producing its documentation and deliverables
+			or for official group discussions
+			<em class=rfc2119>must</em> be usable
+			without cost (direct or indirect)
+			by all active or would-be active group participants
+			to allow their effective participation
+			regardless of disability
+			or geographical location.
```
dontcallmedom, line 1514: I think a really hard question here is not so much about geographical location as much as administrative prohibitions of usage, many linked to governmental decisions (loosely bound to geographical locations), but some linked to very different authorities (e.g. organizational ban on certain tools).

I think the rule in the process needs to address that aspect more than geography (which as far as I know is rarely an issue); and the associated separate guidance should probably hints to how far a group needs to adapt the many edge cases that arise when applying the rule.

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+			Any discussion system whose archive is separate from its active interactive operation
+			<em class=rfc2119>must</em> make that archive available
+			at the same level of confidentiality.
+
+		<li>
+			Any tooling used by the group
+			for producing its documentation and deliverables
+			or for official group discussions
+			<em class=rfc2119>must</em> be usable
+			without cost (direct or indirect)
```
dontcallmedom, line 1510: "indirect" costs is a pretty sharp edge; an internet connection costs money, privacy-invasive (or so perceived) services can be considered costs to an individual, so can ads-based services. If there were specific indirect costs behind this, it might be better to list them explicitly than having a rule that could end paralyzing work.

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+		<li>
+			All tools and archives used by the group
+			for its discussions and recordkeeping
+			<em class=rfc2119>must</em> be documented
+			such that new participants and observers
+			can easily find the group’s tools and records.
+
+	</ul>
+
+	Note: Community Groups producing material
```
dontcallmedom, line 1529: The Process document never talks about Community Groups; this is an odd place to start doing so.

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+	to allow future participants and observers to understand the rationale and origins of current decisions,
+	and to guarantee long-lived access to its publications,
+	W3C requires that:
+
+	<ul>
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			(i.e. intended for use or reference outside its own membership)
+			<em class=rfc2119>must</em> be published and promoted at a W3C-controlled URL,
```
dontcallmedom, line 1479: in general, *must* statements benefit from being in the active form and applied to an agent - I don't know how systematically the Process Document follows that practice (although from a cursory glance, it does it often); I think it would be preferable to rephrase the MUST in this PR to say who (the group? the team? the chairs?) has what requirements .

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+			such that if the underlying service is discontinued,
+			W3C can continue to serve such content without breaking incoming links
+			or other key functionality.
+
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			<em class=rfc2119>must</em> follow best practices for accessibility worldwide
+			and for persons with disabilities.
+			Network access to w3.org itself may be assumed.
```
dontcallmedom, line 1490: the rule above is on "W3C-controlled URL" which is distinct from w3.org (and not just in theory); it may be useful to align the phrasing and the intent.

index.bs
dontcallmedom: on the specific W3C-controlled URL piece, it would probably be useful to evaluate how this rule would apply to Web Platform Tests (where a non-insignificant amount of W3C groups effort gets invested)

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+			<em class=rfc2119>must</em> make that archive available
+			at the same level of confidentiality.
+
+		<li>
+			Any tooling used by the group
+			for producing its documentation and deliverables
+			or for official group discussions
+			<em class=rfc2119>must</em> be usable
+			without cost (direct or indirect)
+			by all active or would-be active group participants
```
dontcallmedom, line 1511: is it clear what "would-be active" means? maybe a way to think about the phrasing is to figure out who needs to be explicitly consulted before a tool can be adopted; in the same line as above, another question is what recourse might such a would-be active participant have if the tool that is adopted doesn't follow that rule

index.bs
dwsinger: OK. Trying to quantify "indirect" in the process is probably an exhausting exercise. Maybe we should delete the parenthesis and allow anyone who thinks that there is a 'cost' to complain or push-back?

index.bs
dwsinger: agreed, we should put this in the Guide, alas.

index.bs
dwsinger: I agree in general, but it's a mix of the team contact working with the chair, editor(s), and systeam here. We could say that the [team contact | chair] must work with those people to ensure that …
?

index.bs
dwsinger: I think we were careful here. If a firewall or other device blocks access to w3.org, you can ignore that problem, is what we're saying. but if we re-direct, then you need to consider whether the target is reasonably reachable.

index.bs
dwsinger: The recourse in general is to complain, cite this rule, and escalate as needed. I agree, would-be active needs work.

index.bs
```diff
@@ -1457,6 +1459,79 @@ Resignation from a Group</h3>
...
+			<em class=rfc2119>must</em> be published and promoted at a W3C-controlled URL,
+			and backed up by W3C systems
+			such that if the underlying service is discontinued,
+			W3C can continue to serve such content without breaking incoming links
+			or other key functionality.
+
+		<li>
+			All reports, publications, or other deliverables
+			produced by the group for public consumption
+			<em class=rfc2119>must</em> follow best practices for accessibility worldwide
```
plehegar, line 1488: is there a link for what those best practices are?

dwsinger: WCAG? not sure what the equivalent is for i18n

index.bs
fantasai: I think we mainly care about geographical restrictions. I don't think we can mandate accommodating all of the IT teams of the world (though we can certainly take that into account voluntarily). But I think we should take into consideration the governmental bans, because there's no way practical around that.

index.bs
frivoal: Agree to delete that note.

index.bs
fantasai: I agree with dsinger, the subject in an active sentence of the same meaning would be too vague to state here. Also, although this is stated in the passive voice, the actor is clearly identified because it is not a stative verb but an active one: the actor is whoever is doing the publishing or promoting. I don't think we need a change here.

(We could say at the end of the list that the W3C Team MUST help our groups adhere to these rules; but I also think it's fine without such a statement.)

index.bs
fantasai: Would-be active is some actual person who would be active if the tooling barrier to participation were removed. It's intended not apply to a theoretical person or theoretical participation. If you've got a better way to phrase that...

index.bs
fantasai: W3C-controlled (as opposed to w3.org) was intentional. We want to allow other domains, provided systeam has the keys.

index.bs
frivoal: I think we should point to /Guide (which should point to WCAG and i18n equivalents) once that section of /Guide exists, and that we can leave this as is in the meanwhile.

index.bs
samuelweiler: And I think [we must not be held hostage to government bans](https://github.com/w3c/w3process/issues/435#issuecomment-672267652), lest [they become a means for a downgrade attack.](https://github.com/w3c/w3process/issues/435#issuecomment-717924343).

index.bs
```diff
@@ -973,15 +973,17 @@ Advisory Board and Technical Architecture Group Vacated Seats</h4>
...
-	These policies apply to <span id="participant">participants</span> in the following groups:
+	For the purposes of this Process, a <dfn>W3C Group</dfn> is one of W3C’s
+	<a href="#wgparticipant">Working Groups</a>,
+	<a href="#igparticipant">Interest Groups</a>,
 	<a href="#AC">Advisory Committee</a>,
 	<a href="#ABParticipation">Advisory Board</a>,
-	<a href="#tag-participation">TAG</a>,
-	<a href="#wgparticipant">Working Groups</a>,
-	and <a href="#igparticipant">Interest Groups</a>.
+	or <a href="#tag-participation">TAG</a>,
```
samuelweiler, line 981: Is the exclusion of Business Groups from this list intentional?

index.bs
samuelweiler: What does "and promoted" mean in this context?  (Is it unnecessary verbiage that can be cut?)

dwsinger: It means that if (as likely will happen) you get a proxy from a W3C URL to github.io, then you satisfy the "published at" (there is a w3c URL); but you must also give out the W3C URL and not the github.io one when you talk about it, at least outside your group.

chaals: "Groups must publish all content intended for public consumption at..."

The key being to know who was responsible if it went wrong, and therefore whom to talk to about fixing it.

index.bs
chaals: Not being held hostage to government bans is a nice idea. But for an organisation that works primarily with organisations who are subject to various local government rules, it makes no sense in practice.

We should pay attention to the concerns of actual people who are unable to use tools because their employer (or similar) will not allow it. The workarounds I have dealt with often involve having to set up a separate machine (which has a direct cost), an alternative email system which works poorly with the general W3C assumption that we can identify who represents what member, and other inconveniences that add up to a notable barrier to participation. For someone who is concerned their issues are likely to be very hard to move forward anyway, this is a pretty significant disincentive to participation.

index.bs
chaals: If a group decides to publish something elsewhere, that is an invalid decision because it falls outside the scope of what the group is authorised to do. The Staff contact is responsible for noticing that, the Chair is responsible for ensuring it doesn't happen.

index.bs
fantasai: Yes, because they are not governed by the this Process document.

index.bs
dwsinger: "all active participants, or those whose activity is curtailed because of a tooling choice."

index.bs
dontcallmedom: I think the "systeam has the keys" may hide quite a bit of complexity; I know from experience that some of the non w3.org servers don't have the same level of reliability, and I don't think there is much the systeam can do about it, nor is it clear they would know how to replace the said servers should that be needed.

(this may be fixable, and the details of fixing it lies outside of the process; but if we want to endorse that type of approaches, I think there needs to be more operational details as to what they entail)

fantasai: Sure, but details like that don't go into the Process. :) The key point is that systeam can take over hosting if it the hosting service goes down permanently.

dontcallmedom: Some recent events in IETF where "who owns the infrastructure" (beyond DNS) is having pretty consequential impact

dontcallmedom: https://mailarchive.ietf.org/arch/msg/wgchairs/J3RxuqR6bhvlmoNGwfuilMrVjmU/

index.bs
```diff
@@ -1457,6 +1459,81 @@ Resignation from a Group</h3>
...
+			<em class=rfc2119>should</em> follow best practices for accessibility worldwide
+			and for persons with disabilities.
+			Network access to w3.org itself may be assumed.
+
+		<li>
+			Persistent text-based discussions sponsored by the group
+			pertaining to their work
+			and intended to be referenceable by all group members <em class=rfc2119>should</em>--
+			and official meeting minutes and especially records of decisions made <em class=rfc2119>must</em>--
+			be archived by W3C for future reference.
```
dwsinger, line 1497: (I'd probably avoid the parenthetical – – by saying it this way:)

Official meeting minutes and especially records of decisions made <em class=rfc2119>must</em> be archived by W3C for future reference, and other persistent text-based discussions sponsored by the group pertaining to their work and intended to be referenceable by all group members <em class=rfc2119>should</em> be.
 			.

fantasai: Hmm, I'm not sure I like the missing predicate on the second clause better than the parenthetical. Although I'm totally fine with putting the MUST clause before the SHOULD one.

frivoal: Breaking the tie: I took in the proposed changed.

index.bs
frivoal: The view expressed there seems fine by me. If W3C controls the domain, it doesn't matter who operates the service under that domain, they merely operate it, not own it.

As for the bigger point about assuming network access to $FOO, I actually agree with @dontcallmedom. The rule, as phrased, would imply that if a country blocks both w3.org and W3C-controlled foo-drafts.org, we'd have to move the stuff from foo-drafts.org to w3.org, even though it doesn't make anything better. So let's align the phrasing to “W3C-conrolled URL”.

index.bs
frivoal: I agree with @chaals' first paragraph in the comment above.

As for the situation discussed in the second one, I think this is one where being accomodating is good, but I don't think we want to constrain it with RFC2119 language in the Process. It probably does make for a relevant discussion in /Guide though.

plehegar:
> Yes, it excludes w3c.github.io and it would exclude csswg.org unless someone at W3C has edit access to the DNS records such that W3C can take over hosting if the current provider shuts down.

2 points on this:
1. Groups will have a really hard time figuring why the policy is making a difference between w3c.github.io and github.com/w3c . If we're comfortable enough to share github.com/* , we ought to do the same for *.github.io, otherwise the policy will not make sense.
2. Editor's drafts are seen as being outside of the W3C Process so we'll get push back that the policy applies to those. If the W3C Process starts applying to them, we should engage with the Groups.

dontcallmedom:
re editors draft & `w3c.github.io`,  I think a good chunk of getting the policy applied would revolve around:
* providing a turn-key solution that groups can adopt for their github-maintained drafts
* making the editors tools (respec, bikeshed, specref) replace old URIs with new ones (semi-?)automatically

Once you have that, it's not clear groups would have any reason to push back

sideshowbarker:
It’s not clear me what actual problems would be solved in practice for the vast majority of our spec editors and for users of our specs by doing what seems to be proposed here — that is, by creating (I guess) some w3.org subdomain that editors are required to use for their current drafts — or, short of being required, something that we try to corral them into using.

I recognize there are some people other than the majority of our spec editors and actual users of our specs who’d apparently like to see us make some change here, but I think it’d be a very big mistake to optimize for the needs of those other people.

The current https://w3c.github.io/ way for delivering current spec drafts has basically been wildly successful — and is widely (nearly universally, in my experience) appreciated by spec editors, by implementors, by the key stakeholders such as the MDN maintainers, by web developers, and by pretty much everybody using the specs to solve an actual information need.

Maybe some people forget that we previously did have almost exactly what seems to be being proposed anew here. It was called https://dev.w3.org/. We maintained it for years at a considerable resource expense — including considerable costs of time for W3C staff (e.g., me) in maintaining it and working around (on our own) all its quirks. It was not fun and was highly suboptimal with respect to meeting the needs of our editors and the actual users of our specs

But eventually we had the good sense to dig ourselves out of that, and move to using GiHub and https://w3c.github.io/ — which has been immeasurably better at meeting the needs of our editors and the users of our specs, and significantly reduced the costs we were wasting on having W3C staff maintain something which, after the advent of GitHub, was essentially a (retroactive) re-inventing of the wheel, very poorly.

So I hope my lack of enthusiasm about the prospect of getting back into that business is understandable.

And along with all that, as @plehegar points out in https://github.com/w3c/w3process/pull/436#issuecomment-801883981, editor’s drafts have (thankfully so far) never been a formal part of the W3C Process. And trying to unilaterally make them part of the Process without getting very strong buy-in from all the editors doing the actual editing work — and all the working groups supporting that work — is not going to fly well.

And so I can imagine that the next logical place this discussion would necessarily take us is into is some proposal to update the Process document to make editor’s drafts a formal part of the process. And pursuing that would really complete the foundational hole-digging for ourselves that I think what’s being proposed here would end up leading us into.

sideshowbarker:
By the way, if part of what’s implied here is that we’d have a drafts.w3.org (or whatever) domain that mostly just proxied w3c.github.io, there are a lot of reasons why that’s not a good idea.

Among those reasons such proxying would be bad are: Consider what happens when some w3c.github.io/foo/ goes 404 (which is not hypothetical — it happened a lot recently when branches were switched over to `main` from `master`).

What would happen when a w3c.github.io URL went 404 is that our drafts.w3.org proxy would probably start showing users a 502 error — in which case, the users are reasonably going to assume something’s broken at the W3C, and start trying to figure how/who to report that to at the W3C. Which means the W3C and the W3C systems team now start having to deal with blame/fixing for problem they can’t actually solve.

Users who instead go to that w3c.github.io/foo/ URL directly still get a 404 of course — but that’s gonna be a lot more familiar to them than the arcane W3C 502 would be, and the users are much more likely to have encountered that error before with other projects hosted at GitHub using GitHub pages, and they’re much more likely to be able to figure out that they need to go raise a bug in the github.com/w3c/foo/ repo to report, Hey, your GitHub pages stuff seems to be broken right now.

And if the proposal here is to instead do have the drafts.w3.org domain be something more than just a w3c.github.io proxy, then like I said in my previous comment, we would be buying into a much bigger set of problems and permanent costs — that would amount to basically regressing back to the state of things in the bad old dev.w3.org days.

dwsinger:
@sideshowbarker I sort-of agree with you, in that I would like to get back to the state where groups maintain their working draft actively, and that that is the official w3.org-URL-addressed resource, and that editor's drafts have only slightly more status than anyone else's forks. The WG accepts changes from forks into their active WD.

I think that the changes we're making, to allow WGs to maintain their WD in the WG's repo and have a URL that proxies to that repo from w3.org should remove the barriers to and impediments that mean that groups are bad at maintaining their WD.

I note that it's a group decision over how closely they manage their editor and what the editor commits to the group's WD. In the early stages of a project, that can be quite loose, and might only need to tighten up as the project completes.

So, rather than spending our energies formalizing EDs, I'd rather spend them on getting WGs back to having a clear concept of the WG's working draft.

frivoal:
>The current https://w3c.github.io/ way for delivering current spec drafts has basically been wildly successful — and is widely (nearly universally, in my experience) appreciated by spec editors, by implementors, by the key stakeholders such as the MDN maintainers, by web developers, and by pretty much everybody using the specs to solve an actual information need.

> Maybe some people forget that we previously did have almost exactly what seems to be being proposed anew here. It was called https://dev.w3.org/. 

I think this is a misunderstanding of what is being proposed here. Using github.io as your tech stack is fine, and there is no suggestion to move away from it. If it works for you (and it works for almost everybody), great, no need to rehost into something similar to what dev.w3.org was.

However, using github.io **as your URL** is not fine for long lived documents / output of a working group. It's fine *for now*, but github is a commercial company with no obligation to keep the service available or free of charge.

So the request here is that any deliverable of the WG produced for other people's consumption, you may serve it from github.io if you want to, but if so, you need to get w3c to proxy it through some W3C controlled URL (presumably under w3.org), and refer people to that URL. This does not require any body to do anything different from today after an initial set up step for the repo.

Sure, that feels like a (tiny) hassle in the short term, because it doesn't make much of a difference to reading the document right now. However, if 5 years down the road, gihub.io discontinues its free service and all our links break, this will be a giant nuisance. Fixing internal links after the fact will be a massive pain. Fixing the links of everybody else in the world will be impossible. Even if we have archives of everything that was published there, we won't be able to restore the content in place unless we control the domain through which it is being accessed.

frivoal:
> So, rather than spending our energies formalizing EDs…

Nothing in the text is ED specific, or even mentions them. It's just being generic about documents "produced by the group for public consumption (i.e. intended for use or reference outside its own membership)". If you don't do that with your EDs, then the rule doesn't apply to them.

But for outputs of the group that are being "produced by the group for public consumption", the rule matters, regardless of whether it's an **** Draft, an explainer, a finding, a disposition of comments, an Implementation Report, an Article, a Best Practice… Cool URLs don't change, and you cannot guarantee that if you let a third party be in charge of your URLs.

sideshowbarker:
> I think this is a misunderstanding of what is being proposed here.

Actually, at this point I’m very clear on what exactly it being proposed here. And I think it’s a bad idea.
 
> However, using github.io **as your URL** is not fine for long lived documents

I recognize you’re asserting that’s not fine. But, with respect, that doesn’t make it so. It’s not axiomatic.

> output of a working group.

An editor’s draft is by definition not output of a working group. The current Process document intentionally says nothing about editor’s drafts. I recognize that this proposal wants to make the Process document say something about editor’s drafts — essentially, I guess, to redefine editor’s drafts as “output of a working group”. But that is not what the current Process document actually says.

We already have w3.org/TR URLs for all documents that the Process document actually recognizes as output of a working group. And those URLs already provide the persistence being sought for here. 

What instead seems to be getting proposed here is to to create URLs in some w3.org subdomain for documents that … what? That don’t have w3.org/TR URLs yet because they’re not yet recognized by the Process document as output of a working group? Or at some w3.org subdomain in addition to and redundant with the w3.org/TR URLs?

> So the request here is that any deliverable of the WG

Editor’s drafts are not deliverables of the WG. Documents published at w3.org/TR URLs are deliverables of the WG — at least per the Process document.

An editor’s draft is something that may have content which will eventually be published by the group as meeting its criteria for one of its chartered deliverables — after the working group makes a formal decision to publish it as such. But until then it’s simply an editor’s draft.

> produced for other people's consumption you may serve it from github.io if you want to, but if so, you need to get w3c to proxy it through some W3C controlled URL (presumably under w3.org), and refer people to that URL. This does not require any body to do anything different from today after an initial set up step for the repo.

In fact it does require multiple parties to do something very different from what they are doing today. The proposal here creates significant new work and new obligations for editors, for working-group chairs, and for W3C staff — both for working-group team contacts and for the W3C systems team. 

> Sure, that feels like a (tiny) hassle in the short term,

No, not just tiny, and not just short term. Knowing what I know intimately well from years of handling publication tasks for a variety of working groups, what’s being proposed here feels to me like something that is in practice going to end up being a very big hassle, and a very long-term one. The proposal would in practice create significant work, for all of our editors and chairs and team contacts — and for a long time, because enshrines it into the Process document forever.

And it will create all that new work without actually solving any problem that the vast majority of our spec editors and actual users of our specs recognize as a real problem — but instead will be trying to solve a problem that none of them are actually asking us to try to solve.

> because it doesn't make much of a difference to reading the document right now. However, if 5 years down the road, gihub.io discontinues its free service and all our links break, this will be a giant nuisance.

So to be clear: The problem this proposal is really trying to solve doesn’t actually exist today, nor is it anticipated that the problem will ever be a real problem any time soon — *if* it ever actually becomes a real problem at all.

And that is a big *if*. It’s by no means certain that the anticipated problem is ever going to actually occur. At least I’ve so far not seen anyone assert that it’s certainly going to happen. And in fact, I think very many or even most people might believe that anticipated problem is very unlikely to actually occur.

sideshowbarker:
> > So, rather than spending our energies formalizing EDs…
> 
> Nothing in the text is ED specific, or even mentions them. It's just being generic about documents "produced by the group for public consumption (i.e. intended for use or reference outside its own membership)". If you don't do that with your EDs, then the rule doesn't apply to them.

If the intent of the proposal isn’t to have the Process document require it for EDs, then the proposal needs to be updated to explicitly and unambiguously state that — *“This requirement does not apply to editor’s drafts”*, or something like that.

Or otherwise is there a serious assertion being made here that it only applies to EDs if the EDs are *“produced by the group for public consumption (i.e. intended for use or reference outside its own membership)”* — as if there actually are such EDs?

Are we aware of any actual cases of EDs not produced for public consumption and not intended for use or reference outside a group’s own membership?

It occurs to me now that the proposal seems for some reason to be intentionally written without explicit reference to EDs, but with the apparent intent that it will in practice be applied to all EDs — since all EDs are in practice *“produced for public consumption”*.

And if that’s the case, then it’s disingenuous to not be explicitly including EDs in the concrete text of the proposed Process document change. If it’s believed that the requirement will in practice actually be placed on EDs, then it seems to me you have an obligation in principle to update the language of the proposal to explicitly mention EDs — so that everybody reading the proposal can clearly understand what the actual effect of the change is going to be in practice.

dwsinger:
@sideshowbarker we're trying to deal with a simple problem. At the moment we have documents that purport to be W3C work that are referred to by the WG, and people outside the WG are asked to refer to, using URLs that are not W3C URLs and sometimes (often?) don't even have any mention of the W3C in their name. Since anyone can fork a W3C repo, and keep the W3C logo on the document, this is a strange place to be. So when someone sees a document at <randomstring>.github.io, what are they supposed to understand? That's it's a working fork of someone who wants to make a pull request.

Have a look at https://immersive-web.github.io/webxr/  The URL has nothing in it that assures the reader it's published by the W3C or has any status at all, so simply from an address point of view it's unclear. The status section says 

> This document was published by the Immersive Web Working Group as an Editors' Draft. This document is intended to become a W3C Recommendation. 

and 

> Publication as an Editors' Draft does not imply endorsement by the W3C Membership.

which are confusing. Working drafts are produced by Working groups; Editor's drafts are produced by Editors. Editor's drafts are not even endorsed by the Working Group. It's fine for a WG to decide to defer the update of the WD to the editor, of course (and catch 'in arrears' when the Editor makes a mistake), but that then is the WG's WD.

This is not an isolated example. We need to clean up. If a W3C group is maintaining/developing a document, and they want to give a reference to it to anyone else — inside the W3C, and particularly outside — then we ask that it be a W3C URL that they use. "Our current working draft is at <x>." "Our editor maintains a forward-looking document that contains matters for the WG to consider, but we don't look for group consensus on the editor's drafts." "We have a requirements document and roadmap at <y>."

Now, I'd also prefer that we separate ED and WD better conceptually, but that's separate.

frivoal:
> Actually, at this point I’m very clear on what exactly it being proposed here. And I think it’s a bad idea.

> I recognize that this proposal wants to make the Process document say something about editor’s drafts — essentially, I guess, to redefine editor’s drafts as “output of a working group”.

No, it does not want to do this.

> if the intent of the proposal isn’t to have the Process document require it for EDs, then the proposal needs to be updated to explicitly and unambiguously state that — “This requirement does not apply to editor’s drafts”, or something like that.

The intent is that it neither applies to EDs on virtue of them being EDs, nor does it exempt them on virtue of being EDs.

It depends on what the working group does with them. If they are the documents that the working group promotes to the general public, that the working group sends for TAG review or horizontal review, or that sort of things, then the working group is treating them as its output, and the rule applies to them. If the WG does all these things using the TR publication, and the ED just happens to be a publicly visible scratch space that the WG does not promote as the version of the work people should look at, then it does not apply.

The text doesn't single out EDs. That's not out of a desire to mislead, but to be generic. Things that are published on TR satisfy these criteria automatically. If individual WG participants want to publish whatever wherever, they may do so. But if the WG collectively produces something that it wants the world to look at other than the things it puts on TR, whatever these things are, they need to have stable URLs, which can only be guaranteed if they are w3c controlled. "whatever these things are" would cover EDs to the extent the group itself chooses to treat the ED as the document it promotes, and it would cover things like [implementation reports](https://drafts.csswg.org/css-contain-1/implementation-report-2019-09), [Disposition of Comments](https://drafts.csswg.org/css-writing-modes-3/issues-cr-2019), [Explainers](https://drafts.csswg.org/css-nav-1/explainer), [Articles](https://www.w3.org/International/articlelist)… 


> Knowing what I know intimately well from years of handling publication tasks for a variety of working groups, what’s being proposed here feels to me like something that is in practice going to end up being a very big hassle, and a very long-term one. The proposal would in practice create significant work, for all of our editors and chairs and team contacts — and for a long time

Really? Have a look at https://drafts.csswg.org/ All it takes for this large pile of documents to comply with the proposed rule is for the person maintaining that server to give admin access to the DNS records to the w3c team, which I believe has been done already (and if not, is trivially fixable). How does that create "significant work for all our editors and chairs and team contacts — and for a long time"?

As for github.io specifically, a minimal but sufficient solution would be to build a form where you can put a github.io/foobar url, which sets up (automatically, except for a Team approval step) a foobar.drafts.w3.org proxy, then we're done. (And all that needs creating is this set up form, we already have the proxy.) How does that create "significant work for all our editors and chairs and team contacts — and for a long time"?

> So to be clear: The problem this proposal is really trying to solve doesn’t actually exist today, nor is it anticipated that the problem will ever be a real problem any time soon — if it ever actually becomes a real problem at all.

The time to put your seat-belt on is not after the accident has started.

github.io, or any other third party service, cannot be depend on to be available and free forever. Once upon a time, it would have seemed outlandish and far fetched to suggest that Yahoo Groups would ever shut down. A great service, offered by a company widely regarded (at the time) as one of the pillars of the web, relied upon my so many. Of course it'll stay up. That's a perfectly reasonable place to do standardization work, say about fonts, under the MPEG group: https://groups.yahoo.com/neo/groups/mpeg-OTspec/info Oops, it's gone. What about Google Code? That's a great service used dependably by many and offered by a very respectable company… Except it's gone too. Well, maybe, but surely github is different, and they wouldn't [discontinue popular free services](https://docs.github.com/en/enterprise-server@2.21/developers/overview/replacing-github-services)…

Just because a service is popular today does not force the service provider to keep providing it forever. For things that are short lived, whatever. For things that are supposed to be referenceable in the long run, it's not fine that

dontcallmedom:
I wonder if instead of focusing on rules for inclusion in the Process 2021, we should focus on the underlying goals we want to achieve and work toward getting these goals implemented (e.g. by the end of 2021). Based on the lessons from trying to achieve these goals, we may then look into encoding what is confirmed as necessary rules.

In other words, it feels we lack implementation experience and enforcing rules in these conditions is likely to backfire. Once it's clear the goals we have in mind are correct and implementable, it might be easier to see where rules are necessary to avoid regressing.

dwsinger:
@dontcallmedom That's the conversation that we had over the last year, and these rules are intended to be the basic principles indeed. 

The only one people seem to be struggling with is "use W3C URLs to reference W3C documents that are referenced outside their group". The obvious two reasons for that are (a) we might change the service we use, either generally, or for specific cases; Github might change in a way we don't like, for example; (b) to establish branding, and make clear the difference in status between a W3C document and a fork. These seem pretty uncontroversial to me.

dontcallmedom:
To be clear, I'm not pushing back on the quality of the conversations, and I think the basic principles identified sound reasonable; but before enshrining them in the process (which is hard & slow to change), I think we should learn from adopting these principles and delay putting them in the Process Document until they've been tested in the wide diversity of W3C groups. 

In terms of adoption too, it feels much better to have groups do their part in building the policy by testing & implementing it, rather than having it mandated suddenly. I guess I'm advocating for an incubation approach here…

fantasai:
@dontcallmedom That's why they're written with SHOULD in almost all cases. Putting them in the Process makes sure that we are absolutely clear about what we're aiming for, and that we're serious about it and expect all groups to make an effort towards conformance. But we have leeway for time to get there and for figuring things out. If the rules need adaptation after experience, then we can adjust them next year: the Process isn't set in stone. Consider the SHOULD as a CR phase: we're done with the design phase, and now we're asking for implementation.

As for incubation, we already have groups who are largely in conformance with these requirements. We are past incubation. This is Call for Implementations time.

index.bs
mallory: I agree with Sam for a couple of reasons. First if your tool choice is determined by how authoritarian censors behave then you're at the whim of dictatorial decision making. Second you might be inadvertently trading off better access for some over others. There will be those in places where blocking and filtering uses one tactic (TLS 1.3 blocking, for example), and those in places where blocking and filtering uses another tactic that the same protocol makes the tool resistant to censorship (DNS/domain blocking, respectively in this example).

Suggest thinking of proactive anti-censorship mechanisms that provide blanket privacy, and therefore censorship resistance, to the most people, like tools that provide an onion service, or actually the W3C participating in the Tor network through bridges and relays (if it doesn't already).

dwsinger: It's phrased as a 'should' so that we (a) can think about and examine hard cases and (b) decide, when appropriate, not to follow it. So there may be cases where we regretfully decide that some members will have problems, but it's the right choice. But that should be a conscious decision; we do not want the case where some members can't access, and no-one cares or has thought about it.

Issue #393, opened 2020-02-28
[meta] Eliminate role of The Director and redistribute authority from The Team to W3C members & community


tantek:
This is a meta issue following-up from participating in the [AB meeting on Tuesday](https://www.w3.org/wiki/AB_Agenda_Wiki_Feb2020#T1500) with the overall goal of eliminating the role of The Director from [the Process](https://www.w3.org/2019/Process-20190301/) (as an individual, delegated, or renamed) and redistribute authority from The Team in the process to W3C members &amp; community. These changes will likely be incremental, and require additional significant changes (e.g. considering elimination of Formal Objections and their processes instead of [#331](https://github.com/w3c/w3process/issues/331)).

There are several existing issues that cover aspects of this issue. Here are a few of them:

  * [#392](https://github.com/w3c/w3process/issues/392)
  * [#360](https://github.com/w3c/w3process/issues/360)
  * [#316](https://github.com/w3c/w3process/issues/316)
  * [#223](https://github.com/w3c/w3process/issues/223)

Additional follow-up issues will be filed for specific changes and sets of related changes.

(Originally published at: https://tantek.com/2020/058/b1/redistribute-authority-w3c-members-community)

michaelchampion:
@tantek, let me ask a clarifying question.  There's been rough consensus for awhile now that the roles the Director fills now should be transferred to others.  And I assume you're not saying that the totally routine roles nominally performed by the Director (such as announcing things) shouldn't be assigned to the CEO or Team, right?

What I think you're saying is that rather than giving someone else the real decision making roles (such as those @mnot identified in #316  or I identified in #392) that the Director held, the process should be changed to make those collective decisions of the community or membership rather than inventing a new body to decide them or give the Team that power.  For example, if the Process eliminated Formal Objections, there would be no need for a Council or Technical Director or whatever to adjudicate them.  Instead, have the community or membership decide whether to create a WG, advance a spec to Rec, approve a MoU, etc.  by voting or some other mechanism to determine rough consensus or the preponderance of opinion.

Is that about right?

dwsinger:
I believe that this _is_ the plan of record: re-assign to the team only 'routine' operations that they handle today, and that all questions of judgment and authority need to be re-architected.

So is the new thing about this is that this should be done incrementally?

michaelchampion:
> So is the new thing about this is that this should be done incrementally?

[...]

PR #438, opened 2020-08-13
Define "CEO" and "Team Decision"

frivoal:
One thing that is annoying from an editorial standpoint when working on director-free related issues and pull requests is that even though we may want to reassign some decisions/responsibilities/actions to the CEO and some to the Team, "CEO" and "Team decision" aren't currently defined terms on the main branch, so I can't use them with a bikeshed cross-link in any pull request based on the main branch.

This pull request formally defines these terms, so that they can be used as a basis for further work. It does not by itself make any change to who decides what. Based on discussions held in the AB, including at the Quebec F2F, this also includes a Note that clarifies the relationship between the Team and the CEO/Director.

I've had this written up for a while on a separate branch (and it's been part of the [director-free consolidated branch](https://github.com/w3c/w3process/tree/director-free)), with all my other director-free topic branches being based off that, but branches on top of branches is annoying to maintain.

Since this is only defining terms and adding a note that should already be true today, this is editorial, and I would appreciate if we could merge this (or something similar) now, so that I can spend more of my "director free" time on the actual issues and less on rebasing branches.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -526,10 +526,16 @@ The W3C Team</h3>
 	<dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
 	even when they are carried out by other members of the [=Team=].
 
+	<div class=issue>
+	It's bee suggested that the following note may be helpful,
+	but <a href="https://github.com/w3c/w3process/pull/438#issuecomment-693001060">this remains controversial</a>.
+	Should we include it?
+
 	Note: It is generally expected
 	that the CEO and Director will take the Team's input into consideration when making decisions,
 	and that they will delegate various responsibilities to individuals or groups within the Team.
 	Nevertheless, the CEO and Director have the ultimate authority within the Team.
+	</div>
 
 	The <dfn id="def-Director" export lt="Director|W3C Director">Director</dfn> is the lead technical architect at W3C,
 	whose responsibilities are identified throughout this document in relevant places.
--- a/index.bs
+++ b/index.bs
@@ -523,8 +523,7 @@ The W3C Team</h3>
 	(generally to other individuals in the Team)
 	for any of their roles described in this document,
 	except <a href="#tag-participation">participation in the TAG</a>.</p>
-	Decisions being described in this document
-	as being <dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
+	<dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
 	even when they are carried out by other members of the [=Team=].
 
 	Note: This does not imply that the CEO or Director are expected to make all decisions personally,
--- a/index.bs
+++ b/index.bs
@@ -507,7 +507,7 @@ The W3C Team</h3>
 
 	The <dfn export lt="Team | W3C Team">Team</dfn> consists of
 	the Director,
-	CEO,
+	<dfn>CEO</dfn>,
 	W3C paid staff,
 	unpaid interns,
 	and W3C Fellows.
--- a/index.bs
+++ b/index.bs
@@ -523,6 +523,9 @@ The W3C Team</h3>
 	(generally to other individuals in the Team)
 	for any of their roles described in this document,
 	except <a href="#tag-participation">participation in the TAG</a>.</p>
+	Decisions being described in this document
+	as being <dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
+	even when they are carried out by other members of the [=Team=].
 
 	The <dfn id="def-Director" export lt="Director|W3C Director">Director</dfn> is the lead technical architect at W3C,
 	whose responsibilities are identified throughout this document in relevant places.
--- a/index.bs
+++ b/index.bs
@@ -526,12 +526,9 @@ The W3C Team</h3>
 	<dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
 	even when they are carried out by other members of the [=Team=].
 
-	Note: This does not imply that the CEO or Director are expected to make all decisions personally,
-	and that the Team merely executes.
-	It is generally expected
-	that the CEO and Director will take the Team's input into consideration when making decisions personally,
-	and that they will delegate various responsibilities to individuals or groups within the Team,
-	as well as define the mechanisms by which this delegation is managed.
+	Note: It is generally expected
+	that the CEO and Director will take the Team's input into consideration when making decisions,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -523,6 +523,17 @@ The W3C Team</h3>
 	(generally to other individuals in the Team)
 	for any of their roles described in this document,
 	except <a href="#tag-participation">participation in the TAG</a>.</p>
+	Decisions being described in this document
```
chaals, line 526: s/Decisions being described in this document as being//

It's unnecessary wordiness

index.bs
```diff
@@ -523,6 +523,17 @@ The W3C Team</h3>
...
+	as being <dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority,
+	even when they are carried out by other members of the [=Team=].
+
+	Note: This does not imply that the CEO or Director are expected to make all decisions personally,
```
chaals, line 530: This note seems redundant with the paragraph just above.

index.bs
frivoal: Fair enough. I'll take it out unless some third voice argues otherwise.

index.bs
frivoal: Well, it's a note, so it's good that it doesn't contradict the normative text, nor adds requirements that aren't otherwise there. So, it's not entirely wrong to describe it as redundant, but I see it as clarifying intent.

Before there was this note, there were concerns (at least from @swickr), that the text above, without a note somewhere along these lines, could be read as a step away from the current culture, and as attempting to make the team much more top-down than it currently is.

wseltzer:
The Process should not describe the internal workings of the Team.

index.bs
jwrosewell: How about -

Note: The CEO and Director will take the Team's input into consideration when making decisions. They will delegate responsibilities to Team individuals or groups and define the mechanisms by which this delegation is managed. At all times the CEO and Director have the ultimate authority within the Team.

jeffjaffe: As a practical matter, the CEO rarely makes decisions, and almost always decides via a W3M consensus.  The language of the pull request is a bit strange in that it omits W3M.  One approach is a shorter note which merely says "typically, the CEO and Director involve Team members - specifically W3M - for input and internal Team consensus".

chaals: W3M has never been defined in the process, and I don't think it should be (in line with the comment below from @wseltzer).

The CEO is responsible for the decision - and I suspect is the one who determines when there is a W3M consensus. In which case it makes much more sense that the process states the CEO is responsible for the decisions.

dwsinger:
I think we're converging on simply defining what a Team Decision *is*, and similarly that the CEO is in charge of the team, and dropping the note.  Does this clause really _define_ the term _CEO_?

frivoal:
> The Process should not describe the internal workings/management of the Team.

No, and this proposed text doesn't do that. It says that the Team reports to the CEO (and Director while we have one), but that things can be delegated. Describing in what way delegation is managed would be totally wrong for the Process. That's why, among other things, I don't want to mention W3M directly.

The reason for this text being written the way it is:
* The one hand, we did not want to imply that the Process saying "The Team does/decides xyz" was granting Team members independent authority and an ability to ignore the wishes of the CEO on such topics, should there be disagreement.
* On the other hand, a statement clarifying that, but implying that the Team fully runs on top-down authority with no concern for Team member's opinions isn't something we wanted to say either.

I'd like to remind people that this text isn't something I just came up with a couple of weeks ago when making this pull request, but was the subject of face to face discussion of the AB, with input from the team, and part of the director-free branch for well over a year. If we can find better text, great.

But maintaining branches of branches in git and re-basing a whole tree of changes every few months because we can't agree to say that the Team reports to the CEO without being drones is frustrating.

frivoal:
> I would rather leave in undesired references to "The Director" as thorns in our sides to focus on solving the hard problems, than add any kind of stop-gap or reformist reforms that introduce (or add more references/definitions to) CEO and/or Team as explicit authorities.

This pull request does no such thing. All mentions of the Director stay as is.

Now, this pull request does make it easier in future pull requests to remove the Director from some roles, and replace them with the CEO or the Team. But of course those pull requests should be discussed on their own merits. Gradually reducing the involvement of the Director, when we have consensus over how to do it on a case by case basis, seems like the right thing to do for me. Rejecting any change until we have figured out everything seems like a good way to deadlock to me.

For instance, moving disciplinary actions to the CEO isn't "a stop gap reformist reform until we solve the hard problem", it's the right thing to do (haven't heard anyone say anything to the contrary at least). Delaying the ability to write a Pull Request proposing things this because we don't want to attribute any authority to the CEO or Team until we've found answers to all the roles of the Director isn't productive.

Unless know for sure that we intend to write a process under which neither the Team nor the CEO will have any authority over anything, having words to describe what they do merely facilitates us the discussion.

jeffjaffe:
Looking at the diversity of opinion on this topic - it seems that much of this discussion can move to the guide.  We've recently done that to clarify delegation of disciplinary action.

dwsinger:
Ok, the two changes

a) change CEO to <definition>CEO</definition>

b) add 

<dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority, even when they are carried out by other members of the [=Team=].

are fine since they don't change anything, make the editor's job easier, and give us a platform for referring to the CEO and team decisions.

I think we can and should make these changes and not insert the note.

fantasai:
> I think we can and should make these changes and not insert the note.

I would like to take @swickr’s concerns into account, and that means keeping some form of the note. That said, I think we can strip about half of the words out and still convey the same, thing: 

> Note: It is generally expected that the CEO and Director will take the Team's input into consideration when making decisions personally, and that they will delegate various responsibilities to individuals or groups within the Team. Nevertheless, the CEO and Director have the ultimate authority within the Team.

tantek (changes requested):
Leveling this up a bit, per #393 our goal should be that running the process does not require the CEO. Thus if anything we should remove references to the CEO, not define them / make it easier to add more references.
References to the CEO may be appropriate for the member agreement, but not the process IMO.

dwsinger:
Well, there are certain aspects of the Process that we agreed (fairly recently) are CEO decisions, notably the CEO is the decider for discipline-related questions. Don't forget the process is about how we work overall, not just how we produce recs.

dwsinger:
Proposed conclusion:

a) change `CEO` to `<dfn>CEO</dfn>`

b) add 

`<dfn noexport lt="Team Decision | Team's Decision">Team Decisions</dfn> derive from the [=Director=] and [=CEO=]'s authority, even when they are carried out by other members of the [=Team=]`.

are fine since they don't change anything, make the editor's job easier, and give us a platform for referring to the CEO and team decisions.

c) Add the note:

> Note: It is generally expected that the CEO and Director will take the Team's input into consideration when making decisions, and that they will delegate various responsibilities to individuals or groups within the Team. Nevertheless, the CEO and Director have the ultimate authority within the Team.

wseltzer:
still prefer to delete the note.

PR #444, opened 2020-08-14
Editorial tweak in Superseded Recommendation note

tantek:
use "previous" instead of "old" to be more specific, and fix typo: s/or a by a/or by a


frivoal:
@dwsinger This is very clearly editorial, I don't believe it needs group discussion, so I'm using editor discretion and merging in this PR. If you think it did warrant time on the call, feel free to schedule it.

Issue #449, opened 2020-09-01
Rename Candidate Change / Proposed Change to Candidate Amendment / Proposed Amendment


fantasai:
Apologies for the late editorial request, but: I'm working on the TR stylesheets and documentation related to the amendment process we introduced for RECs in Process 2020, and... my conclusion working on this is that "change" is such a generic term that is so frequently used in our day-to-day processes to refer to much more informal things, that it would be less confusing if we named these very formal things to use a different term.

Since we've been calling it an amendment process, I suggest calling them amendments. (It also sounds more formal, which is appropriate given what they are.)

I hope we can just fold this into P2020 as an editorial change...

chaals:
This seems like a trivial editorial change we should be able to adopt as proposed.

frivoal:
I support this, but only if we drop Ammended Recs as proposed in https://github.com/w3c/w3process/pull/428 (which I also support). Doing this without that would be confusing. Doing both is best.

jeffjaffe:
Retiring Amended Recs feels like a substantive change (and hence would need to wait for P2021), whereas I believe @fantasai was hoping that this change could be folded into P2020.

frivoal:
I don't mind if they're implemented somewhat out of sync. Temporary confusion, especially over a term that isn't actually used in practice, seems ok to me. As long as we don't get stuck into this long term.

fantasai:
@frivoal Worst case we can rename Amended Rec to something else, seeing as it's never been used anyway.

frivoal:
not my favorite outcome, but not something I'd oppose.

dwsinger:
Yes, I support calling these amendments. I don't think we can simultaneously clean up Amended Recs, as others have noted, alas. But since they are not used, they shouldn't be too confusing; perhaps a Note "Amended Recommendations should not be confused with the process of Amending a Recommendation [linked]"? (suitably worded to match the terms chosen)?

nigelmegitt:
> perhaps a Note "Amended Recommendations should not be confused with the process of Amending a Recommendation [linked]"?

That's okay, but if we know there's a likelihood of confusion it would be better still to avoid the use of similar terms and remove the source of confusion at source, and then not need the note.

dwsinger:
Perhaps we could safely rename "Amended Recommendation" to "Team-amended Recommendation" safely enough?

PR #451, opened 2020-09-11
Amend

frivoal:
Proposal to apply the changes suggested in #449


frivoal:
Resolution in: https://www.w3.org/2020/09/16-w3process-minutes.html#x093

PR #485, opened 2020-12-25
possible missing link in 'Relation of Process Document to Patent Policy' section

himorin:
only closing link exists.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -111,7 +111,7 @@ Relation of Process Document to Patent Policy</h2>
 	W3C Members' attention is called to the fact
 	that provisions of the Process Document are binding on Members
 	per the <a href="https://www.w3.org/Consortium/Agreement/Member-Agreement">Membership Agreement</a> [[MEMBER-AGREEMENT]].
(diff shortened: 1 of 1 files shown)


dwsinger:
Hi.

thanks, but if you look at the published document I think you'll find that there is a hyperlink there; it's made by Bikeshed processing the [[! ]] into a bibliography entry and hyperlink.

himorin:
In one line before (from this line), there is both a link and bikeshed [[! ]] to a bibliography entry for Membership Agreement. And I thought isolated closing `</a>` were inserted to put similar markup for patent policy.

Which is better: to keep, or to remove `</a>`?

dwsinger:
weird.  I'll let the editor sort me out

frivoal:
I suspect the link on Membership Agreement is an accidental leftover from before the document was edited in Bikeshed. We should probably remove it, as well as the stray closing `</a>` tag after "The W3C Patent Policy".
That said, having a direct link as well isn't wrong either, so we could go either way. If anyone has a strong opinion, please state it, otherwise, I'll figure out what my favorite way to clean this up is and do it in a little while.

PR #488, opened 2021-01-07
Note track

### LINKED ISSUE #342: Decoupling Notes and REC track
This issue is not about Notes when they are used as a retirement path for abandoned REC Track documents. I'm talking about Notes that are meant from the start to be Notes.

Notes can be updated in place, so they don't necessarily need to go through a WD phase first. But the process does allow that. However, the first such Working Draft would be a FPWD. This has patent implications, which are uncessary for a document not meant to try and become a REC, but still technically binding.

We end up in the strange situation where we have to ask lawyers to review a draft to check if they want to exclude any patent when it becomes a REC, while the status section says that it won't become a REC.

We should remove that work mode (and update the Process accordingly), possibly by simply publishing preliminary versions of Notes as Notes (rather than as WDs), or by creating a special maturity that signals the lack of maturity ("Preliminary Note"? "Draft Note?").

frivoal:
Separates the Note Track from the REC Track. The REC track is under the patent policy, while the Note Track isn't. The fact that they shared some stages has been messy. There were previously two connections: FPWD (and other WD) could be used as a first stage for Notes, and Notes were a way to discontinue REC track documents. This pull request separates them by introducing "Draft Note" to replace working drafts of notes and "Discontinued Draft Recommendation" to replace notes of abandoned REC-track work.

If accepted, this would close #342

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/488.html" title="Last updated on Feb 24, 2021, 4:32 PM UTC (3105678)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/488/7bdea3e...frivoal:3105678.html" title="Last updated on Feb 24, 2021, 4:32 PM UTC (3105678)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2819,6 +2819,13 @@ Maturity Levels on the Recommendation Track</h4>
 					See also “W3C Royalty-Free (RF) Licensing Requirements”
 					in the W3C Patent Policy [[!PATENT-POLICY]].
 			</dl>
+		<dt>
+			<dfn id="discontinuedREC" export >Discontinued Draft</dfn>
+		<dd>
+			A [=technical report=] representing the state of a Recommendation-track document
+			at the point at which work on it was discontinued.
+			See [[#abandon-draft]].
+
 	</dl>
 
 	Only sufficiently technically mature work should be advanced.
@@ -3144,6 +3151,7 @@ Revising a Working Draft</h4>
 		<li>Revised <a href="#revising-wd">Working Draft</a>
 		<li><a href="#transition-cr">Candidate Recommendation</a>
 		<li><a href="#Note">Working Group Note</a>
+		<li><a href="#abandon-draft">Discontinued Draft</a></li>
 	</ul>
 
 <h4 id="transition-cr" oldids="candidate-rec, last-call">
@@ -3196,7 +3204,7 @@ Transitioning to Candidate Recommendation</h4>
 		<li>A revised <a href="#publishing-crrs">Candidate Recommendation Snapshot</a></li>
 		<li>A revised <a href="#publishing-crud">Candidate Recommendation Draft</a></li>
 		<li><a href="#transition-pr">Proposed Recommendation</a></li>
-		<li><a href="#Note">Working Group Note</a></li>
+		<li><a href="#abandon-draft">Discontinued Draft</a></li>
 	</ul>
 
 	[=Advisory Committee=] representatives <em class="rfc2119">may</em> initiate an
@@ -3302,7 +3310,7 @@ Publishing a [=Candidate Recommendation Draft=]</h5>
 		<li>A revised <a href="#publishing-crud">Candidate Recommendation Draft</a></li>
 		<li><a href="#transition-pr">Proposed Recommendation</a>,
 		if there are no [=substantive change=] other than dropping [=at risk=] features</li>
-		<li><a href="#Note">Working Group Note</a></li>
+		<li><a href="#abandon-draft">Discontinued Draft</a></li>
 	</ul>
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3627,19 +3637,25 @@ Abandoning an Unfinished Technical Report</h5>
...
+	an explanation of why it was discontinued.
 
-	Such a document should be documented as discontinued in its status section,
-	with a <a href="#abandon-draft">link to this section of this document</a>,
-	and accompanied by an explanation of why it was discontinued.
+	A [=Working Group=] <em class="rfc2119">may</em> resume work
+	on a [=technical report=]
+	within the scope of its charter
+	at any time,
+	at the maturity level the specification had before publication as a [=Discontinued Draft Recommendation=].
```
nigelmegitt, line 3658: This text reproduces what was present before, so this comment shouldn't block merging, but possibly generate a new issue if others agree.

Consider:

WD -> CR -> DDR -> [years pass] -> CR 

It seems potentially risky to permit this without some need to demonstrate wide review of the document again before re-entering CR. For example HR requirements may have changed.

index.bs
```diff
@@ -3815,55 +3830,36 @@ Working Group and Interest Group Notes</h3>
...
 
-	In order to publish a [=Note=],
+	In order to publish a [=Note=] or [=Draft Note=],
 	a [=Working Group=] or [=Interest Group=]:
 
 	<ul>
-		<li>
-			<em class="rfc2119">may</em> [=publish=] a [=Note=]
-			with or without its prior publication as a [=Working Draft=].
-
```
nigelmegitt, line 3838: It's not clear why this has been removed?

nigelmegitt (commented):
I like this, generally. Two queries, one about a possible new issue regarding re-starting work on a previously abandoned Rec, the other about a removed item.

index.bs
frivoal: two thoughts:
* to resume at a previous maturity higher than WD, we probably should require no textual change (or at least no substantive change)
* Maybe we shouldn't allow resuming as anything other than WD, though we need to be careful about the patent policy implications

index.bs
frivoal: This is meant to be an editorial rather than substantive change: the old text said how to publish a note, and noted that you could do so with or without prior publication as draft. The new text says how you can publish Notes or Draft Notes (same conditions), and that you can go from either to either. Having written that, it seemed that the bit of text you're inquiring about no longer adds anything. But there's nothing wrong with it either (other than s/Working Draft/Draft Note), so we could reinstate it if its absence makes things too unclear.

nigelmegitt: Thanks, I wondered if that was the rationale @frivoal . I think the text does help to clarify that it is normatively permitted to go from WD -> Note or just publish a Note directly so I would not remove it as part of this change. But I don't think this is a major blocking concern.

frivoal: I think that when you're reading the diff, the removal of that text might be surprising, but if you rather look at the resulting text, it's already plenty clear that it is both allowed to go from Draft to Note or straight to Note, as we already say:

> 	Some W3C Notes are developed through successive Draft Notes
	before publication as a full Notes,
	while others are published directly as a Note.

So if you don't mind, I'd rather not reinstate that text.

index.bs
```diff
@@ -2819,6 +2819,13 @@ Maturity Levels on the Recommendation Track</h4>
 					See also “W3C Royalty-Free (RF) Licensing Requirements”
 					in the W3C Patent Policy [[!PATENT-POLICY]].
 			</dl>
+		<dt>
+			<dfn id="discontinuedREC" export >Discontinued Draft Recommendation</dfn> (<abbr>REC</abbr>)
```
dontcallmedom, line 2823: I'm very supportive of the direction of change; the name "Discontinued Draft Recommendation" feels suboptimal: too easy to confuse with Obsolete or Rescinded Recommendation when they are in practice very different. “Discontinued Draft" is both shorter and less confusion prone. (I'm also guessing REC would not be a good abbreviation for that status no matter what >:)

dwsinger:
There was some suggestion on the process call that WGs might avoid a status that had any hints of negativity in the name. Would "Inactive" be better than "Discontinued"?

PR #490, opened 2021-01-07
Create .pr-preview.json

### LINKED ISSUE #417: Enable pr-preview
Enabling pr-preview would allow PRs to be previewed in finished and diff form more easily.

nigelmegitt:
Closes #417.


frivoal:
This PR wouldn't work as-is, as the Process's bikeshed needs a bunch of boilerplate files to be taken into account as well (see all the `*.include` files in https://github.com/w3c/w3process). I don't believe PR-preview supports that.
I suppose we could still solve this by up streaming all these boilerplate files to bikeshed itself, rather than having them live in-repo. They're pretty stable and haven't changed since February 2019, so that's probably doable.

This will be a bit of an annoyance for the editor (me), as I find it useful to fiddling with them (without committing) when I generate the drafts for AC Review or the official process. But I suppose that the benefits of enabling PR-review trump those of making the editor's life a bit easier a couple of times a year.

I'll try and upstream those, and then we can merge this.

nigelmegitt:
Thanks @frivoal much appreciated.

frivoal:
I've added a commit to this branch that will make it possible to build without the local boilerplate files once https://github.com/tabatkins/bikeshed/pull/1892 has been merged.

PR #506, opened 2021-03-12
Shorten registry intro and improve statements about Patent Policy

frivoal:
***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/506.html" title="Last updated on Mar 15, 2021, 11:14 AM UTC (2dcd764)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/506/a32b070...frivoal:2dcd764.html" title="Last updated on Mar 15, 2021, 11:14 AM UTC (2dcd764)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2292,7 +2292,7 @@ W3C Technical Reports</h3>
 			These are typically published in a separate [=registry report=],
 			although they can also be directly embedded in [=Recommendation Track=] documents
 			as a [=registry section=].
-			Establishing a [=registry=] requires [=wide review=] and [=consensus=],
+			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
 			and can even be done without a [=Working Group=].
 			See [[#registries]] for details.
--- a/index.bs
+++ b/index.bs
@@ -2287,16 +2287,13 @@ W3C Technical Reports</h3>
 
 		<dt>Registries
 		<dd>
-			[=Working Groups=] can also publish registries
-			in order to document collections of values or other data
-			that have no normative implementation requirements.
-			Registries are generally companion to [=Recommendation Track=] documents
-			which contain the related normative requirements,
-			and are typically published in a separate [=registry report=],
-			although they can also be directly embedded in [=Recommendation Track=] documents.
-			The registry track requires [=wide review=] and [=consensus=]
-			on what the registry will contain and how it will be managed.
-			Once set up, changes to registry entries are lightweight
+			[=Working Groups=] can also publish [=registries=]
+			in order to document collections of values or other data.
+			These are typically published in a separate [=registry report=],
+			although they can also be directly embedded in [=Recommendation Track=] documents
+			as a [=registry section=].
+			Establishing a [=registry=] requires [=wide review=] and [=consensus=],
+			but once set up, changes to registry entries are lightweight
 			and can even be done without a [=Working Group=].
 			See [[#registries]] for details.
 	</dl>
@@ -4064,6 +4061,9 @@ Publishing Registries</h4>
 	is purely documentational,
 	is not subject to the W3C Patent Policy,
 	and <em class=rfc2119>must not</em> contain any requirements on implementations.
+	For the purposes of the Patent Policy [[PATENT-POLICY]],
+	any [=registry section=] in a [=Recommendation track=] document
+	is not a normative portion of that specification.
 
 	The [=registry report=] or [=registry section=] <em class=rfc2119>must</em>:
 	<ul>
@@ -4116,8 +4116,8 @@ Publishing Registries</h4>
 
 	<ul>
 		<li>
-			Since [=Registry Reports=] are not subject to the [[PATENT-POLICY]],
-			none of their publications correspond,
+			[=Registry reports=] are not subject to the [[PATENT-POLICY]],
+			and therefore none of their publications correspond,
 			to [=First Public Working Draft=],
 			[=Working Draft=],
 			or [=Patent Review Draft=]


index.bs
```diff
@@ -2287,16 +2287,13 @@ W3C Technical Reports</h3>
...
-			The registry track requires [=wide review=] and [=consensus=]
-			on what the registry will contain and how it will be managed.
-			Once set up, changes to registry entries are lightweight
+			[=Working Groups=] can also publish [=registries=]
+			in order to document collections of values or other data.
+			These are typically published in a separate [=registry report=],
+			although they can also be directly embedded in [=Recommendation Track=] documents
+			as a [=registry section=].
+			Establishing a [=registry=] requires [=wide review=] and [=consensus=],
+			but once set up, changes to registry entries are lightweight
```
nigelmegitt, line 2296: The phrase "Establishing a registry" is doing a lot of work here: it would be helpful to clarify that establishing a Registry means creating a registry definition, and that the registry entries are conceptually distinct.

index.bs
```diff
@@ -4064,6 +4061,9 @@ Publishing Registries</h4>
 	is purely documentational,
 	is not subject to the W3C Patent Policy,
 	and <em class=rfc2119>must not</em> contain any requirements on implementations.
+	For the purposes of the Patent Policy [[PATENT-POLICY]],
+	any [=registry section=] in a [=Recommendation track=] document
+	is not a normative portion of that specification.
```
nigelmegitt, line 4066: Would be good to clarify if this applies to the registry definition or only the registry entries. Seems to me that the registry definition is essentially a data model + a process, and potentially that _could_ somehow be subject to the patent policy, at least theoretically?

index.bs
fantasai: Hmmm, maybe we should use "defining" instead of "establishing" and then link it to the registry definition term?

index.bs
```diff
@@ -2287,16 +2287,13 @@ W3C Technical Reports</h3>
...
-			although they can also be directly embedded in [=Recommendation Track=] documents.
-			The registry track requires [=wide review=] and [=consensus=]
-			on what the registry will contain and how it will be managed.
-			Once set up, changes to registry entries are lightweight
+			[=Working Groups=] can also publish [=registries=]
+			in order to document collections of values or other data.
+			These are typically published in a separate [=registry report=],
+			although they can also be directly embedded in [=Recommendation Track=] documents
+			as a [=registry section=].
+			Establishing a [=registry=] requires [=wide review=] and [=consensus=],
```
fantasai, line 2295:
```suggestion
			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
```

index.bs
fantasai: If that's possible, then we'll need to do it on the Recommendation Track. The Patent Policy only applies to the Rec Track...

index.bs
nigelmegitt: That would be a good improvement.

frivoal: Done in 2dcd7648be90254db1895e0a19a561f5c89bee3d

index.bs
dwsinger: This is where life gets complicated. The registry definition part of the registry section is (though I doubt it matters); the registry tables are not. (But that should be clear from the fact that the section contains no normative requirements and there are no expressed normative requirements about its content).

index.bs
```diff
@@ -4116,8 +4116,8 @@ Publishing Registries</h4>
 
 	<ul>
 		<li>
-			Since [=Registry Reports=] are not subject to the [[PATENT-POLICY]],
-			none of their publications correspond,
+			[=Registry reports=] are not subject to the [[PATENT-POLICY]],
+			and therefore none of their publications correspond,
```
dwsinger, line 4120: A Registry definition is a normal subject-to-AC-and-PP recommendation, no?

fantasai: No, only documents on the Recommendation Track are subject to the Patent Policy (per the Patent Policy)... if we want to have the registry definition subject to the Patent Policy, it's going to have to be on the REC track. (We can make special rules about skipping CR, but it can't be on a separate track.)

fantasai: (We probably need to get PSIG review on whether registry definitions should or should not be subject to the Patent Policy, but in any case, if they need to be then they need to be on the REC track, and if they do not need to be, then it's better for them to not be on the REC track.)

index.bs
fantasai: We probably need to get PSIG review on whether registry definitions should or should not be subject to the Patent Policy, but in any case, if they need to be then they need to be on the REC track, and if they do not need to be, then it's better for them to not be on the REC track. The Patent Policy only applies to documents on the REC track.

fantasai: https://www.w3.org/Consortium/Patent-Policy-20200915/#sec-Licensing

frivoal: That said, I think we should be clear about what that would do: if the definitions of the registry get to be on the REC track and to be covered by the Patent Policy, what would be protected is implementing the definitions: in other words, *creating and operating the registry tables*. This would do nothing for software implementations of whatever is described in the registry tables. Are we actually concerned that we'd design a registry so that editing it and publishing it could be a violation of someone's patents, and that someone could attempt to sue the W3C for our editorial practices for that registry? That seems far-fetched to me.

nigelmegitt: > Are we actually concerned that we'd design a registry so that editing it and publishing it could be a violation of someone's patents, and that someone could attempt to sue the W3C for our editorial practices for that registry? That seems far-fetched to me.

Me too, and yet, I'm willing to bet that someone somewhere has tried such a trick, or may do in the future (call me a cynic!). Maybe we could sidestep this and avoid potential unnecessary complexity if someone is able to document publicly (maybe in the guide, or some other technical report) typical and useful existing registry patterns that are known to be patent-free.

frivoal: Maybe I'm just underinformed, but I am not aware of any standards organization ever being sued because the way they develop documents infringes on someone's patent, nor of any standards organization having some form of patent protection against that. Unlike patent protection for implementations of specifications, which is very valuable, this seems to me to be a theoretical exercise of very limited interest.

index.bs
dwsinger: When I suggested this division (referencing spec., definition, tables) I had the first two on the Rec. Track. The argument that definitions don't need a CR phase for implementation caused the design of the separate track. I don't think we discussed whether that should mean taking it off the Patent Policy as well. I need to think, and the question needs visibility.

frivoal:
Merging, with the chair's approval. The discussion in the comment threads starting at https://github.com/w3c/w3process/pull/506#pullrequestreview-612744915 and https://github.com/w3c/w3process/pull/506#pullrequestreview-610643600 remain valid, but what they're discussing exists prior to this editorial pull request, which merely made the question more obvious.

Issue #511, opened 2021-03-15
Define minutes


fantasai:
We have some requirements applying to the minutes of meetings, e.g. in
  https://www.w3.org/2020/Process-20200915/#GeneralMeetings
and in the new tooling requirements
  https://www.w3.org/wiki/W3C_Tooling_Policy

But we don't actually define what the minutes are or that they need to exist. Seems like such a definition would be helpful both for these sections to be able to reference, but also to help distinguish them from the transcripts being addressed by https://github.com/w3c/w3process/issues/334

PR #512, opened 2021-03-15
Define minutes

frivoal:
See #511

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/512.html" title="Last updated on Mar 24, 2021, 3:20 PM UTC (4d1475c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/512/e37ef1a...frivoal:4d1475c.html" title="Last updated on Mar 24, 2021, 3:20 PM UTC (4d1475c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1152,7 +1152,7 @@ Meeting Scheduling and Announcements</h4>
 			<td><span class="time-interval">three days</span>
 			<td><span class="time-interval">24 hours</span>
 		<tr>
-			<th scope=row>Minutes available (after)
+			<th scope=row>[=Minutes=] available (after)
 			<td><span class="time-interval">two weeks</span>
 			<td><span class="time-interval">48 hours</span>
 	</table>
@@ -1162,6 +1162,16 @@ Meeting Scheduling and Announcements</h4>
 	about the date and location of a meeting.
 	Shorter notice for a meeting is allowed
 	provided that there are no objections from group participants.
+
+<h4 id="meeting-minutes">
+Meeting Minutes</h4>
+
+	Groups are expected to take and retain <dfn>minutes</dfn> of their meetings,
+	and <em class="rfc2119">must</em> record in particular
+	any official [=group decisions=] made during the meeting discussions.
+	Precise details of the discussion leading to such decisions
+	is recommended, but not required
+	provided that the rationale for the [=group decision=] is nonetheless clear.
 
 <h4 id="meeting-recording">
 Meeting Recordings and Transcripts</h4>
@@ -1368,7 +1378,7 @@ Votes</h3>
 	through technical discussion and compromise have failed,
 	and that a vote is necessary to break a deadlock.
 	In this case the [=Chair=] <em class="rfc2119">must</em> record
-	(e.g., in the minutes of the meeting or in an archived email message):
+	(e.g., in the [=minutes=] of the meeting or in an archived email message):
 
 	<ul>
 		<li>
@@ -1686,7 +1696,7 @@ Requirements for All Working and Interest Groups</h3>
 
 	Each group <em class="rfc2119">must</em> have an archived mailing list
 	for formal group communication
-	(e.g., for meeting announcements and minutes,
+	(e.g., for meeting announcements and [=minutes=],
 	documentation of decisions,
 	and [=Formal Objections=] to decisions).
 	It is the responsibility of the [=Chair=] and [=Team Contact=]
@@ -4465,7 +4475,7 @@ Workshops and Symposia</h2>
 	The Call for Participation in a Workshop or Symposium
 	<em class="rfc2119">may</em> indicate participation requirements or limits,
 	and expected deliverables
-	(e.g., reports and minutes).
+	(e.g., reports and [=minutes=]).
 	Organization of an event does not guarantee
 	further investment by W3C in a particular topic,
 	but <em class="rfc2119">may</em> lead to proposals for new activities or groups.
--- a/index.bs
+++ b/index.bs
@@ -1169,8 +1169,7 @@ Meeting Minutes</h4>
 	Groups <em class=rfc2119>should</em> take and retain <dfn>minutes</dfn> of their meetings,
 	and <em class="rfc2119">must</em> record
 	any official [=group decisions=] made during the meeting discussions.
-	Precise details of the discussion leading to such decisions
-	is recommended, but not required
+	Details of the discussion leading to such decisions are not required,
 	provided that the rationale for the [=group decision=] is nonetheless clear.
 
 <h4 id="meeting-recording">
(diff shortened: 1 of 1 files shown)


jeffjaffe:
By making it a MUST to record decisions, does it imply that any decisions that are not recorded - in fact are not **decisions** (and hence may not be implemented, may not be objected to, etc.)?

nigelmegitt:
Another possible gap: is it required that when decisions are recorded, they are explicitly identified in the minutes as decisions? For example:

1. Informal style:

```
group: [some discussion of a potential change]
[chair]: Ok let's go ahead and make that change.
```

2. Formal style:

```
group: [discussion of a change]
PROPOSAL: Change X to Y
[chair]: Any objections to the proposal?
group: No objections
[chair]: We have consensus: this is a group decision.
RESOLUTION: Change X to Y
```

My reading of the current proposed wording is that either would be acceptable. I'm not sure if that's reasonable though. What's the worst that could happen with the informal style? Someone casually reviewing the minutes misses the discussion, or misunderstands what the decision actually was, and then later there is pain, distress and expense when an objection to the decision is raised some time after the group's decision review period has ended, and maybe only on transition to a higher maturity level of a technical report. Not sure if this is bad enough to require any degree of formality though.

index.bs
```diff
@@ -1163,6 +1163,16 @@ Meeting Scheduling and Announcements</h4>
...
 
+<h4 id="meeting-minutes">
+Meeting Minutes</h4>
+
+	Groups are expected to take and retain <dfn>minutes</dfn> of their meetings,
+	and <em class="rfc2119">must</em> record in particular
+	any official [=group decisions=] made during the meeting discussions.
+	Precise details of the discussion leading to such decisions
+	is recommended, but not required
+	provided that the rationale for the [=group decision=] is nonetheless clear.
```
dwsinger, line 1174: provided that the rationale, and consensus of support, for the …

index.bs
```diff
@@ -1163,6 +1163,16 @@ Meeting Scheduling and Announcements</h4>
 	Shorter notice for a meeting is allowed
 	provided that there are no objections from group participants.
 
+<h4 id="meeting-minutes">
+Meeting Minutes</h4>
+
+	Groups are expected to take and retain <dfn>minutes</dfn> of their meetings,
```
chaals, line 1169: s|are expected to|<em class="rfc2119">should</em>|

index.bs
```diff
@@ -1163,6 +1163,16 @@ Meeting Scheduling and Announcements</h4>
...
+	and <em class="rfc2119">must</em> record in particular
```
chaals, line 1170: s/in particular//

chaals:
@jeffjaffe 

> By making it a MUST to record decisions, does it imply that any decisions that are not recorded - in fact are not decisions

That seems to be the logical conclusion, and I think that is a positive thing. It should be noted in the chapter on decisions, as well.

chaals:
@nigelmegitt 

>  is it required that when decisions are recorded, they are explicitly identified in the minutes

I agree that we can leave this unspecified, but that if we do the outcome will indeed be that groups where there is disagreement on issues eventually learn that it is helpful to have explicit documentation.

frivoal:
@jeffjaffe If a consensus falls in a forest and no one is around to record it, does it make a decision?

index.bs
fantasai: It's not a [=group decision=] if there wasn't consensus backing as determined by the chair (or a vote, or ...; there's an entire section on [decisions when there's no consensus](https://www.w3.org/Consortium/Process/Drafts/#managing-dissent)), so I would rather not include anything about consensus here. If a group just wants to take minutes as a list of 

RESOLVED: thing decided
RATIONALE: why we did it
RESOLVED: other thing decided
RATIONALE: why we did it

That should be acceptable. (Might even be more understandable than the transcription style for groups who tend to have very sparse transcribing.)

fantasai:
A decision is defined to exist when the chair as assessed the consensus of the group (or the results of a vote), and its definition doesn't hinge on the existence of minutes: https://w3c.github.io/w3process/#decisions

That said, it's pretty impractical to use unrecorded decisions as the basis for any formal steps in the Process, as we generally ask for a link to the record of the decision...

nigelmegitt:
> That said, it's pretty impractical to use unrecorded decisions as the basis for any formal steps in the Process, as we generally ask for a link to the record of the decision...

Worth comparing with §3.3.4, which says, in the context of reopening decisions:
> The Chair should record that a decision has been reopened, and must do so upon request from a group participant.

So it would make sense to have symmetry for when decisions are made in the first place, e.g.:
> The Chair should record a decision, and must do so upon request from a group participant.


it also sounds like a Note would be useful along the lines of:
> Note: Some steps defined in the Process require that Decisions are recorded, such as advancement of Recommendation track specifications.

index.bs
dwsinger: OK. I suppose that we can leave to the group/chair the evidence that consensus was measured somehow.

index.bs
```diff
@@ -1163,6 +1163,16 @@ Meeting Scheduling and Announcements</h4>
...
+	any official [=group decisions=] made during the meeting discussions.
+	Precise details of the discussion leading to such decisions
```
wseltzer, line 1172: Remove "Precised details".

frivoal: Do you mean to suggest removing the word "precise" only? Removing both words would make the sentence ungrammatical.

dwsinger: 'records of the discussion…'

index.bs
```diff
@@ -1163,6 +1163,16 @@ Meeting Scheduling and Announcements</h4>
...
 	provided that there are no objections from group participants.
 
+<h4 id="meeting-minutes">
+Meeting Minutes</h4>
+
+	Groups <em class=rfc2119>should</em> take and retain <dfn>minutes</dfn> of their meetings,
+	and <em class="rfc2119">must</em> record
+	any official [=group decisions=] made during the meeting discussions.
+	Precise details of the discussion leading to such decisions
+	is recommended, but not required
```
TallTed, line 1173:
```suggestion
	Relevant details of the discussion leading to such decisions
	(a/k/a resolutions) is recommended, but not required.
```

frivoal:
@wseltzer We've reworded the sentence you didn't like. Is now reads:

> Details of the discussion leading to such decisions are not required, provided that the rationale for the is nonetheless clear.

Is this acceptable?

chaals:
@frivoal quoted

> provided that the rationale for the is nonetheless clear.

s/for the// otherwise the sentence is grammatically broken.

frivoal:
> s/for the// otherwise the sentence is grammatically broken.

Sorry, that was a failure to escape a bit of markup.

It's supposed to read:
> …provided that the rationale for the group decision is nonetheless clear…

frivoal:
Given that we had conceptual agreement on this during the call, and the blocker was Wendy's concern with wording, which has now been resolved, I'm taking the liberty to merge this, in the interest of no overloading the next call with minor things.

Notice to the chair: @dwsinger, if you think this was overstepping as an editor, feel free to schedule it for discussion.

Issue #505, opened 2021-03-11
Define a role of "Alternate AC Rep" & enable them to perform AC duties when necessary


hober:
Each Member organization has one AC Rep. That rep is the only person who can do a number of things, including voting on AC ballots & joining / leaving groups. But AC Reps, like any other W3C participants, go on vacation, take sick days, and are otherwise periodically unavailable for work-related tasks.

It would be fairer to AC Reps if they could each have a designated alternate or substitute for such times, so that they don't have to file ballots etc. while not at work.

(A sufficiently motivated Member organization could simulate this currently by temporarily changing AC Reps to the alternate every time the normal AC Rep is out, but that requires a level of coordination amongst the two of them and the W3C Team that seems far too heavyweight and cumbersome.)

hober:
Filed this after a conversation with @tantek and @dwsinger on this topic.

TzviyaSiegman:
This is a great idea

dwsinger:
I have such residual guilt feelings from when I took vacation and didn't alert my colleagues to ballots in a timely fashion, and wasn't able/willing to respond to them from the beach.

cwilso:
Strong, STRONG +1 on this idea.  I was just discussing the same thing with someone internally this week.

cwilso:
It seems, though, like this should be able to be handled at the sysops level - swapping out or enabling a backup responder - rather than having to detail it in the Process?

dwsinger:
I think that the ideal would be that an alternate is permanent, and that either the AC Rep or alternate can act on any ballot. Whether the balloting system can handle that I do not know. I wondered if it was a process change, or just practice, but the process doesn't mention any idea of alternative AC Voters.

jeffjaffe:
I support having this capability.  Anything that helps and simplifies engagement is good, and this is a terrific example.

We need to work out the details of the empowerment and then, as David says, the complexity of implementation.  Here are some example details.

- In terms of designating and empowering the backup AC rep, I assume that the primary AC rep has the authority to empower when they go on leave, and end the empowerment when they come back.
- Do both the primary and backup have posting privileges in ac-forum  (I assume yes, Section 2.1.3.1 already empowers an AC rep to request an additional person to be on ac-forum)?
- Do both primary and backup have permission to attend the semi-annual AC meeting?
[...]

PR #524, opened 2021-04-30
Alternate AC Rep

wseltzer:
Address issue #505 Based on [text proposal](https://github.com/w3c/w3process/issues/505#issuecomment-797266570) from @fantasai

(Additional parenthetical mine; may be unecessary detail.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/pull/524.html" title="Last updated on Apr 30, 2021, 1:53 PM UTC (a49f6df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/524/ea8220a...a49f6df.html" title="Last updated on Apr 30, 2021, 1:53 PM UTC (a49f6df)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -418,6 +418,8 @@ Advisory Committee (AC)</h4>
 	by disclosing information according to the mechanisms described in the New Member Orientation.
 	See also the additional roles of [=Advisory Committee representatives=]
 	described in the W3C Patent Policy [[PATENT-POLICY]].
+	
+	The AC representative may delegate any of their rights and responsibilities to an alternate (except the ability to designate an alternate).
 
 	Additional information for Members is available at the <a href="https://www.w3.org/Member/">Member Web site</a> [[MEMBER-HP]].
 


tantek (changes requested):
This is a good start (including the parenthetical), however IMO this is insufficient to capture the needs of this issue. The "may delegate" implies an ephemeral act, and the "any of their rights" etc. implies the need for specific enumeration. Neither of these is desirable.

The need here is to designate an explicitly ongoing alternate, that persists in parallel with the "primary" AC rep, as in all the same abilities to be on the same lists, receive all the same notifications, answer all the same forms etc. with the presumption that any "conflict" between primary & alternate would be up to the primary to resolve (not W3C or any systems thereof) if necessary by un-designating the alternate. The key here is for the alternate to be fully empowered to receive all comms and act as the AC rep, so that if the primary is incapacitated or otherwise unavailable for any reason (unexpected or planned like a vacation), the member's AC participation is expected to continue without interruption.

dwsinger:
I think this is probably ok as the formal process text. However, I am mighty curious to know the details, and that might inform us whether any other process text is needed.

wseltzer:
@tantek, that's the idea for implementation -- the designation sticks until changed, and includes all rights without enumeration.  What do you think needs to be added or changed in Process text?

dwsinger:
Agenda+ to decide if it lands in P2021

wseltzer:
Proposed implementation: 
An organization's AC Rep may designate an Alternate AC Rep, which designation persists until the AC Rep changes or revokes it. Any other action the AC Rep is entitled to take (e.g. responding to AC reviews, voting in elections, joining a group, approving participation requests) may be completed by the AC Rep or Alternate, and binds the organization to the same extent as if the AC Rep had acted. In the event of multiple responses to a request for organizational input, the last response before deadline controls. (Ideally, a shared dashboard helps them coordinate.)

Alternates receive the same email notifications and list subscriptions as AC Reps, and may attend meetings and calls. All participants are asked to ensure that people know "with which hat" they are speaking when commenting, and to avoid overwhelming message frequency.

Questions on the implementation: 
* Should the number of alternates be a maximum of 1? 
* Should alternates and AC Reps both be able to speak on ac-forum, or only one in a given timeframe?
* Should team communicate by default with both AC Rep and Alternate when reaching out to a member?

dwsinger:
General discussion is in #505 thanks

tagawa:
Just to clarify, would designating an alternate AC rep mean both people could perform AC duties simulatenously, or would the original AC rep lose that ability until the alternate designation is revoked? In other words, would they work as a team or pass the baton?

wseltzer:
@tagawa see the proposed implementation in the issue discussion  #505

PR #538, opened 2021-06-09
Explicitly require tools be accessible to people with disability 

### LINKED ISSUE #537: Clarify that tooling must be accessible for people with disabilities
Currently in the tooling section there is a bullet stating: 

> "Any tooling used by the group for producing its documentation and deliverables or for official group discussions should be usable without additional cost by all who wish to participate, to allow their effective participation."

It is not clear from this that tooling must be accessible for people with disabilities who contribute to the W3C community. The term "usable... by all" doesn't have the clarity or get the outcome needed, and W3C Team are having to scramble more and more often to deal with barriers from tools, or arrange assessments, etc. 

Here is one possibility: 

""Any tooling used by the group for producing its documentation and deliverables or for official group discussions should be accessible for people with disabilities according to relevant standards, and usable without additional cost by all who wish to participate, to allow their effective participation."

Thoughts welcome.

chaals:
fixes #537 

Adds the wording proposed, and makes cost a separate requirement (so it can be discussed separately).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/pull/538.html" title="Last updated on Jun 9, 2021, 3:01 PM UTC (fe5308e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/538/231cf4e...fe5308e.html" title="Last updated on Jun 9, 2021, 3:01 PM UTC (fe5308e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -674,9 +674,12 @@ Tooling for Discussions and Publications</h4>
 			Any tooling used by the group
 			for producing its documentation and deliverables
 			or for official group discussions
-			<em class=rfc2119>should</em> be usable
-			without additional cost
-			by all who wish to participate,
+			<em class=rfc2119>should</em> be accessible for people with disabilities
+			
+		<li id=tool-nocost>	
+			Any tooling used by the group additional cost
+			<em class=rfc2119>should</em> be available at no additional cost
+			to all who wish to participate,
 			to allow their effective participation.
 
 			Note: If a new participant joins who cannot use the tool,
--- a/index.bs
+++ b/index.bs
@@ -674,19 +674,14 @@ Tooling for Discussions and Publications</h4>
 			Any tooling used by the group
 			for producing its documentation and deliverables
 			or for official group discussions
-			<em class=rfc2119>should</em> be accessible for people with disabilities
+			<em class=rfc2119>should</em> be accessible to people with disabilities.
 			
 		<li id=tool-nocost>	
 			Any tooling used by the group 
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>should</em> be available at no additional cost
-			to all who wish to participate,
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):

dwsinger:
I think it might be better as a single requirement so that the following Note applies to both?

shawna-slh:
Thanks for the review ping.

1. Easy things:
   * [ ] I strongly prefer "accessible to people with disabilities" (not ....._for_ people...)
   * [ ] new sentence missing terminal punctuation


2. I appreciate the idea of splitting up accessibility and cost in different bullets; however, it doesn't work well as drafted, because:
   * The Note (“Note: If a new participant joins who cannot use the tool…”) applies to both bullets
   *  “, to allow their effective participation.” applies to both bullets

Some options:
* Could put as single bullet.
* Could make sub-bullets.
* Could delete “, to allow their effective participation.” if feel that point isn’t needed here given it is in the paragraph before the list (“allow all would-be participants to effectively participate”).

shawna-slh (changes requested):
https://github.com/w3c/w3process/pull/538#issuecomment-857757864

fantasai:
This change regresses the case of other reasons a participant might not be able to use the tool, so I can't agree with this change.

(I also think that it should stay as one rule, because the rules are split largely on their target, and both of these have the same target.)

frivoal:
Filed a separate pull request (#539) to propose a less invasive edit, to solve the issue without regressing the [cases fantasai mentions](https://github.com/w3c/w3process/issues/537#issuecomment-857879755).

Alvinchongry (approved):

Alvinchongry (approved):

frivoal:
Rejected in favor of #539

PR #539, opened 2021-06-09
Minimal edit to put emphasis on people with disabilities rather than cost for tooling accessibility

frivoal:
See #537

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/539.html" title="Last updated on Jun 9, 2021, 5:13 PM UTC (9c7a727)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/539/231cf4e...frivoal:9c7a727.html" title="Last updated on Jun 9, 2021, 5:13 PM UTC (9c7a727)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -675,8 +675,9 @@ Tooling for Discussions and Publications</h4>
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>should</em> be usable
-			without additional cost
+			(without additional cost)
 			by all who wish to participate,
+			including people with disabilities,
 			to allow their effective participation.
 
 			Note: If a new participant joins who cannot use the tool,


index.bs
```diff
@@ -675,8 +675,9 @@ Tooling for Discussions and Publications</h4>
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>should</em> be usable
-			without additional cost
+			(without additional cost)
 			by all who wish to participate,
```
chaals, line 679: I don't think the "by all...participants" is necessary...

index.bs
```diff
@@ -675,8 +675,9 @@ Tooling for Discussions and Publications</h4>
 			for producing its documentation and deliverables
 			or for official group discussions
 			<em class=rfc2119>should</em> be usable
```
chaals, line 677: I agree with the original comment in #537 that if you don't call this out more explicitly, there will be problems. My experience is that while in many cases staff are alert to the issue, they are easily put under pressure by Working Groups who are keener to get their work finished than to be inclusive, and it is important to draw fairly bright red lines. In part this is because even with goodwill and effort, it is often dificult to find accessible tooling - many of the cpmmon tools people know from their work are in fact seriously problematic :(

index.bs
fantasai: No, it is very intentional. We are not scoping this rule to the entire population of the world (as we are with the deliverables) but rather to the people who are actually intending to participate. If a group of 5 people want to set up a meeting, they do not need to clear their preferred conferencing tool by a professional accessibility auditing organization, they just need to clear it with each other.

index.bs
fantasai: I think you missed the other half of the edit on line 680?

dwsinger (approved):

index.bs
brewerj: This wording is an improvement over previous wording which did not call out the user needs of people with disabilities. Nevertheless I'm still not picturing a good scenario for this in practice. So, here's the practical part that I'd like to see if we can address. 

Some of the tools that the W3C Community uses have the potential to be especially useful for how we work; yet if they are not accessible, and if there is not a comparable and accessible alternative, then either the accessibility barriers become blockers to inclusive development of standards, or individuals with disabilities may have to initiate blocking usage of a tool. So there's a reluctance to raise this, and often tools don't get fixed for long periods of time.

Given that W3C is a large community, of potential interest to developers wanting to showcase their tools, I'd be interested in language that encourages a more deliberate effort, for a limited time only (for instance, one month maximum) during which W3C groups pro-actively reach out to tool owners and encourage them to fix their accessibility bugs, before removing a tool from the lineup of tools used by a group. 

It's possible that language in the Process Document doesn't need further tweaking, only in the Art of Consensus language that backs this up. But the hoped-for outcome would be an approach for accessibility of tools that gets us a greater variety of accessible tools, more quickly.

nigelmegitt: > We are not scoping this rule to the entire population of the world (as we are with the deliverables) but rather to the people who are actually intending to participate.

Possibly reiterating or echoing @brewerj's point, the decision on whether or not to participate might be at least partly based on the accessibility of the tools being used. We need to find a path to enough proactivity in choosing accessible tools so that the 6th person who _would_ join that group of 5 in @fantasai 's example is not excluded from doing so by the 5 agreeing amongst themselves to using Inaccessible Meeting Tool (inc).

dwsinger: I am hopeful that people (notably team) will be proactive and guide groups to accessible tools. The purpose of process rules is to give the team and others the formal standing to say "not OK", which we currently lack. There is always a struggle and a balance to choose between utility — and exploring the cutting edge — and i18n and accessibility, and I don't think we should be excessively prescriptive for fear that we'll end up being able to use nothing, and be miles away from using the best tools. Should is right here, I think.

nigelmegitt: @dwsinger I understood this mini-thread to be about the text "by all who wish to participate," rather than "should".

Does this overlap with the CEPC, if we're worried about inclusion and what is "not OK"? Do we want/need to say something in both places?

dwsinger: Sorry yes, but I think that scoping to the audience is also right. "This tool does not work with right-to-left text" could be an i18n problem – but if the group has no users who regularly work in RTL, should we really stop the WG from using it? I think not. We should report the issue to the toolmaker and try/hope to get it fixed before our first RTL user arrives, for sure. The problem is that if we restrict ourselves to perfect tools, we risk finding ourselves miles from the leading edge and current tooling. The purpose of the rule is so that when a user arrives who can't use something, we are obliged to try to fix it or change — if we can. We don't have control over all tools, alas. The rule says that that user has a legitimate request for action that cannot be ignored (which is a huge improvement over the status quo).

samuelweiler:
The parentheses don't matter to me one way or the other.  It sounds like there's a bigger discussion to be had re: the language on the whole.

TzviyaSiegman:
Because this discussion is happening in multiple threads and in a few forks of this commit, I'm unsure what the proposed language is at this point. I support @brewerj 's proposal "Any tooling used by the group for producing its documentation and deliverables or for official group discussions should be accessible for people with disabilities." (I am not sure about mentioning cost, but I'll leave that to others to decide.)

frivoal:
> I'm unsure what the proposed language is at this point.

https://github.com/w3c/w3process/commit/9c7a727a649dd4f37c3cc79124f81b95a81c72de

>  I support @brewerj 's proposal

The problem with that proposal is that it only requires accommodating for people with disabilities, instead of the existing and proposed wording which require accommodating for anyone. The proposed edit clarifies and emphasizes that "anyone" includes people with disabilities, which is a good idea. But I don't think it would be acceptable to narrow this requirement so that it would become OK to use a windows only tool even if someone who only has a mac wants to join, or so that it would become OK to use tools that aren't available in certain countries even if people from those countries are asking to be included.

dwsinger:
Yes, we have to consider at least international accessibility (both linguistic and access blocks)

LJWatson:
> The problem with that proposal is that it only requires accommodating for people with disabilities, instead of the existing and proposed wording which require accommodating for anyone. The proposed edit clarifies and emphasizes that "anyone" includes people with disabilities, which is a good idea. But I don't think it would be acceptable to narrow this requirement so that it would become OK to use a windows only tool even if someone who only has a mac wants to join, or so that it would become OK to use tools that aren't available in certain countries even if people from those countries are asking to be included.

I tend to agree. I also think it avoids a narrow definition of disability because it implicitly encompasses temporary and situational disabilities, as well as permanent disabilities.

Issue #535, opened 2021-05-21
W3C Statement Comment from APA Chairs


JaninaSajka:
The Co-Chairs of the Accessible Platform Architectures (APA) Working Group support the proposal to establish a mechanism to elevate some W3C Working Group Notes to a status that puts the entire W3Con record in support of that document:

https://www.w3.org/Consortium/Process/Drafts/#memo

We suggest providing clear rationale for what content should become a working group note, as opposed to a W3C "Statement" in addition to clarifying the process for each.

Frankly, we would prefer to extend  the existing W3C Notes definitions instead of introducing an entirely new "Statements" category  ... Introducing a third TR category is likely to create more rather than less confusion for the public who are not closely aware of W3C processes. We suggest resurrecting the "W3C Note" variant to indicate that Notes have been reviewed and approved. It would be important to clarify the difference in the Note masthead, in a manner that would be easily seen. Taking that opportunity to redesign the Note boilerplate to make Notes more clearly different from Recomendations would help further reduce potential confusion.

We further advise:

* All potential Notes be first published as Draft Notes for no less than 30 days to provide a reasonable time frame for horizontal (and general) review. We should always leverage organization strengths, and horizontal review is one distinguishing value of the W3C brand. The "Draft Note" status would be an appropriate time for horizontal review regardless of whether the
content is destined to be a working group note, or a W3C publication, though horizontal review is especially critical in the latter instance.

* The current github-based checkoff mechanism for horizontal review could be used to ascertain sign-off by horizontal review groups before a Note is published. We've found it to be an effective tracking mechanism.

* Groups should be advised to expect that it may be relevant to include horizontal review guidance, e.g. an "Accessibility Considerations" section (where relevant).
 
Janina Sajka & Becky Gibson, CoChairs
APA

dwsinger:
Hi Janina

I think there may be some misunderstanding here. Notes are already defined as the output of a WG. So what you are asking for — that we use the existing Note mechanism — is what we're doing. However, the question came up what we should call a Note that gets W3C approval, and we felt that causing it Note might cause confusion with WG Notes. We had a long debate over the name, wanting something clear but generic, and Statement for the name of a WG elevated to W3C status was our best so far. We're open to better!

We also agree that we have an urgent need for better branding of works in progress vs. completed, consensus documents of a working group (WDs) vs. other people's drafts, Notes vs. WDs, and Statements vs. Recommendations. That work is under way (but not a process question).

JaninaSajka:
[...]

PR #541, opened 2021-06-11
Clarification about horizontal review

frivoal:
See #535

If this (or a simple tweak based on this) is acceptable, I suspect we can land it this year. We probably should be doing more in the long term to clarify HR and wide review, but *that* is significantly harder and doesn't fit in a last minute tweak.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2502,7 +2502,9 @@ Wide Review</h5>
 	and tracking those comments
 	and the [=Working Group=]'s responses,
 	is generally a good practice which would often be considered positive evidence of wide review.
-	[=Working Groups=] <em class="rfc2119">should</em> announce to other W3C Working Groups
+	[=Working Groups=]
+	<em class="rfc2119">should</em> follow the <a href="https://www.w3.org/Guide/process/charter.html#horizontal-review">W3C Horizontal Groups</a>’ review processes [[HREVIEW]],
+	and <em class="rfc2119">should</em> announce to other W3C Working Groups
 	as well as the general public,
 	especially those affected by this specification,
 	a proposal to enter [=Candidate Recommendation=]
@@ -5485,6 +5487,11 @@ Changes since earlier versions</h3>
 	"TRANSITION": {
 		"href": "https://www.w3.org/Guide/transitions",
 		"title": "Organize a Technical Report Transition",
+		"publisher": "W3C"
+	},
+	"HREVIEW": {
+		"href": "https://www.w3.org/Guide/documentreview/",
+		"title": "How to do Wide Review",
 		"publisher": "W3C"
 	}
 }


dwsinger (approved):

index.bs
```diff
@@ -5486,6 +5488,11 @@ Changes since earlier versions</h3>
 		"href": "https://www.w3.org/Guide/transitions",
 		"title": "Organize a Technical Report Transition",
 		"publisher": "W3C"
+	},
+	"HREVIEW": {
+		"href": "https://www.w3.org/Guide/documentreview/",
+		"title": "How to do Wide Review",
```
nigelmegitt, line 5494: It's super confusing to have a term definition of "Horizontal Review" that points to a document about Wide Review. See also w3c/documentreview#12. Any reason why not to extend the change in this pull request from Horizontal Review to Wide Review?

nigelmegitt (changes requested):

index.bs
frivoal: I think the issue raised in https://github.com/w3c/documentreview/issues/12 is right, and that the title of that document is wrong. It claims to be about wide review, but really is about horizontal review. For this change, we need a reference to a document that explains how horizontal review is done, and that document is it, even if it has a weird title.

nigelmegitt: For the record, I disagree regarding w3c/documentreview#12. In my view there should be a document about wide review, and be clear in that document about the subset of wide review that is horizontal review. Conflating them or forgetting about the "inconvenient" part of wide review that is not horizontal review is, in my view, a mistake.

plehegar (approved):

dwsinger (changes requested):

plehegar:
I provided https://github.com/w3c/documentreview/pull/21 to improve our wide review guidance.

index.bs
frivoal: Given that that document is still a moving target with disagreements about what it ought to become, I'm dropping that reference for now, and keeping the rest as is.

plehegar:
why drop the reference? /Guide gets updated continuously in any case.

frivoal:
> why drop the reference? /Guide gets updated continuously in any case.

As I mentioned in the comment earlier: 

> Given that that document is still a moving target with disagreements about what it ought to become, I'm dropping that reference for now, and keeping the rest as is.

This is meant to be a small, non controversial last minute editorial clarification, while we wait for another cycle to do a deeper clean up of HR and wide review. Given that there's ongoing discussion about what the scope of that document ought to be, it seems premature to me to link to it from the process.

index.bs
```diff
@@ -2502,7 +2502,9 @@ Wide Review</h5>
 	and tracking those comments
 	and the [=Working Group=]'s responses,
 	is generally a good practice which would often be considered positive evidence of wide review.
-	[=Working Groups=] <em class="rfc2119">should</em> announce to other W3C Working Groups
+	[=Working Groups=]
+	<em class="rfc2119">should</em> follow the <a href="https://www.w3.org/Guide/process/charter.html#horizontal-review">W3C Horizontal Groups</a>’ review processes,
+	and <em class="rfc2119">should</em> announce to other W3C Working Groups
```
dwsinger, line 2507: Perhaps we can say that Working Groups should follow the [review guidelines](https://www.w3.org/Guide/process/charter.html#horizontal-review) and in particular should enable Horizontal Review by announcing to other W3C Working Groups

nigelmegitt: Isn't that what it says on the lines you highlighted, @dwsinger ? I'm having trouble decoding your comment! Is it just the delta "enable HR by announcing" vs just "should announce"?

dwsinger: No, at the moment it references a document about Charters with a link that on text about Horizontal Groups. It's very mixed and could be confusing.

nigelmegitt: Thanks that's clear after you edited the comment.

samuelweiler:
The link should be to the [document HR guidance](https://www.w3.org/Guide/documentreview/#how_to_get_horizontal_review,) not the charter HR guidance

frivoal:
Good point. The link went to the charter HR guidance because for quite some time, that was the only documentation which could reliably point you to all the HR groups and their requirements, even if it wasn't its primary purpose. I suspect that swapping it out for the one you suggested would work, and it also sidesteps the question of the overall scope of that document (and its title) not being set in stone yet, which would be an issue if we were to add it as a bibliography entry. By linking directly to the relevant section of that document we no longer need to be blocked by the rest being unstable.

frivoal:
Updated the PR to reflect the proposal in the past two comments

PR #549, opened 2021-06-24
Use the username 'w3cbot' to interact with GitHub

plehegar:
The [GITHUB_TOKEN](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret) needs a valid username but it seems it could be any user, not just [GITHUB_ACTOR](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context).
Email address can be anything.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -21,14 +21,14 @@ jobs:
       - name: Setup python 3.7
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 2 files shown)


plehegar:
Python 3.8 is used by bikeshed. See https://github.com/tabatkins/bikeshed/blob/main/.github/workflows/update-tests.yml

PR #556, opened 2021-08-09
Make the CEO rather than the Director responsible for MoUs

frivoal:


chaals (changes requested):
The PR should update https://www.w3.org/2020/Process-20200915/#ACAppeal which describes MoU as a director's decision.

plehegar:
> The PR should update https://www.w3.org/2020/Process-20200915/#ACAppeal which describes MoU as a director's decision.

https://www.w3.org/Consortium/Process/Drafts/director-free/#ACAppeal doesn't mention the Director anymore, so I believe the proposed change is fine and enough.

plehegar (approved):

frivoal:
@chaals is  right. Some other change in the director-free branch cleans that up, but those changes haven't landed yet on the main branch, which means that if we land this one first, we'd have an inconsistent state. I'll find that bit and include it here as well.

jeffjaffe:
Thanks, David.  I agree with this intent.  But I'm not sure that documenting it in GH is sufficient.  Once we close this issue, people will forget that the issue exists.  But the Process Document is forever :).

I have raised https://github.com/w3c/AB-memberonly/issues/105

PR #571, opened 2021-08-27
Require document license to go through formal approval for updates

frivoal:
Ensures all policies binding on members either are dated or go through AC approval.

This addresses the part of @jwrosewell's formal objection on normative dependencies. With this in place, the normative references are as follows:

Covered by section 11 formal approval:
* [CEPC]
* [DOC-LICENSE] (just added)
* [PATENT-POLICY]

Not binding on Members via Member agreement:
* [COLLABORATORS-AGREEMENT] (applies to signers only)
* [CONFLICT-POLICY] (applies to Team only, not a concern for members)
* [PUBRULES] (applies to documents, not people; merely documents what is formally defined by the Process as being the Team's judgement call on publishability)

No need to be dated, because RFCs [cannot change after publication](https://www.rfc-editor.org/faq/#errata):
* [RFC2119]
* [RFC3797]

@jwrosewell, can you review and let us know whether this change addresses your comment? (See https://github.com/w3c/w3process/pull/572 for the issue about precedence.)

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/571.html" title="Last updated on Aug 31, 2021, 5:59 AM UTC (bd4ebc7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/571/b9d22e7...frivoal:bd4ebc7.html" title="Last updated on Aug 31, 2021, 5:59 AM UTC (bd4ebc7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -66,7 +66,7 @@ Boilerplate: repository-issue-tracking off
 </pre>
 <pre class=biblio>
 {
-	"PATENT-POLICY": {
+	"PATENT-POLICY-2020": {
 		"href": " https://www.w3.org/Consortium/Patent-Policy-20200915/",
 		"title": "The W3C 2020 Patent Policy"
 	}
@@ -2790,9 +2790,9 @@ The W3C Recommendation Track</h3>
 
 	This Process defines certain [=Recommendation Track=] publications as <dfn>Patent Review Drafts</dfn>.
 	Under the 2004 (updated in 2017) Patent Policy,
-	these correspond to “Last Call Working Draft” in the Patent Policy [[!PATENT-POLICY-2017]];
+	these correspond to “Last Call Working Draft” in the Patent Policy [[!PATENT-POLICY-2004]];
 	Under the 2020 Patent Policy,
-	these correspond to “Patent Review Draft” in the Patent Policy [[!PATENT-POLICY]].
+	these correspond to “Patent Review Draft” in the Patent Policy [[!PATENT-POLICY-2020]].
 
 	W3C <em class="rfc2119">may</em> <a href="#tr-end">end work on a technical report</a> at any time.
 
@@ -5474,7 +5474,7 @@ Changes since earlier versions</h3>
 		"title": "The W3C Patent Policy",
 		"publisher": "W3C"
 	},
-	"PATENT-POLICY-2017": {
+	"PATENT-POLICY-2004": {
 		"href": "https://www.w3.org/Consortium/Patent-Policy-20170801/",
 		"title": "The W3C 2004 Patent Policy, Updated 2017",
 		"publisher": "W3C"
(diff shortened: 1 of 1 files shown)


frivoal:
> don't we expect "patent review draft" to remain a current feature of the patent policy in future revisions?

We do. But this is in a descriptive sentence that characterizes what the term means in a particular version, in contrast with another particular version. I expect it's meaning won't change, but it's hard to state what it will mean in versions that have not yet been written. Further, this is not a change: prior to this change, the link in this paragraph is already to the dated URL, and after the change, it continues to be.

All other references to the patent policy, since they're not discussing a change introduce in a particular version, have been updated to the undated URL.

> To ensure lack of ambiguity, I would recommend that anything that we change which clarifies patent policy references should get a PSIG sign-off before we finalize.

There is no change or clarification in the text, and all links, after redirects, still point to the same place. The difference is linking to the undated URL that points to the latest version instead of the same version through a dated URL (with the one exception discussed above).

jeffjaffe:
> > To ensure lack of ambiguity, I would recommend that anything that we change which clarifies patent policy references should get a PSIG sign-off before we finalize.
> 
> There is no change or clarification in the text, and all links, after redirects, still point to the same place. The difference is linking to the undated URL that points to the latest version instead of the same version through a dated URL (with the one exception discussed above).

Agree vis-a-vis this current change.  But in terms of determining precedence between PP and Process document, for example, it should get a PSIG review.

dwsinger:
> > don't we expect "patent review draft" to remain a current feature of the patent policy in future revisions?
> 
> We do. But this is in a descriptive sentence that characterizes what the term means in a particular version, in contrast with another particular version. 

That's where we disagree; it doesn't apply to a particular version, it applies to this and all future versions.

> I expect its meaning won't change, but it's hard to state what it will mean in versions that have not yet been written. 

That's true for the entire document.

I don't want to introduce subtle errors if we *forget* to update a dated reference. That's pretty safe for referring to history, very unsafe for referring to the future.

frivoal:
@dwsinger Ok, updated accordingly. How is that now?

fantasai:
@jeffjaffe I defer to @wseltzer’s judgement here and in #571. Personally I think informing PSIG about it (rather than getting their explicit signoff) should be enough, as I doubt they will have any concerns that Wendy would not raise herself.

jeffjaffe:
> @jeffjaffe I defer to @wseltzer’s judgement here and in #571. Personally I think informing PSIG about it (rather than getting their explicit signoff) should be enough, as I doubt they will have any concerns that Wendy would not raise herself.

I, too, respect Wendy's judgement and doubt that PSIG would have concerns.  But in the past, we have made changes to the process document which touch on the patent policy without PSIG sign-off (since we seemed to have a ProcessCG/Member/Team consensus that we were OK), only to be criticized by PSIG later because they did not agree.  So it is at least a simple courtesy to allow them to weigh in.

dwsinger:
The courtesy might be to have the PSIG chairs be informed and make the call over how much signoff they want.

PR #585, opened 2021-11-09
Make the Team responsible for Group creation and Closure

frivoal:
Make the Team responsible for Group creation and Closure, rather than the Director. This includes some slight rephrasing to make it explicit that these are decisions, so that the usual recourses to are available.

(We might want to further adjust the process for creating or closing groups, but that would be a separate issue.)

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/585.html" title="Last updated on Nov 9, 2021, 6:18 AM UTC (5f7e1f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/585/15d3ab4...frivoal:5f7e1f7.html" title="Last updated on Nov 9, 2021, 6:18 AM UTC (5f7e1f7)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/585.html" title="Last updated on Jan 12, 2022, 4:51 PM UTC (b884c8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/585/15d3ab4...frivoal:b884c8b.html" title="Last updated on Jan 12, 2022, 4:51 PM UTC (b884c8b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1800,7 +1800,7 @@ Content of a Charter</h3>
 <h3 id="CharterReview">
 Advisory Committee Review of a Charter</h3>
 
-	The Director <em class="rfc2119">must</em> solicit <a href="#ACReview">Advisory Committee review</a>
+	The [=Team=] <em class="rfc2119">must</em> solicit <a href="#ACReview">Advisory Committee review</a>
 	of every new or substantively modified Working Group or Interest Group charter,
 	except for either:
 	* a charter extension
@@ -1827,7 +1827,7 @@ Advisory Committee Review of a Charter</h3>
 	Such a request must be submitted with a Member's comments
 	in response to the Call for Review.
 	Upon receipt of any such request,
-	the Director <em class="rfc2119">must</em> ensure
+	the [=Team=] <em class="rfc2119">must</em> ensure
 	that the Call for Participation for the Working Group
 	occurs at least 60 days
 	after the Call for Review of the charter.
@@ -1835,11 +1835,14 @@ Advisory Committee Review of a Charter</h3>
 <h3 id="cfp">
 Call for Participation in a Chartered Group</h3>
 
-	After [=Advisory Committee review=] of a [=Working Group=] or [=Interest Group=] [=charter=],
-	the Director <em class="rfc2119">may</em> issue a Call for Participation to the Advisory Committee.
+	Deciding whether to adopt a proposed [=Working Group=] or [=Interest Group=] [=charter=]
+	is a [=W3C Decision=].
 	Charters <em class="rfc2119">may</em> be amended based on review comments
+	per [[#ACReviewAfter]]
 	before the Call for Participation.
 
+	If the decision is to charter the group,
+	the [=Team=] <em class="rfc2119">must</em> issue a Call for Participation to the Advisory Committee.
 	For a new group, this announcement officially creates the group.
 	The announcement <em class="rfc2119">must</em> include a reference to the <a href="#WGCharter">charter</a>,
 	the name(s) of the group's [=Chair=](s),
@@ -1864,9 +1867,12 @@ Call for Participation in a Chartered Group</h3>
 <h3 id="charter-extension">
 Charter Extension</h3>
 
-	To extend a Working Group or Interest Group charter
-	with no other substantive modifications,
-	the Director announces the <dfn id=dfn-charter-extension export lt="charter extension">extension</dfn> to the Advisory Committee.
+	The [=Team=] <em class="rfc2119">may</em> decide
+	to extend a Working Group or Interest Group charter
+	with no other substantive modifications.
+	The [=Team=] <em class="rfc2119">must</em> announce
+	such <dfn id=dfn-charter-extension export lt="charter extension">extensions</dfn>
+	to the [=Advisory Committee=].
 	The announcement <em class="rfc2119">must</em> indicate the new duration.
 	The announcement <em class="rfc2119">must</em> also include rationale for the extension,
 	a reference to the <a href="#WGCharter">charter</a>,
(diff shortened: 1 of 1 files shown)


mnot:
This is not appropriate. Why are you suggesting we go in this direction?

frivoal:
This is not a new direction, it is a more realistic description of what has been happening for years already. I am not saying that we shouldn't also make deeper changes to how groups are started, but my understanding of one of the AB's goal for this year is to get to a point where we have a Process that can operate without a Director, and here acknowledging what has been the reality gets us there.

I absolutely think we should make chartering related things better, in much deeper ways that this PR does, but I don't think it's appropriate to continue pretending that we cannot possibly let the current director go because then we wouldn't be able to create groups, when we've been doing it without him for years.

Besides, the Team is accountable all powerful and without checks and balances here, the decisions it can make here can be formally objected to, so we have recourse.

index.bs
```diff
@@ -1880,14 +1886,14 @@ Charter Extension</h3>
...
-	an <a href="#ACAppeal">Advisory Committee Appeal</a> against a Director's decision
+	an <a href="#ACAppeal">Advisory Committee Appeal</a> against a [=Team decision=]
 	regarding the extension of a Working Group or Interest Group charter.
 
 <h3 id="GeneralTermination">
 Chartered Group Closure</h3>
 
 	A [=Working Group=] or [=Interest Group=] charter specifies a duration for the group.
-	The [=Director=] <em class="rfc2119">may</em> decide to close a group
+	The [=Team=] <em class="rfc2119">may</em> decide to close a group
```
dwsinger, line 1896: This one I question. The team could advise, perhaps, but decide?

index.bs
```diff
@@ -3805,7 +3811,7 @@ Abandoning an Unfinished Recommendation</h5>
 	This can happen if
 	the [=Working Group=] decided
 	to abandon work on the report,
-	or the [=Director=] required the [=Working Group=]
+	or the [=Team=] required the [=Working Group=]
```
dwsinger, line 3814: I similarly doubt that the Team should be making absolute requirements of WGs.

index.bs
jeffjaffe: If the team is advising instead of deciding, then we need a process that clarifies who the team is advising so they may make a decision.

jwrosewell:
See [RACI](https://racichart.org/the-raci-model/) decision making. My preference is to add RACI to the Process and then align all decisions.

LJWatson:
The AC decides if a WG is created and so it could also decide if one closes.

chaals:
@LJWatson wrote

> The AC decides if a WG is created and so it could also decide if one closes.

That makes good sense in principle, but the implementation gets tricky, first because we need a mechanism to propose closing a group (which would likely involve the team), and second because the social interactions are unlikely to be only positive and friendly and constructive.

mnot:
> The AC decides if a WG is created and so it could also decide if one closes.

I'm not happy with this -- given the relative inactivity of the AC, and its composition, its' not really a good custodian of the technical roadmap for activity at the W3C. Better to invest this authority with a body that's accountable to the AC, like the TAG.

caribouW3:
> The AC decides if a WG is created and so it could also decide if one closes.

The AC reps can already oppose to charter renewal, which is equivalent to closure.
Would there be a reason that could make it really urgent to close a WG right away in the middle of the charter duration?

LJWatson:
@caribouW3
>Would there be a reason that could make it really urgent to close a WG right away in the middle of the charter duration?

That's a good question. Do we have any data about how often and why early closure has been needed in the past?

LJWatson:
@mnot 
>I'm not happy with this -- given the relative inactivity of the AC, and its composition, its' not really a good custodian of the technical roadmap for
activity at the W3C. Better to invest this authority with a body that's accountable to the AC, like the TAG.

It's a fair point about AC engagement, though if the AC is not the appropriate custodian of the technical roadmap, does that also include deciding which WG are created do you think?


As an aside, I'm worried that the TAG and/or AB are often suggested as an alternative to many of TimBL's erstwhile responsibilities though, when they're both already extremely busy.

mnot:
> It's a fair point about AC engagement, though if the AC is not the appropriate custodian of the technical roadmap, does that also include deciding which WG are created do you think?

I think the AC has a mostly supervisory / oversight role, to add legitimacy to the decisions made. The actual decisions need to be made by a smaller, more targeted body that 'owns' those decisions (including creation).
 
> As an aside, I'm worried that the TAG and/or AB are often suggested as an alternative to many of TimBL's erstwhile responsibilities though, when they're both already extremely busy.

People keep on saying that. In my mind creating and approving work *has* to be owned by an elected body, and if not the TAG, who? OTOH I don't see how spec reviews require the authority of an elected body -- the TAG can and should delegate some of this work if necessary.

caribouW3:
> @caribouW3
> 
> > Would there be a reason that could make it really urgent to close a WG right away in the middle of the charter duration?
> 
> That's a good question. Do we have any data about how often and why early closure has been needed in the past?

Maybe in a few cases we closed a WG before the end of the charter because there was no longer any activity after reaching REC + some maintenance time (6 months or so). I can't recall any case of closure for another reason.

frivoal:
Early closure may also be necessary due to a Patent Advisory Group (see point 4 of https://www.w3.org/Consortium/Patent-Policy-20200915/#sec-PAG-conclude-possible). Not that this is expected to be common, mind you.

LJWatson:
>People keep on saying that. In my mind creating and approving work has 
to be owned by an elected body, and if not the TAG, who? OTOH I don't 
see how spec reviews require the authority of an elected body -- the TAG 
can and should delegate some of this work if necessary.


That could help alleviate the impending problem, yes.

dwsinger:
If teh AC has to approve group creation and renewal, I don't see a problem with the AC approving early closure.

The only issue I see is if a group has to be closed for some reason (can't think of one), and it's not an option for the AC to say "no, keep it open". Can that occur?

So sure the team can propose a group be closed early, and I doubt the AC will be very active in disagreeing, but they should be given the option.

frivoal:
Approved on the January 13 call, with the modification that AC review is required in the (exceptional) case of  forcible group closure or work termination.

nigelmegitt:
Just reviewed this pull request since I realised it was relevant to #653. I note that there's no linked issue and none of the discussion in the comments is about the removal of the two specific criteria under which the Director (formerly) was permitted to propose closure of a group. I just opened #685 to request reinstatement of those criteria.

PR #586, opened 2021-11-09
Let the Team rather than the Director handle progression on TR

frivoal:
Note that:
* transitions on TR are not initiated by the Team
* there's always an opportunity to object if the Team does get it wrong

Therefore, even though the criteria that we're asking the Team to judge are somewhat fuzzy, this is not a particularly strong delegation of authority to the Team. (We can also consider clarifying the fuzzy criteria in /guide or in a separate issue.)

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/586.html" title="Last updated on Nov 9, 2021, 6:30 AM UTC (a947b6c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/586/15d3ab4...frivoal:a947b6c.html" title="Last updated on Nov 9, 2021, 6:30 AM UTC (a947b6c)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/586.html" title="Last updated on Jul 13, 2022, 2:56 PM UTC (4387d8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/586/15d3ab4...frivoal:4387d8c.html" title="Last updated on Jul 13, 2022, 2:56 PM UTC (4387d8c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2547,7 +2547,7 @@ Wide Review</h5>
 	early enough that comments and suggested changes
 	can still be reasonably incorporated in response to the review.
 	Before approving transitions,
-	the [=Director=] will consider who has been explicitly offered
+	the [=Team=] will consider who has been explicitly offered
 	a reasonable opportunity to review the document,
 	who has provided comments,
 	the record of requests to and responses from reviewers,
@@ -2788,7 +2788,8 @@ The W3C Recommendation Track</h3>
 
 	W3C <em class="rfc2119">may</em> <a href="#tr-end">end work on a technical report</a> at any time.
 
-	The Director <em class="rfc2119">may</em> decline a request to advance in maturity level,
+	The [=Team=] <em class="rfc2119">may</em> [=Team Decision|decide=]
+	to decline a request to advance in maturity level,
 	requiring a [=Working Group=] to conduct further work,
 	and <em class="rfc2119">may</em> require
 	the specification to return to a lower <a href="#maturity-levels">maturity level</a>.
@@ -3023,7 +3024,7 @@ Implementation Experience</h4>
 	of each feature of the specification will be realized.
 	While no exhaustive list of requirements is provided here,
 	when assessing that there is <dfn id="adequate-implementation">adequate implementation experience</dfn>
-	the [=Director=] will consider (though not be limited to):
+	the [=Team=] will consider (though not be limited to):
 
 	<ul>
 		<li>
@@ -3068,7 +3069,7 @@ Advancement on the Recommendation Track</h4>
 			<em class="rfc2119">must</em> record the group's decision to request advancement.
 
 		<li>
-			<em class="rfc2119">must </em> obtain [=Director=] approval.
+			<em class="rfc2119">must </em> obtain [=Team=] approval.
 
 		<li>
 			<em class="rfc2119 ">must</em> publicly document all new features
@@ -3121,7 +3122,7 @@ Updating Mature Publications on the Recommendation Track</h4>
 	Certain requests to re-publish a specification
 	within its current maturity level
 	(called <dfn export>Update Requests</dfn>)
-	require [=Director=] approval.
+	require [=Team=] approval.
 	For such [=update requests=], the Working Group:
 
 	<ul>
@@ -3132,7 +3133,7 @@ Updating Mature Publications on the Recommendation Track</h4>
 			<em class="rfc2119">must</em> show that the changes have received [=wide review=].
 
 		<li>
-			<em class="rfc2119">must </em> obtain [=Director=] approval,
+			<em class="rfc2119">must </em> obtain [=Team=] approval,
 			or fulfill the criteria for [[#streamlined-update]].
 
 		<li>
@@ -3169,7 +3170,7 @@ Updating Mature Publications on the Recommendation Track</h4>
 	</ul>
 
 	There is usually a formal review meeting
-	to ensure the requirements have been met before [=Director=]'s approval is given.
+	to ensure the requirements have been met before [=Team=]'s approval is given.
 
 	Note: [=Update request=] approval is expected to be fairly simple
 	compared to getting approval for a [=transition request=].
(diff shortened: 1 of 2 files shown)


dwsinger:
I think we need to discuss this one. I am not 100% comfortable with the phrase "team approval". I am fine with the team assessing whether the criteria have, or have not, been met; I am not sure we should convert that into an absolute authority to decide. 99% of the time it's uncontroversial, I agree.

chaals:
The RACI model doesn't seem bad. But it doesn't address with the issue here which is pretty simple in RACI terms: who is responsible / accountable?

dwsinger:
I wonder if we can advance this by using the idea of proposal/action, i.e. that the team communicates "if there is no objection from an AC Rep by date X, we propose to decide to do Y; if objections are received, we will work with the objectors to try to resolve their concerns by consensus, and if that fails, we will hold an AC ballot". 99% of the time there will be silence.

The reason I like this is that reversing a decision made is both awkward and "raises the ante", whereas someone saying "um, no, not a good idea" before the  fact leaves room for debate and consensus.

jeffjaffe:
> I wonder if we can advance this by using the idea of proposal/action, i.e. that the team communicates "if there is no objection from an AC Rep by date X, we propose to decide to do Y; if objections are received, we will work with the objectors to try to resolve their concerns by consensus, and if that fails, we will hold an AC ballot". 99% of the time there will be silence.

Of late we have seen an increase in the frequency of formal objections.  I'm not loving a proposal that could encourage even more objections.  TR progression for the most part is a seamless part of our process.  Proposals that add complexity to solve imagined problems will not necessarily help the agility of the organization.

dwsinger:
Sorry, maybe I should not have used the word "objection", as I didn't mean FO. "We propose to extend the charter unless concerns are expressed before Dec 15th" is what I am thinking; much lighter weight.

LJWatson:
I think the criteria for sufficient review (both wide and horizontal) are [documented well enough](https://www.w3.org/Guide/documentreview/) for that part of this PR to be non-controversial.

The rest I'm less sure about because the criteria are less well documented, and hence I share @dwsinger's caution about the word "approve".

I cannot think of an alternative to the Team taking on this function, at least one that will not add to the already overloaded alternatives we're currently exploring. I wonder therefore if requiring the rationale for any such decisions might work?

As in, if the Team decided that transition should go ahead despite insufficient implementation experience, the rationale for that should be documented and presented as part of the AC review.

If this is already included somewhere, apologies. Reading diffs is not my favourite thing with a screen reader.

dwsinger:
The PR has many instances of 'require/obtain Team approval'. I am not comfortable with that; it makes it an open judgment question. If we mean 'requires/must obtain the Team's assent that the requirements ('must' statements in this process) have been met' that's fine (and the team's job). That's still a judgment call, but it's 'closed' to only the Process requirements.

mnot:
I share the discomfort that David expresses. This might make more sense if the Team's role were much more prescribed -- in terms of the guidelines/criteria for the decision, the discretion they're able to exercise, transparency of their decisions (or failure to decide), ability to object (both to the decision and the lack of one), etc. @frivoal you characterise it as a "check" which sounds reasonable, but that's not what the proposed text conveys.

plehegar:
Discussed on [June 8](https://www.w3.org/2022/06/08-w3process-minutes.html).

Suggested approach to move forward was a 2 steps:

1. replace the Director by the Team (FOs would go to the Council in any case)
2. look into fixing the pre-AC review period to avoid surprising the AC when starting a review. (no proposal yet).

Step 1 is now waiting on Florian to create a new pull request. That pull request should take into account [@dwsinger 's comment](https://github.com/w3c/w3process/pull/586#issuecomment-1150063659) at the minimum.

We'll come back to step 2 after that (in the fall most probably).

frivoal:
Did a second pass on this PR.

@LJWatson 's https://github.com/w3c/w3process/pull/586#issuecomment-967144665 was already covered in the text (see https://github.com/w3c/w3process/pull/586/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985L3537).

@dwsinger's concerns in  https://github.com/w3c/w3process/pull/586#issuecomment-1150063659 should now also be addressed, because in the key places ( and ), it's made clear what the team's approval or rejection is supposed to be based on: expectations from the Process (such as wide review, implementation exprience, etc), not individual team members opinions.

Also added AB+TAG's approval to the Team/Director (now Team)'s pre-existing ability to forcefully republish a spec at a lower maturity level.

I suggest reviewing the new commits one at a time, as some text is moved around slightly, and it's probably easier to review step by step.

index.bs
```diff
@@ -3132,8 +3136,16 @@ Updating Mature Publications on the Recommendation Track</h4>
 			<em class="rfc2119">must</em> show that the changes have received [=wide review=].
 
 		<li>
-			<em class="rfc2119">must </em> obtain [=Director=] approval,
+			<em class="rfc2119">must </em> obtain [=Team=] approval,
 			or fulfill the criteria for [[#streamlined-update]].
+			[=Team=] approval (a [=Team decision=]),
```
mnot, line 3141: spurious comma?

index.bs
```diff
@@ -3169,7 +3181,7 @@ Updating Mature Publications on the Recommendation Track</h4>
 	</ul>
 
 	There is usually a formal review meeting
-	to ensure the requirements have been met before [=Director=]'s approval is given.
+	to ensure the requirements have been met before [=Team=]'s approval is given.
```
dwsinger, line 3184: I would prefer we did not use the phrase 'team approval'. I'm struggling with the phrasing "before the team verifies that the process requirements are met" is awfully wordy…

frivoal:
Follow up in https://github.com/w3c/w3process/issues/609 about "approval" vs "verification"

plehegar:
Agreed for merge at https://www.w3.org/2022/07/13-w3process-minutes.html#t04

Issue #552, opened 2021-07-01
Archiving discussion attachments


fantasai:
The Process 2021 draft requires:

> Official meeting minutes and other records of decisions made must be archived by W3C for future reference; and other persistent text-based discussions sponsored by the group, pertaining to their work and intended to be referenceable by all group members should be. This includes discussions conducted over email lists or in issue-tracking services or any equivalent fora.

But in https://github.com/w3c/w3process/issues/551 it came up that we aren't very clear about whether documents referenced in those discussions, like slide decks and proposals, ought to be archived as well. To the extent that a discussion doesn't make sense without it, they probably should (but not MUST), I think. Do we want to make a clarification about that?

dwsinger:
This could easily be a long discussion, and I am hesitant to put it in the process. Some decisions stand alone:

* Decided to advance the 27th June Draft of Whatits and Thingummies to FPWD

Others need the supporting material:
* Decided to adopt Göthe GMBH proposal for the solution to the 2nd problem

needs a copy of the proposal and identification of what the 2nd problem is.

How we say this in process text I am not sure.

fantasai:
@dwsinger Something like “Materials referenced from discussions and necessary to understand them should be available at a stable URL at the same level of confidentiality as the discussion minutes.” ?

dwsinger:
maybe we just need to add to the end of the paragraph that the records should be sufficiently complete to be comprehensible?

samuelweiler:
I think this adds unnecessary verbiage and we should just skip it.

dwsinger:
it's important that official records are complete enough to be an effective record. decided to adopt X where X cannot be found any more is not an effective record

dwsinger:
We ought to be able to make a requirement that the records are complete enough to be comprehensible. Small edit?

PR #591, opened 2022-01-03
Expect supporting materials to be archived

frivoal:
See #552

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/591.html" title="Last updated on Jan 12, 2022, 4:05 PM UTC (3087640)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/591/636277c...frivoal:3087640.html" title="Last updated on Jan 12, 2022, 4:05 PM UTC (3087640)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -672,7 +672,7 @@ Tooling for Discussions and Publications</h4>
 			Materials referenced from discussions
 			and necessary to understand them
 			should be available at a stable URL
-			at the same level of confidentiality as the discussion minutes.
+			at a level of confidentiality no stricter than the discussion minutes.
 
 			Note: The lack, or loss, of such archives does not by itself
 			invalidate an otherwise-valid decision.
--- a/index.bs
+++ b/index.bs
@@ -671,7 +671,7 @@ Tooling for Discussions and Publications</h4>
 			or any equivalent fora.
 			Materials referenced from discussions
 			and necessary to understand them
-			should be available at a stable URL
+			should be available at a stable URL,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -669,6 +669,10 @@ Tooling for Discussions and Publications</h4>
 			This includes discussions conducted over email lists
 			or in issue-tracking services
 			or any equivalent fora.
+			Materials referenced from discussions
+			and necessary to understand them
+			should be available at a stable URL
+			at the same level of confidentiality as the discussion minutes.
```
TallTed, line 675: It seems this should be somewhat looser; i.e., the level of confidentiality of referenced materials need not be exactly the same as the discussion minutes, as long as it's not more restrictive.  Striving to be clear in one comment -- I mean, as long as anyone who can see the discussion minutes can also see the referenced materials, the specific permissions on the referenced materials are irrelevant.  (Simple example, referenced materials may be accessible without restriction, a/k/a publicly accessible a/k/a world readable, though discussion minutes *might* be confidential to participants or any other subgroup of the world.)

frivoal: Nice catch. Fixed.

TallTed (approved):

samuelweiler:
While this is a good practice, I don't see the need to clutter the process doc with it - particularly with the wiggle room, which is appropriate, this doesn't _need_ to be included.  And as I imagine going through the process doc with a red pen, I think I'd strike it.

PR #614, opened 2022-08-01
editorial: rm undefined COO from informal Note

tantek:
remove "COO" from a list in an informal Note. "COO" is not defined in the document either (in contrast, CEO is in the glossary), so this seems like it was a mistake that made it past review, and thus should be removed from a non-normative Note. if there's an actual use-case for adding "COO" to the Process, that should be raised as a separate issue for proper definition and consideration.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/pull/614.html" title="Last updated on Aug 1, 2022, 8:53 PM UTC (d19ccf5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/614/e4481e8...d19ccf5.html" title="Last updated on Aug 1, 2022, 8:53 PM UTC (d19ccf5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1955,7 +1955,7 @@ Consensus</h4>
 	or <a href="#distributed-meeting">distributed</a>)
 	as well as through <a href=#discussion-archiving>persistent text-based discussions</a>.
 
-	Note: The Director, CEO, and COO have the role of
+	Note: The Director and CEO have the role of
 	assessing consensus within the Advisory Committee.
 
 	The following terms are used in this document


index.bs
```diff
@@ -1955,7 +1955,7 @@ Consensus</h4>
 	or <a href="#distributed-meeting">distributed</a>)
 	as well as through <a href=#discussion-archiving>persistent text-based discussions</a>.
 
-	Note: The Director, CEO, and COO have the role of
+	Note: The Director and CEO have the role of
 	assessing consensus within the Advisory Committee.
```
TallTed, line 1959: As long as this is being massaged, `have the role of assessing consensus` should become either `have the role of consensus assessor` or `perform the task of assessing consensus` ... or something requiring a bit more bikeshedding.

tantek: I have objection to further improving this Note, however this edit seems orthogonal to removal of "COO", so would prefer it be done in another PR, especially since it appears to request some discussion of options.

frivoal (approved):

plehegar (approved):

plehegar:
fyi @swickr

Issue #634, opened 2022-09-21
Improving the definition of consensus


frivoal:
(Moved from #314)

"consensus", in the Process is currently defined as the lack of Formal Objection.

This doesn't always work great, as we occasionally want to speak of consensus in contexts where making formal objections would not be appropriate (such as in the W3C Council addressing formal objections, in a potential appointment committee, or in this catch 22 https://github.com/w3c/w3process/issues/624).

So FOs isn't quite what we're trying to measure here. Consider this section about the Council where we are locally overriding the definition:

> […] in this section, the terms “consensus“, “unanimity”, and “dissent” have slightly different meanings compared to the rest of this document: dissent—and therefore the lack of consensus—occurs when a member of the [W3C Council](https://www.w3.org/Consortium/Process/Drafts/director-free/#w3c-council) opposes a tentative conclusion, even though they cannot register a [Formal Objection](https://www.w3.org/Consortium/Process/Drafts/director-free/#FormalObjection).

I think the better solution may actually be to adjust the general definition of consensus. Beside its formal definition of a lack of formal objection, consensus is a term frequently used in our community to refer to a lack of sustained disagreement. Clearly, filling a formal objection is a sign of sustained disagreement, but in common usage, it isn't necessary to escalate things that much. For instance, we often say that the AB makes decisions by consensus, and often conclude that we do not have consensus when some people indicate that they disagree, yet that doesn't involve Formal Objections.

Note that this would not prevent chairs from making decisions where they can today. The process states:
> In some cases, even after careful consideration of all points of view, a group might find itself unable to reach consensus. The [Chair](https://www.w3.org/Consortium/Process/Drafts/director-free/#GeneralChairs) may record a decision where there is [dissent](https://www.w3.org/Consortium/Process/Drafts/director-free/#def-Dissent) […] so that the group can make progress

I've gone through the Process, and I think the one case where we do want to necessarily gate consensus on the absence of formal objection is AC reviews.

So here's my proposal: 
1. Adjust the definition of consensus (respectively, of dissent) to be based on the lack of (resp. presence of) sustained disagreement, rather than of Formal Objection specifically. An FO always indicates sustained disagreement, but as per current practice, an FO isn't necessary to express dissent. (Mere dislike for a proposed decision, however, doesn't constitute sustained disagreement, as one could still be willing to accept the compromise).
2. Indicate in the section about AC Reviews that in that particular context, Formal Objections are the way to express sustained disagreement.
[...]

PR #635, opened 2022-09-22
Improving the definition of consensus

frivoal:
See #634

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/635.html" title="Last updated on Oct 13, 2022, 9:56 AM UTC (542e7f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/635/2dfe1d2...frivoal:542e7f5.html" title="Last updated on Oct 13, 2022, 9:56 AM UTC (542e7f5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2746,7 +2746,7 @@ Errata Management</h4>
 Candidate Amendments</h4>
 
 	An [=erratum=] <em class="rfc2119">may</em> be accompanied by an informative,
-	<dfn>candidate correction</dfn> approved by the consensus of the [=Working Group=].
+	<dfn>candidate correction</dfn> approved by [=group decision=].
 	When annotated inline,
 	errata--
 	including their [=candidate corrections=]--
--- a/index.bs
+++ b/index.bs
@@ -2024,7 +2024,8 @@ Consensus</h4>
 	</dl>
 
 	Note: A [=Formal Objection=] always indicates sustained disagreement,
-	but isn't necessary to express [=dissent=].
+	but isn't necessary to express it,
+	except in som formal contexts such as [=AC Reviews=].
 	Disagreement with a proposed decision,
 	however, does not always rise to the level of sustained objection,
 	as individuals could be willing to accept a decision
--- a/index.bs
+++ b/index.bs
@@ -2007,7 +2007,7 @@ Consensus</h4>
 		<dd>
 			A substantial number of individuals in the set
 			support the decision
-			and nobody in the set registers a <a href="#FormalObjection">Formal Objection</a>.
+			and there is no sustained objection from anybody in the set.
 			Individuals in the set <em class="rfc2119">may</em> abstain.
 			Abstention is either an explicit expression of no opinion
 			or silence by an individual in the set.
@@ -2020,8 +2020,15 @@ Consensus</h4>
 
 		<dt><dfn id="def-Dissent">Dissent</dfn>:
 		<dd>
-			At least one individual in the set registers a <a href="#FormalObjection">Formal Objection</a>.
+			At least one individual in the set sustains an objection.
 	</dl>
+
+	Note: A [=Formal Objection=] always indicates sustained disagreement,
+	but isn't necessary to express [=dissent=].
+	Disagreement with a proposed decision,
+	however, does not always rise to the level of sustained objection,
+	as individuals could be willing to accept a decision
+	while expressing disagreement.
 
 	By default, the set of individuals eligible to participate in a decision is the set of group participants.
 	The Process Document does not require a quorum for decisions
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2277,6 +2285,10 @@ Start of a Review Period</h4>
 	Each Member organization <em class="rfc2119">may</em> send one review,
 	which <em class="rfc2119">must</em> be returned by its [=Advisory Committee representative=].
 
+	For clarity,
+	in the context of an [=AC Review=],
+	[=dissent=] <em class=rfc2119>must</em> be expressed as a [=Formal Objection=].
```
chaals, line 2290:
```suggestion
	[=dissent=] <em class=rfc2119>may</em> be expressed as a [=Formal Objection=].
```

There are options in the standard review forms we use to say "I disagree but ths is **not** a formal objection", which I think is appropriate.

dwsinger: Yes, it is, but that's not dissent in our formal definition, it's mere disagreement. Dissent is the presence of one or more FOs.

plehegar: I'm confused, maybe due to the difficult of seeing the preview. I read "A [=Formal Objection=] always indicates sustained disagreement,	but isn't necessary to express [=dissent=]." So, is dissent a sign of the presence of one or more FOs?

dwsinger: In some contexts, Dissent can only be expressed as an FO (AC Review). But in a WG, if the chair is trying to assess consensus, they can ask the something vaguely like "do we have any sustained objections to this proposed decision?" and if there is, they shouldn't (can't?) declare consensus. The member doesn't need (shouldn't) file an FO from the room in a hurry.

frivoal: > "A [=Formal Objection=] always indicates sustained disagreement, but isn't necessary to express [=dissent=]."

The presence of "sustained disagreement" and "dissent" are the same, so using different words here might have been confusing. I think we should rephrase as

> "A [=Formal Objection=] always indicates sustained disagreement, but isn't necessary to express it."


"sustained objection" = "I insist I cannot agree to this, and therefore, because of that disagreement, we do not have consensus."

"Formal objection" = "sustained objection + I want kick off the council process"

(Note: As per https://www.w3.org/2021/Process-20211102/#managing-dissent, even if there is not consensus, the chair can still record decisions (possibly by taking a vote), so we're not stuck.)

dwsinger: yes. "isn't necessary to express it in all circumstances"? or flip it and "and is the only way to express sustained disagreement in some (formal) contexts, such as AC votes"?

index.bs
```diff
@@ -2020,9 +2020,16 @@ Consensus</h4>
 
 		<dt><dfn id="def-Dissent">Dissent</dfn>:
 		<dd>
-			At least one individual in the set registers a <a href="#FormalObjection">Formal Objection</a>.
+			At least one individual in the set sustains an objection.
```
nigelmegitt, line 2023: I think we're missing a class here. I agree with comments made elsewhere that there can be dissent by disagreement that is not escalated to Objection level. 

To make this more concrete, consider that a WG is faced with two competing proposals to resolve the same issue. There are no objections to either, but there is also a split of preferences amongst members. The Chair may reasonably assert a Decision that is the choice with the least negative preference, i.e. dissent.

dwsinger: Hi. I think you may be missing that we have a difference of definition in the process between mere disagreement, and dissent (which is an FO or equivalent).  Maybe this suggests we should use "sustained dissent" instead of plain "dissent"?

nigelmegitt: Bit puzzled about this @dwsinger. Disagreement isn't a defined term in the Process. The only relevant reference I can find is in §5. Decisions where individuals are encouraged (SHOULD) to register an FO if they disagree _strongly_ with a decision (my emphasis). 

This recommendation actually seems like an anti-pattern, since the goal is to resolve strong disagreements _before_ they need to be escalated to an FO. Maybe we need a more nuanced section about handling disagreements as well?

dwsinger: disagreement doesn't need defining in the process, it has its usual meaning. Dissent is formally defined, however, to be an FO or its equivalent. So in the Process, not all disagreement is Dissent. As I say, we might make this clearer by using the phrase "sustained dissent".

nigelmegitt: I don't see any problem using "sustained dissent", but I fear there's a logical problem with saying that disagreement has its usual meaning, and then also saying that disagreement _excludes_ (sustained) dissent. If it's important that the two categories are mutually exclusive then we would have to define "disagreement" more explicitly.

fantasai: I'd rather not rename the concept of “dissent” in the Process. This pull request is about making its definition (and consequently the definition of “consensus”) more broadly applicable than just to REC-track documents. If we want to rename concepts, that should be a separate issue...

index.bs
fantasai: I agree with taking @dwsinger's suggested rephrasing.

PR #640, opened 2022-09-22
Rename Acknowledgement to Acceptance (for Member submission requests)

frivoal:
Acknowledgement sounds like "yes, we got your request", while it intends to mean "yes, we agree to publish your submission". Using the "accept" to mean that seems less confusing.


plehegar (approved):

swickr:
I'm skeptical about this proposed change in language.

The status label for published Submissions should be neutral regarding the content of the Submission.  This is why the boilerplate emphasizes "_Publication of a Note by W3C indicates no endorsement by W3C, the W3C Team, or any W3C Members. The acknowledgment of a Submission request does not imply that any action will be taken by W3C._"

To my ears "accepted" is moving closer to -- or could be inappropriately claimed as --  "we like/support/approve the content".  I'm confident there was long discussion of this two decades ago when [Process 2019 was explicit about the intended interpretation](https://www.w3.org/Consortium/Process/Process-19991111/submission.html#SubmissionYes).   That boilerplate as well as any specific prose about the content; e.g. whether it might be brought to the attention of an existing group, whether it is simply published for archival purposes, or whatever is appropriate for each case is appropriately provided in the SoTD.

Perhaps another term might work, but "accepted [for publication]" doesn't do it for me.

dwsinger:
@swickr ah, got it. Somehow we need to find a better middle ground between "acknowledgement" (usually used to mean 'yes, we received this', without the 'and we agree to publish it') and "acceptance" (which has the implications of support that you indicate and we do not want.) Maybe there is no better word.

nigelmegitt:
Recognition?

TallTed:
"Confirmation of submission"?

dwsinger:
"Confirmation of submission" and "recognition" both seem to suggest only "yes, we got it". Somehow we need to be able to say "yes, we got it, _and_ OK, we'll publish it" without saying or implying "and we endorse/support/agree-with the contents". I'm struggling. The original concern was that "acknowledge" only has the "yes, we got it" without the "yes, we'll publish it"; but "acceptance", as Ralph says, implies including "and we like it" which we explicitly do not wish to say or imply.

jeffjaffe:
I am sensitive to Ralph's point that "acceptance" is too strong.

I don't think we are fixing any big problem.

I suggest leaving things as they are.

koalie (commented):
-1 for the reasons Ralph developed in comment.

fantasai:
@swickr @koalie What about changing to “received” rather than “acknowledged”? I'm reviewing the diff and I think this works?

koalie:
Thanks, @fantasai this works for me!

nigelmegitt (commented):
This all works except that it's an odd situation that a submission could be rejected without having been received, from an English language perspective. It looks as though W3C is in the business of rejecting documents without reading them first. What I'd expect as a typical process is something more like:

1. Receive document
2. Review document
3. Accept or reject document
4. If accepting, use as the basis for a publication.

I don't agree that acceptance implies endorsement; further, the eventual publication that is likely to result from acceptance is in a form that is required to explicitly state the level of endorsement, i.e. a Rec or not-a-Rec.

TallTed (approved):

frivoal:
Sorry @nigelmegitt merged before seeing your push back. I am reverting now (confirmed with the chairs). We'll talk about it again later.

fantasai:
@nigelmegitt I think of it like if someone sent a package (in the old days) and you could either accept or reject it. If you rejected, did you really receive it? I think receiving has a bit of connotation about taking it in, it's not just that it landed on your doorstep. And in any case I think it's better than the current "acknowledge" verb we're using. Happy to debate alternatives, but I really don't think the current text is better.

frivoal:
Since the pull request was landed first, it is marked as merged, but the commit has now been reverted. I'll make a new pull request when/if we agree on what to do.

nigelmegitt:
> @nigelmegitt I think of it like if someone sent a package (in the old days) and you could either accept or reject it. If you rejected, did you really receive it? I think receiving has a bit of connotation about taking it in, it's not just that it landed on your doorstep. And in any case I think it's better than the current "acknowledge" verb we're using. Happy to debate alternatives, but I really don't think the current text is better.

`<flippant>`Good luck with your courier if you expect them to wait around while you open the package, inspect the contents, and then ask them to take it back again because you aren't happy with it!`</flippant>`

"Receive" is good for saying "we have it": the problem is lumping acceptance or rejection in with that. They are two separate states. It could be that a small tweak in the wording for how "accept" and "reject" are communicated would resolve this.

frivoal:
Other potential alternatives:
* verified / verification
* validated / validation
* confirmed / confirmation
* recognized / recognition
* admitted / admission
* permitted / permission

frivoal:
Some more:
* Authorize / Authorization
* Ratify / Ratification
* Sanction / Sanctioning
* Permit / Permission
* Grant / Granted
* Clear / Clearance
* Greenlight / Greenlighting
* OK / Okayed
* Affirm / Affirmation

TallTed:
> Since the pull request was landed first, it is marked as merged, but the commit has now been reverted. I'll make a new pull request when/if we agree on what to do.

It would be best to take the discussion to an issue, for better preservation and ease of future review, and thence to the new PR once action is agreed upon.

frivoal:
See https://github.com/w3c/w3process/issues/714 for continuing this discussion

PR #642, opened 2022-09-23
Adopt the W3C Council for handing Formal objections and related matters

frivoal:
This pull requests integrates all the changes prepared over the years for the director-free branches related to handling of formal objections by the W3C Council, and related matters.

The commits presented here do not attempt to reflect the full history of changes: these would be too many to comfortably review, often adding changes and the reverting or changing them further. Instead, these commits try to split the matter in largely thematic edits.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/pull/642.html" title="Last updated on Oct 13, 2022, 3:39 PM UTC (b9a0119)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/642/c83bf6b...b9a0119.html" title="Last updated on Oct 13, 2022, 3:39 PM UTC (b9a0119)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2338,6 +2338,11 @@ Extraordinary Delegation</h5>
 	(i.e., at least twice as many votes in favor as against).
 	Delegation in such cases cannot be later revoked.
 
+	The [=Team=] <em class=rfc2119>must</em> inform the [=Advisory Committee=]
+	when a [=Formal Objection=] has been delegated,
+	and to whom it has been delegated.
+
+
 <h5 id=council-participation>
 Council Participation, Dismissal, and Renunciation</h5>
 
--- a/index.bs
+++ b/index.bs
@@ -2295,7 +2295,7 @@ Investigation and Mediation by the Team</h4>
 
 	Otherwise,
 	upon concluding that consensus cannot be found,
-	and within at most 90 days of the [=Formal Objection=] being registered,
+	and no later than 90 days after the [=Formal Objection=] being registered,
 	the [=Team=] <em class=rfc2119>must</em> initiate formation of the [=W3C Council=],
 	which should be <a href="#council-convention">convened</a> within 45 days of being initiated.
 	Concurrently, it must prepare a report for the [=Council=]
--- a/index.bs
+++ b/index.bs
@@ -2414,14 +2414,20 @@ Council Chairing</h5>
 Council Deliberations</h5>
 
 	Having reviewed the information gathered by the [=Team=],
-	and after sufficient sufficient deliberation,
+	the Council <em class=rfc2119>may</em> conduct additional research or analysis,
+	or request additional information or interviews from anyone,
+	including the Team.
+
+	The Council <em class=rfc2119>may</em>
+	further attempt to broker consensus,
+	which, if successful, disposes the formal objection.
+
+	Otherwise,
+	after sufficient deliberation,
 	the [=W3C Council=] decides whether to
 	<dfn>sustain</dfn> or <dfn>overrule</dfn> the objection.
 	When [=sustaining=] an objection,
 	it <em class=rfc2199>should</em> recommend a way forward.
-
-	Individuals registering [=Formal Objections=] often support their reasoning
-	with several arguments.
 	The [=W3C Council=] <em class=rfc2119>may</em> [=overrule=] the [=Formal Objection=]
 	even if it agrees with some of the supportive arguments.
 
--- a/index.bs
+++ b/index.bs
@@ -2381,7 +2381,17 @@ Council Participation, Dismissal, and Renunciation</h5>
 	falling back to a vote if that fails.
 	The chair must be a member of the [=W3C Council=].
 
-	Having reviewed the information gathered by the [=Team=] and its recommendation,
+	In extraordinary cases,
+	if the [=Council=] determines that it would not be the appropriate body to handle a particular [=Formal Objection=],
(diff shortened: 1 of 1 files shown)


dwsinger:
@chrisn (But Chris isn't taggable here?)

the problem with "No one shall be both a party and an adjudicator" was the realization of defining exactly who a party _is_. The AC Rep that filed the FO maybe, but they formally file on behalf of anyone in their organization; they might not have a strong personal opinion (or even agree). On the other side, identifying the key supporting parties is much much harder. The chair of the WG? They might be 'just' a facilitator? Editor? The same. Major proponents of the approach or detail in question? How would we identify them?

Hence we tried to leave it to the dismissal. I think adding "Anyone who is clearly a party to the dispute" to the "possible reasons to think about" in the Guide, if not already there, does make sense.

index.bs
```diff
@@ -1971,8 +1977,8 @@ Types of Decisions</h3>
 	(also known as group “resolutions”).
 
 	A <dfn id="def-w3c-decision">W3C decision</dfn> is one
-	where the [=Director=] decides,
-	after exercising the role of assessing consensus of the W3C Community after an [=Advisory Committee review=].
+	where the [=Team=]
+	has exercising the role of assessing consensus of the W3C Community after an [=Advisory Committee review=].
```
wseltzer, line 1981:
```suggestion
	exercises the role of assessing consensus of the W3C Community after an [=Advisory Committee review=].
```
s/has exercising/exercises/

index.bs
```diff
@@ -2211,17 +2217,16 @@ Reopening a Decision When Presented With New Information</h3>
...
+	they <em class="rfc2119">may</em> ask that the decision be confirmed or denied
+	(for representatives of a Member organization, via their Advisory Committee representative).
 	This is a <dfn export id="wg-decision-appeal">Group Decision Appeal</dfn>
 	or a <dfn export id="chair-decision-appeal">Chair Decision Appeal</dfn>.
 	The participants <em class="rfc2119">should</em> also make their requests known
 	to the <a href="#TeamContact">Team Contact</a>.
-	The Team Contact <em class="rfc2119">must</em> inform the Director
-	when a group participant has raised concerns about due process.
+	The Team Contact <em class="rfc2119">must</em> inform the [=CEO=]
+	when a group participant has also raised concerns about due process.
```
wseltzer, line 2227: why "also"?

index.bs
```diff
@@ -2255,6 +2264,253 @@ Recording and Reporting Formal Objections</h3>
...
+
+<h3 id=addressing-fo>
+Addressing Formal Objections</h3>
+
+<h4 id=team-fo-mediation>
+Investigation and Mediation by the Team</h4>
+
+	The [=Team=] considers the [=Formal objection=],
+	researches the question,
+	interviewes parties,
```
wseltzer, line 2278:
```suggestion
	interviews parties,
```

index.bs
chaals:
```suggestion
	assesses consensus of the W3C Community after an [=Advisory Committee review=].
```

index.bs
```diff
@@ -2255,6 +2264,253 @@ Recording and Reporting Formal Objections</h3>
...
+
+	The <dfn export local-lt="Council">W3C Council</dfn> is the body convened to resolve [=Formal Objections=]
+	by combining the capabilities and perspectives of the [=AB=], the [=TAG=], and the [=Team=],
+	and is tasked with doing so in the best interests of the Web and W3C.
+
+<h5 id=council-composition>
+Council Composition</h5>
+
+	The [=W3C Council=] consists of the following members,
+	with the exception of any who are [=dismissed=] or who formally [=renounce=] their position:
```
wseltzer, line 2306:
```suggestion
	with the exception of any who are [=dismissed=] or who formally [=renounce=] their positions:
```

index.bs
```diff
@@ -2255,6 +2264,253 @@ Recording and Reporting Formal Objections</h3>
...
+	In that case,
+	the decision is made by simple majority,
+	with the [=W3C Council Chair=] breaking any tie.
+	In case of a vote,
+	if two members of the Council who share the same affiliation cast an identical ballot,
+	then their ballots count as a one vote,
+	not two.
+
+	In the case of non-unanimous decisions,
+	members of the [=W3C Council=] who disagree with the decision
```
wseltzer, line 2458:
```suggestion
	any member of the [=W3C Council=] who disagrees with the decision
```

wseltzer (commented):
My batch of comments above is purely editorial fixes.

index.bs
fantasai: I think either phrasing would be correct, but the current one evokes more the idea that multiple people could write a single Minority Opinion whereas your proposed phrasing evokes more the idea that each person could write their own Minority Opinion. (Both phrasing allow either, but the emphasis is different.) I would lean towards the collective Minority Opinion since I think that would be a more natural way for the Council to lean.

frivoal: Even if it is subtle, I think it's worth pushing even a little bit for people to try and agree with each other. A single Minority Opinion can cite multiple arguments in any case.

index.bs
fantasai: Because the participant might have raised an FO, in which case the CEO doesn't need to be informed, but if they raised concerns about due process specifically, the CEO must be informed. (The “also” is clarifying that this is a distinct additional requirement for this subcase.)

yahyafurkann (commented):
Cod.v

PR #659, opened 2022-10-14
Clarify expectations about FO mitigations

frivoal:
Here are some suggested improvements to the handling of mitigations of overturned decisions, based on discussions of #645 during the process CG call of October 13th.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/659.html" title="Last updated on Oct 25, 2022, 2:29 PM UTC (04f2fce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/659/8ed8a93...frivoal:04f2fce.html" title="Last updated on Oct 25, 2022, 2:29 PM UTC (04f2fce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2445,11 +2445,8 @@ Council Deliberations</h5>
 
 	Note: This does not create new powers for the [=Team=],
 	such as the ability to “unpublish” documents.
-	If something needs to be changed in an already published document,
-	the group where this document originated
-	would be typically responsible for drafting the necessary changes and getting them published,
-	following the usual rules of the Process.
-	The [=Team=]'s role would be to hold them to it.
+	The [=Team=]'s role is to ensure the responsible parties enact adequate mitigations,
+	by whatever means they already have at their disposal.
 
 	The Council <em class=rfc2119>may</em> form sub-groups for deliberation,
 	who may return with a recommendation,
--- a/index.bs
+++ b/index.bs
@@ -2441,7 +2441,15 @@ Council Deliberations</h5>
 	(e.g., if the objection concerns material already in a published document)
 	the Council <em class=rfc2119>should</em> suggest how these consequences might be mitigated.
 	The [=Team=] is responsible for making sure that adequate mitigations are enacted in a timely fashion;
-	and the [=Formal Objection=] is not considered fully addressed until then.
+	and the [=Formal Objection=] is not considered <dfn>fully addressed</dfn> until then.
+
+	Note: This does not create new powers for the [=Team=],
+	such as the ability to “unpublish” documents.
+	If something needs to be changed in an already published document,
+	the group where this document originated
+	would be typically responsible for drafting the necessary changes and getting them published,
+	following the usual rules of the Process.
+	The [=Team=]'s role would be to hold them to it.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3437,6 +3446,7 @@ Updating Mature Publications on the Recommendation Track</h4>
 			[=Team=] verification (a [=Team decision=]),
 			<em>should</em> be withheld if any Process requirements are not met,
 			and <em class=rfc2119>may</em> be withheld in consideration of unresolved Formal Objections
+			(including any [=sustained=] by the [=Council=] but not yet [=fully addressed=])
```
nigelmegitt, line 3449: Why is this a MAY for mature publications but a MUST for general transition requirements?

frivoal: It is "may" for updating a publication while staying at the same status, like CR->CR, because you may be in the process of dealing with issues. If you've got a pile of 30 issues, and you've fixed 15 of them, you have 15 left and they're still problems, but you've made progress, so republishing is an improvement. So, if you're ignoring FOs, you should be denied. If you're working through a pile of things, and you're not done with all of them yet, then it's fine.

For a transition to a more mature stage, you're actually supposed to have cleared the deck, so it's a "must" in that case.

nigelmegitt: Right, that makes sense (with one exception), thanks @frivoal - unfortunately the heading "6.3.4. Updating Mature Publications on the Recommendation Track" is suggestive of specifications at the higher end of maturity, rather than all stages. It could be a separate ticket, but renaming the heading would avoid this. E.g. "Updating Recommendation Track Publications without advancement".

The exception is Rec -> Rec transition, where I'd expect something like the streamlined requirements (which talk about issues, of which I think FOs are a subset) to apply even for team verification.

frivoal: Ah, right, I see. This is tricky, because the criteria for this doesn't apply to WD->WD transition. "Updating Recommendation Track Publications without advancement" as you suggested implies it would apply to all updates, not just certain ones as clarified in the opening paragraph. I suspect renaming could be useful for the reason you highlighted, but I don't think we've found the right name just yet.

nigelmegitt: Oh, I see that I was even _more_ confused than I first realised when looking at this text. The text in §6.3.4 says "Certain requests" and does not define which ones, and then defines the term "Update requests", which is then referenced from elsewhere, which is where these are brought in normatively.

Maybe this subsection should be called "Requirements for Update Requests" and there could be a note that says that the option to make Update Requests is defined by reference from the relevant sections. Right now update request requirements need to be met for:
* CRS -> CRS
* Rec -> Rec

And they are called out as _not_ being needed for CRD -> CRD and Registry Table updates.

Given that, we could state, even in a note, within §6.3.4, that those "certain categories" are exactly the two I listed above. I think it would help avoid confusion for other readers.

Could it be helpful to define a term for the special nature of CRS and Rec as being normative documents for which we expect/permit there to be normative references from other documents?

(sidenote: I still don't think normative references to CR are a good idea, but the decision to permit them is now ancient history)

fantasai: @nigelmegitt Can you file your comments here as a separate issue? :) I think your suggestions to clarify this make sense, but they're about clarifying pre-existing text.

PR #667, opened 2022-10-21
fix punc, spelling, inconsistent labels, etc.

### LINKED ISSUE #661: editorial: textual inconsistencies in basic-rec-track.svg
See https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/654.html#rec-track
which will update https://github.com/w3c/w3process/blob/main/basic-rec-track.svg

I'm logging this issue so I'll remember to submit a PR on the SVG after #654 gets merged.

I [first noticed](https://github.com/w3c/w3process/pull/654#issuecomment-1264500199) three incorrectly punctuated instances of `e.g.` in the image:

* `WG or Team decision e.g. for further review`
* `AC Review, Team Decision e.g. for editorial changes`
* `Advisory Committee review and Team's Decision, e.g. for further work and review`

Having given these further thought, I don't think the `e.g.` are helpful here, even with proper punctuation (i.e., preceded and followed by commas, according to the relevant style guides; alternatively, they could be made parenthetical and only followed by commas, as in this parenthetical).

Therefore, I would suggest changing those three text snippets as follow, respectively:

* `WG or Team Decision, such as "returned for further review"`
* `AC Review or Team Decision, such as "returned for editorial changes"`
* `AC Review and Team Decision, such as "returned for further work and review"`

I also noticed several other inconsistencies in the text within this image.  These include —
* capitalization of `decision` vs `Decision` (I used `Decision` above)
* possessive vs descriptive, `Team` vs `Team's` (I used `Team` above)
* `AC` vs `Advisory Committee` (I used `AC` above)
* `+` vs `and` (I suggest `and`)

There may be more.

I do not have strong feelings on which direction most of these go, only that the image be made internally consistent, which consistency I'll propose via a PR after #654 is merged.  If anyone has strong feelings about which direction the above issues are pushed, please speak up.

TallTed:
fixes #661

### DIFF
--- a/basic-rec-track.svg
+++ b/basic-rec-track.svg
@@ -40,7 +40,7 @@
 								<polygon points="120,114 116,124 114,113"></polygon>
 							</a>
 						</g>
-					<!-- Advance to CR -->
+					<!-- Advance to CRS -->
 						<g role="listitem" fill="#060">
 							<a href="#cr-1">
 								<g role="img">
@@ -53,7 +53,7 @@
 						</g>
 					</g>
 				</g>
-			<!-- CR -->
+			<!-- CRS -->
 				<g role="listitem">
 					<a href="#pr-1">
 						<ellipse role="img" id="cr-1" ry="18" rx="38" cy="140" cx="260" stroke="black" fill="#fff">
@@ -68,7 +68,7 @@
 						<text aria-hidden="true" font-size="14" y="84" x="260" text-anchor="middle">CRD</text>
 					</a>
 					<g role="list">
-					<!-- CR to CRD -->
+					<!-- CRS to CRD -->
 						<g role="listitem">
 							<a href="#crd-1">
 								<g role="img">
@@ -94,11 +94,11 @@
 								<polygon points="282,54 277,64 275,53"></polygon>
 							</a>
 						</g>
-					<!-- CRD to CR -->
+					<!-- CRD to CRS -->
 						<g role="listitem">
 							<a href="#cr-1">
 								<g role="img">
-									<title>Publish revised Candidate Recommendation</title>
+									<title>Publish revised Candidate Recommendation Snapshot</title>
 								</g>
 								<text font-size="8">
 									<tspan x="290" y="102">WG Decision and</tspan>
@@ -108,15 +108,15 @@
 								<polygon points="286,128 282,118 290,118"></polygon>
 							</a>
 						</g>
-					<!-- Update CR in place -->
+					<!-- Update CRS in place -->
 						<g role="listitem">
 							<a href="#cr-1">
 								<g role="img">
-									<title>Publish revised Candidate Recommendation</title>
+									<title>Publish revised Candidate Recommendation Snapshot</title>
 								</g>
 								<text font-size="8">
 									<tspan x="182" y="112">WG Decision</tspan>
-									<tspan x="288" y="112">Team Approval</tspan>
+									<tspan x="290" y="112">Team Approval</tspan>
 								</text>
 								<path stroke="#000" d="M242,124C238,114 244,104 260,104 271,104 277,108 279,114" stroke-dasharray="5 11" fill="none"></path>
 								<path stroke="#060" d="M242,124C238,114 244,104 260,104 271,104 277,108 279,114" stroke-dasharray="5 11" stroke-dashoffset="8" fill="none"></path>
@@ -173,11 +173,11 @@
 								<polygon points="534,136 544,140 534,144"></polygon>
 							</a>
 						</g>
-					<!-- Return to CR -->
+					<!-- Return to CRS -->
 						<g role="listitem">
 							<a href="#cr-1">
 								<g role="img">
-									<title>Return to Candidate Recommendation</title>
+									<title>Return to Candidate Recommendation Snapshot</title>
 								</g>
 								<text font-size="8">
 									<tspan x="302" y="160">AC Review and/or </tspan>
--- a/basic-rec-track.svg
+++ b/basic-rec-track.svg
@@ -10,8 +10,8 @@
 						</rect>
 						<text font-size="8">
 							<tspan aria-hidden="true" y="132" x="0">First Public WD</tspan>
-							<tspan x="0" y="152">WG decision</tspan>
-							<tspan x="0" y="162">Team's approval</tspan>
+							<tspan x="0" y="152">WG Decision</tspan>
+							<tspan x="0" y="162">Team Approval</tspan>
 						</text>
 						<path d="M0,140h53" fill="none" stroke="#000"></path>
 						<polygon points="47,136 57,140 47,144"></polygon>
@@ -33,8 +33,8 @@
 									<title>Publish a new Working Draft</title>
 								</g>
 								<text font-size="8">
-									<tspan x="30" y="92">WG Decision: review needed, or</tspan>
-									<tspan x="40" y="100">No change for 6 months</tspan>
+									<tspan x="30" y="90">WG Decision, such as "review needed"</tspan>
+									<tspan x="30" y="100">or "no change for 6 months"</tspan>
 								</text>
 								<path d="M78,124C73,114 79,104 97,104 108,104 115,108 117,114" fill="none" stroke="black" stroke-dasharray="6 1"></path>
 								<polygon points="120,114 116,124 114,113"></polygon>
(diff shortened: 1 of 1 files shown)


basic-rec-track.svg
```diff
@@ -116,7 +116,7 @@
 								</g>
 								<text font-size="8">
 									<tspan x="182" y="112">WG Decision</tspan>
```
frivoal, line 118: This label overlaps exactly (and is redundant with) the one from the `<!-- CR to CRD -->` block.

basic-rec-track.svg
```diff
@@ -116,7 +116,7 @@
...
-									<tspan x="288" y="112">Team’s approval</tspan>
+									<tspan x="288" y="112">Team Approval</tspan>
```
frivoal, line 119: This label overlaps almost exactly, but is shifted by two px, with the one from the `<!-- CRD to CR -->` block.

frivoal (changes requested):
There are some overlapping/redundant labels. They were not created by this PR, but after this PR, they don't align exactly, and that makes things look weird and unreadable.

basic-rec-track.svg
TallTed: Would you say one should be deleted (and if so, which), or the redundant overlay should be retained?

basic-rec-track.svg
TallTed: Would you say one should be deleted (and if so, which), or the redundant overlay should be retained?

TallTed:
@frivoal — I have no problem changing the partial overlap to be complete, with the new text taking precedence on the latter instance. I would also have no problem deleting one or the other if that makes sense.

Oddly, GitHub is not offering me a "commit suggestion" button, so I'll have to manually apply the changes. Doing so now to make the obscuring overlap clear. Will delete whichever instance (if either) is determined to be unnecessary whenever that happens.

TallTed:
Digging deeper... @fantasai @frivoal —

The rendered SVG (now) shows no `CR to CRD`, so it seems the comments in the source may need some updating. I think —
* all instances of `Candidate Recommendation` without suffix should become `Candidate Recommendation Snapshot`
* `<!-- Advance to CR -->` (line 43) should become `<!-- Advance to CRS -->`
* `<!-- CR -->` (line 56) should become ```
* `<!-- CR to CRD -->` (line 71) should become `<!-- CRS to CRD -->`
* `<!-- CRD to CR -->` (line 97) should become `<!-- CRD to CRS -->`
* `<!-- Update CR in place -->` (line 111) should become `<!-- Update CRS in place -->`
* `<!-- Return to CR -->` (line 176) should become `<!-- Return to CRS -->`

Related to this, `<!-- Update CR in place -->` (line 111, to become `<!-- Update CRS in place -->`) does not appear to have any labels in the rendered SVG, though there are two in the source (lines 118–119) which are the ones that overlay labels on other sections (lines 105 & 119, labeling `<!-- CRD to CRS -->`; and 78 & 118, labeling `<!-- CRS to CRD -->`).

If you agree with the above, please +1 or similar, and I'll put them all into my PR. If you've partial disagreement, please note that as well, and I'll update what's agreed, and we can continue to discuss what's not agreed.

fantasai:
This diagram has been reviewed for understandability by screenreaders, so whatever you do, _don't break that_.

frivoal:
>This diagram has been reviewed for understandability by screenreaders, so whatever you do, don't break that.

Maybe the easiest thing to do then is to move back the label by 2px so that it lines up again.

TallTed:
>> This diagram has been reviewed for understandability by screenreaders, so whatever you do, don't break that.
>
> Maybe the easiest thing to do then is to move back the label by 2px so that it lines up again.

Yes, I think the 2px nudge is the right answer for the overlaying labels, because these labels actually do apply to two pathways, so for screenreaders, they should be present twice.

I am also changing all the ` CR ` in comments to ` CRS `, so the comments match the rendered image content (and the text being illustrated here).

basic-rec-track.svg
TallTed: Addressed in https://github.com/w3c/w3process/pull/667/commits/d1bd24fa7e625a3b41799d183094133b2e9e607a

TallTed:
@fantasai — You're not in the GitHub reviewer list, so I can't ask for that as such ... but here's a free-text request for your review. :-)

frivoal (approved):

fantasai (changes requested):
Mmmm, I'm not sure you wanted my review. :] Here's my full review:

- If you're intending to obsess over punctuation, I'm going to require you to use curly quotes properly. :)
- I would remove the “such as” clause for republishing WD, it's wordy and doesn't seem helpful.
- I would change “returned for further review” to “returned for further work” on X -> WD
- It's not clear whether the “such as” clause for PR->CRS is attached to both or to Team Decision only, and I'm not sure it's helping, so I would remove it also.
- Team Decision after AC Review should be “W3C Decision”.
- CRS->PR needs a WG Decision in addition to Team Approval.
- Need to be consistent in labeling the WG Decision + Team Approval bits. Some of them have an “and”, some of them don't. I suggest using a plus sign throughout, like in the old version. (This is a diagram, not free text.)

(I also think your extended “such as” clauses are too wordy for a diagram, but I can live with it.)

TallTed:
@fantasai — Keeping in mind that I did not create the diagram, only cleaned up a number of existing inconsistencies and (to my eye) errors, I think several of your bullets are not really appropriate to a review of my cleanup PR.

_**Note:** You can see exactly what I changed (vs. what I was changing and/or left alone) by viewing [the files changed by the PR](https://github.com/w3c/w3process/pull/667/files) and clicking the `<>` button ("Display the source diff") at the upper right of the display pane. At the moment, [this is a direct link](https://github.com/w3c/w3process/pull/667/files#diff-63577492a1562f7d4e72fda469b8a08f99ec4a1d904bca3e412595d223e43db4) — but that link will change as the PR changes. It is probably better to make future reviews and/or suggestions by looking at this page, where you can be exact about where your suggestion applies — and know whether you're addressing *my* PR or the previously existing content._

In more detail —

> * If you're intending to obsess over punctuation, I'm going to require you to use curly quotes properly. :)

My tools don't insert curly quotes unless I force them to. I generally avoid "smart" curly quotes because many automated tools choke on them in varying ways.  They also tend to be used inconsistently and varyingly by international authors and editors, while straight quotes are *usually* consistently used.

_If curly quotes are to be used, I strongly advise using the HTML or other relevant escapes for those characters, because the characters themselves are *often* misread and mistreated by subsequent editors — variously leading to pairs of left-quotes, pairs of right-quotes, reversed left- and right-quotes, mixes of curly and straight quotes, and other offenses against correct punctuation._

Also note that the [_W3C Manual of Style_ says to](https://www.w3.org/Guide/manual-of-style/#Punctuation) _Use quotation marks rather than grave accents and apostrophes to quote text (e.g., ``value'' should read "value")._  Note that this example itself uses straight quotes, suggesting (though not stating) that straight quotes are W3C style.

> * I would remove the “such as” clause for republishing WD, it's wordy and doesn't seem helpful.

Unless the thing(s) following any "such as" instance are a full list of things that might be put there, something like `such as` (which might include `e.g.` if punctuated correctly) is appropriate.  I know this is "just a diagram" but I also know that people reading documents like this often treat the diagrams as equivalent to the nearby text (even when the document text — and/or the diagram itself — explicitly says they are not equivalent).

> * I would change “returned for further review” to “returned for further work” on X -> WD

Please note that "further review" was the existing text.  I am happy to make this change, but perhaps it should be in a different PR?

> * It's not clear whether the “such as” clause for PR->CRS is attached to both or to Team Decision only, and I'm not sure it's helping, so I would remove it also.

I changed an "e.g." into "such as", so again, I'm not sure dropping that clause should be done via this PR.

Upon re-reading the whole diagram, I wonder whether the "AC Review and/or W3C Decision" on this vector (which had used `,` previously) should be `and` rather than `and/or`?

> * Team Decision after AC Review should be “W3C Decision”.

To be clear, this is on the PR->REC vector?  (I wonder if that should change to "AC Review and W3C Decision", as it seems that both are required?)

> * CRS->PR needs a WG Decision in addition to Team Approval.

Again, not my text.  Happy to apply this, if confirmed appropriate for this PR.

> * Need to be consistent in labeling the WG Decision + Team Approval bits. Some of them have an “and”, some of them don't. I suggest using a plus sign throughout, like in the old version. (This is a diagram, not free text.)

This is indeed a diagram. Still, it will pass through text readers and other accessibility tools, and they will treat `+` differently than `and` or `&` or other alternatives.  Note that I changed all `+` to `and`, but I did not delete any `+` or `and` entirely, so any that are missing now were missing before.  Again, I'm happy to add them, upon confirmation that they should be added through this PR.

fantasai:
@TallTed, I pushed a set of changes to this branch that I think address all the relevant concerns with this diagram. I think this now correctly represents the decision pathways.

I did remove the “such as” clauses, because they're a bit much for the diagram given how many other things are going on here. (And also they weren't particularly accurate anyway.)

Let me know what you think.

fantasai (approved):

frivoal (approved):

Issue #652, opened 2022-09-28
Chair appointments


mnot:
The `director-free` branch currently has the Team appointing WG chairs, but there isn't a clear mechanism for disputes; the FO process does not apply to these decisions, as I read it.

Should there be a process for disputing a Team decision like this?

First raised in #316.

frivoal:
I believe there is a path already:
* The Process does not limit the types of decisions that can be formally objected to, so I see no reason why you couldn't Formally Object to chairing appointments / changes / reappointment / removals
* Initial appointment of chairs is done via charter creation / adoption, and these are subject to AC reviews, which clearly people can formally object to.
* Says https://www.w3.org/2021/Process-20211102/#cfp “Advisory Committee representatives may initiate an [Advisory Committee Appeal](https://www.w3.org/Consortium/Process/Drafts/#ACAppeal) against the decision to create or substantially modify a Working Group”. Changing a chair is a substantial modification of a working group.

All in all, these are fairly heavy handed approaches, and I suspect that for personnel issues, expressing concerns in private to the Team is most effective, but if that fails, there are appeal paths.

We could discuss whether these mechanisms are ideal, but by my read, we do have the ability to dispute such decisions, so I don't see that as a pressing issue.

dwsinger:
I would not _like_ to see FO used, but as I read it, anyone can FO any decision, and an appointment is surely a decision. Why do you think FO is not available? (I agree with Florian, I hope it can be resolved without such a heavy tool.)

frivoal:
@mnot Are the responses from @dwsinger and I above sufficient to address you concern, or do you think we need to do something else?

mnot:
I read [section 5.1](https://www.w3.org/Consortium/Process/Drafts/director-free/#decision-types) as constraining the types of decisions that can use the FO mechanism defined in [section 5.5]() to 'chair decisions', 'group decisions' and 'W3C decisions', the latter being defined as:

> determined by the [Team](https://www.w3.org/Consortium/Process/Drafts/director-free/#team) on behalf of the W3C community by assessing the consensus of the W3C Community after an [Advisory Committee review](https://www.w3.org/Consortium/Process/Drafts/director-free/#advisory-committee-review).

Since there is no AC review of the Team's [appointment of a chair](https://www.w3.org/Consortium/Process/Drafts/director-free/#ReqsAllGroups), this action doesn't seem to meet the 'after an AC review' qualification.

[...]

PR #674, opened 2022-11-22
Clarify what can be formally objected to

frivoal:
Earlier refactorings may have given the impression that Team Decisions were not decisions that could be appealed.

See #652


mnot:
Lgtm!

Issue #678, opened 2022-11-22
Possibly dismissing a Formal Objection before it goes to full Council?


dwsinger:
If we continue to have a lot of FOs, it may help to have a mechanism to dismiss them before Council, similar to the [Dismissal before trial](https://www.rothmanlawyer.com/dismissing-a-lawsuit-before-trial/) procedures in (at least) anglo-american jurisprudence.

It is by no means clear what could be grounds for "early dismissal" however, or who might be authorized to make such a decision, so this is probably not easy to pursue.

chaals:
In principle this sounds good. It is not yet clear how well the Council establishes the trust of the community. "Someone convinced the objector to withdraw", is an existing mechanism which I hope would not be through discussion tinged with "we promise to make your life miserable if you don't withdraw". It's also possible that the Council dismisses some objections very fast, following the actual process laid out for it.

I don't think that working out how else to dismiss objections before they are considered by the council is a higher priority than making sure the council does a good job of resolving the objections that are brought to it, so I suspect this issue won't lead to an acceptable change proposal for a while, if at all.

dwsinger:
I have a personal hope that the number of objections goes down, and the work of forming a Council for a decision that could have fallen into this case is rare and not too much trouble.  And that otherwise, working with the objector to find a resolution so that they withdraw the objection may happen.

I filed the issue mostly to put it on the radar as it's long term and could be easily forgotten.

nigelmegitt:
I support the idea that we should not do the costly process of setting up an FO Council for essentially frivolous formal objections. Defining and assessing frivolity is likely to be difficult though. I have some ideas for possible reasons to dismiss early - probably incomplete list:

1. Objector says they cannot accept a Decision but proposes a resolution that is not substantively different. - in this case, doing nothing should also satisfy the objector.

2. The form of the FO is an instance of [Unacceptable Behaviours](https://www.w3.org/Consortium/cepc/#dfn-unacceptable-behaviors) as defined by the CEPC. (it could also happen to be that an instance of the "proposes a resolution that is not substantively different" is actually in the sub-category of "sustained disruption of discussion" that list)

3. "FO is not well formed", e.g. there's no evidence that efforts were made to resolve by consensus, or evidence that the objector did not participate in such efforts. Or that there's apparently no resolution that would satisfy the objector. "I object on the grounds that the Supreme Being told me in a dream that unstated Terrible Things will occur as a result of this Decision, and it's already too late to save the world."

[...]

PR #703, opened 2023-01-25
Introduce path to shortcut the Council

frivoal:
To be used in cases that can be resolved unanimously without debate solely based on the Team's report and proposed resolution.

See #678

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/703.html" title="Last updated on Feb 8, 2023, 3:33 PM UTC (a96be22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/703/c2ab25d...frivoal:a96be22.html" title="Last updated on Feb 8, 2023, 3:33 PM UTC (a96be22)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2426,6 +2426,23 @@ Council Participation, Dismissal, and Renunciation</h5>
 	The [=W3C Council=] <em class=rfc2110>may</em> still solicit and hear their testimony,
 	as they can of anyone else in the W3C community.
 
+<h5 id=council-short-circuit>
+Unanimous Short Circuit</h5>
+
+
+	The full Council process <em class=rfc2119>may</em> be short-circuited if
+	the Team recommends a resolution
+	and every potential member of the Council who is not renouncing their seat
+	votes affirmatively (no abstentions) to adopt this resolution.
+
+	This step <em class=rfc2119>may</em> be run concurrently with [[#council-participation]]
+	and prior to choosing a [=W3C Council Chair|Chair=].
+
+	Note: This is intended for exceptional cases
+	that seem to have an obvious resolution
+	and which are too trivial or repetitive
+	to warrant a full Council response.
+
 <h5 id=council-chairing>
 Council Chairing</h5>
 
--- a/index.bs
+++ b/index.bs
@@ -2440,7 +2440,7 @@ Unanimous Short Circuit</h5>
 
 	Note: This is intended for exceptional cases
 	that don't seem to warrant a full Council response
-	because they are, for instance, too trivial, repetitive, absurd, etc.
+	because they are, for instance, too trivial, duplicative, etc.
 
 <h5 id=council-chairing>
 Council Chairing</h5>
--- a/index.bs
+++ b/index.bs
@@ -2439,9 +2439,8 @@ Unanimous Short Circuit</h5>
 	and prior to choosing a [=W3C Council Chair|Chair=].
 
 	Note: This is intended for exceptional cases
-	that seem to have an obvious resolution
-	and which are too trivial or repetitive
-	to warrant a full Council response.
+	that don't seem to warrant a full Council response
+	because they are, for instance, too trivial, repetitive, absurd, etc.
 
 <h5 id=council-chairing>
 Council Chairing</h5>


index.bs
```diff
@@ -2426,6 +2426,23 @@ Council Participation, Dismissal, and Renunciation</h5>
...
+	the Team recommends a resolution
+	and every potential member of the Council who is not renouncing their seat
+	votes affirmatively (no abstentions) to adopt this resolution.
+
+	This step <em class=rfc2119>may</em> be run concurrently with [[#council-participation]]
+	and prior to choosing a [=W3C Council Chair|Chair=].
+
+	Note: This is intended for exceptional cases
+	that seem to have an obvious resolution
+	and which are too trivial or repetitive
```
dwsinger, line 2443: is this 'and' or 'or'?

dwsinger (approved):

cwilso:
LGTM

index.bs
TallTed: Which `and` or `or` are you asking about?
```
	that seem to have an obvious resolution
	and/or which are too trivial or repetitive
```
— or —
```
	that seem to have an obvious resolution
	and which are too trivial and/or repetitive
```
— or —
```
	that seem to have an obvious resolution
	and/or which are too trivial and/or repetitive
```
— or —
```
	that seem to have an obvious resolution, are
	too trivial, and/or are too repetitive
```
— or, since "obvious resolution" seems redundant with "trivial", my favorite —
```
	that are
	too trivial and/or repetitive
```

dwsinger: The 'and' at the beginning, as you surmised; the condition seems to be `(has obvious resolution) AND (is trivial or repetitive)`, whereas I suspect there may be matters with obvious resolutions that are not trivial or repetitive (e.g. something may be momentous and complex, but out of our scope, and hence has the obvious resolution that we shouldn't touch something out of our scope). Not major, but 'or' or 'and/or' would seem safer to me.

fantasai: On the one hand, I agree that "trivial or repetitive" is an inexhaustive list of why we might want to take this path.

On the other hand, I think "obvious resolution" by itself isn't a good enough reason to take the fast path, because there might be a resolution that's obvious from the Team report, but if the Council were called and asked to deliberate, we might find some merit in the case that was not clear.

I don't want the Council to start forming a habit of dismissing FOs because they're a bit too checked out to dedicate time to digging into the issue and would rather ratify whatever the Team came up with.

I'm not sure yet how to word this better... maybe

```
Note: This is intended for exceptional cases
that don't seem to warrant a full Council response
because they are e.g. too trivial or repetitive or absurd.
```

?

TallTed:
```
Note: This is intended for exceptional cases
that don't seem to warrant a full Council response
because they are, for instance, too trivial, repetitive, absurd, etc.
```
?

frivoal: took @TallTed 's suggested variant on @fantasai's proposal.

cwilso (approved):
I'm a little worried about the word "absurd" here, but I'll let it pass.

chrisn:
I would suggest changing "absurd" ... Also, what is "repetitive" intended to mean here - that an FO can be short-circuited on the basis of precedent from a previous FO?

dwsinger:
> I would suggest changing "absurd" ... Also, what is "repetitive" intended to mean here - that an FO can be short-circuited on the basis of precedent from a previous FO?

I don't think similar is enough; essentially identical would be repetitive.

chrisn:
> I don't think similar is enough; essentially identical would be repetitive.

I agree. So if "repetitive" means "based on precedent from previous FO conclusions", I'd suggest we say that, for clarity.

But, I have a general concern with this change that we're designing process for situations that haven't happened yet, so would prefer to see us gain more experience before adding this.

dwsinger:
> > I don't think similar is enough; essentially identical would be repetitive.
> 
> I agree. So if "repetitive" means "based on precedent from previous FO conclusions", I'd suggest we say that, for clarity.
> 
> But, I have a general concern with this change that we're designing process for situations that haven't happened yet, so would prefer to see us gain more experience before adding this.

perhaps duplicative would be a better word?

index.bs
```diff
@@ -2426,6 +2426,22 @@ Council Participation, Dismissal, and Renunciation</h5>
...
+	the Team recommends a resolution
+	and every potential member of the Council who is not renouncing their seat
+	votes affirmatively (no abstentions) to adopt this resolution.
+
+	This step <em class=rfc2119>may</em> be run concurrently with [[#council-participation]]
+	and prior to choosing a [=W3C Council Chair|Chair=].
+
+	Note: This is intended for exceptional cases
+	that don't seem to warrant a full Council response
+	because they are, for instance, too trivial, repetitive, absurd, etc.
```
TallTed, line 2443:
```suggestion
	because they are, for instance, too trivial, duplicative, etc.
```

Issue #679, opened 2022-11-28
[editorial] Council composition statement spends too much time on exceptions


fantasai:
In https://www.w3.org/Consortium/Process/Drafts/#council-composition most of the words in the introductory sentence are dedicated to outlining the exceptions, and this is a bit awkward. Filing this issue as a reminder to try to improve the phrasing here.

PR #706, opened 2023-01-25
Simplify phrasing about council composition

frivoal:
See #679

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2326,8 +2326,7 @@ The W3C Council</h4>
 <h5 id=council-composition>
 Council Composition</h5>
 
-	The [=W3C Council=] consists of the following members,
-	with the exception of any who are [=dismissed=] or who formally [=renounce=] their positions:
+	The [=W3C Council=] is composed from:
 	* the [=CEO=] or their chosen [=Team=] delegate
 	* the members of the [=Advisory Board=]
 	* the members of the [=Technical Architecture Group=]


dwsinger (approved):

cwilso (approved):

index.bs
```diff
@@ -2326,8 +2326,7 @@ The W3C Council</h4>
 <h5 id=council-composition>
 Council Composition</h5>
 
-	The [=W3C Council=] consists of the following members,
-	with the exception of any who are [=dismissed=] or who formally [=renounce=] their positions:
+	The [=W3C Council=] is composed from:
```
nigelmegitt, line 2329: Each W3C Council is potentially different.

```suggestion
	The members of a [=W3C Council=] are selected from:
```

TallTed:
```suggestion
	The members of each [=W3C Council=] are selected from:
```

dwsinger: that implies some sort of subsetting, whereas it's all of them. "consists of"?

chrisn: But there is subsetting, through the dismissal and renunciation process, so "selected from" seems more accurate

TallTed:
```suggestion
	Each [=W3C Council=] is comprised of the following, modulo renunciations and/or dismissals:
```

cwilso: But this kinda goes back to the original wording.  Regardless, I would advise against the word "modulo", which might not be as accessible to non-native English speakers.  (Yes, I know modulo is Latin.  :) )

dwsinger: I'm back to "Each Council is formed from" or "Each Council consists of" and getting into dismissal and renunciation later…

nigelmegitt: @dwsinger what was the problem with "are selected from"? As I see it, it is no longer than "are comprised from", is just as easy to read, and is a more precise summary of what happens without getting into all the details of dismissal/renunciation.

frivoal: This is very much a nuance, but to me "are selected from" gives the impression that we're going to pick a small group from a big pool. Not that we're intending to use pretty much the entire pool with a few exceptions.

dwsinger: yes, Florian has it.The entire AB and TAG form the council unless they get dismissed or have strong reasons to renounce, both of which should be rare. We do not select a subset.

nigelmegitt: I feel this is actually a contentious point, and I for one have not come round to agreeing that it should be rare, nor that the strong expectation should be that the entire AB and TAG form the council every time. _But_ putting that to one side, here's an alternative suggestion:

```suggestion
	Each [=W3C Council=] is initially composed from:
```

frivoal: I support the "Each", but what's "initially" for? Council composition is not expected to change over time.

nigelmegitt: The "initially" is to say "here are the starting people, there is a process that _could_ change that later". I didn't mean to indicate a high level of change once the Council membership has been decided. The thread indicated discomfort with being lengthily descriptive, but also discomfort with being over-simplistic and therefore inaccurate.

TallTed:
```suggestion
	Each [=W3C Council=] is comprised of the following, minus any renunciations and/or dismissals:
```

frivoal: @TallTed 's latest suggestion isn't as short as the opening proposal, but it is shorter than the existing text, and (to me eye) has no problematic nuance. I'd be inclined to take it.

fantasai:
```suggestion
	Each [=W3C Council=] is composed of the following members (excepting any [[#council-participation|renounced or dismissed]]):
```

frivoal: I'm OK with either @TallTed's or @fantasai 's suggestions above.

nigelmegitt: I'm OK with either, too.

nigelmegitt (approved):

Issue #710, opened 2023-02-06
"The W3C Council" should often be "A W3C Council"


TallTed:
_Originally posted by @TallTed in https://github.com/w3c/w3process/pull/702#discussion_r1094906484_

Given that there may be multiple `W3C Councils` operating at any given moment, there are a number of occurrences of `the W3C Council` which should be replaced by `a W3C Council`.

PR #719, opened 2023-03-20
Speak of *a* Council, not *the* Council

frivoal:
Multiple instances of the Council can exist at the same time, so it seems better to avoid speaking of "the Council" in general, unless context makes it clear which specific Council we are talking about.

See #710

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -864,7 +864,7 @@ Role of the Advisory Board</h5>
 	soliciting Member comments on such issues,
 	and proposing actions to resolve these issues.
 	The Advisory Board manages the <a href="#GAProcess">evolution of the Process Document</a>.
-	As part of the [=W3C Council=],
+	As part of a [=W3C Council=],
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):

index.bs
```diff
@@ -2314,15 +2314,15 @@ Investigation and Mediation by the Team</h4>
 	Otherwise,
 	upon concluding that consensus cannot be found,
 	and no later than 90 days after the [=Formal Objection=] being registered,
-	the [=Team=] <em class=rfc2119>must</em> initiate formation of the [=W3C Council=],
+	the [=Team=] <em class=rfc2119>must</em> initiate formation of a [=W3C Council=],
 	which should be <a href="#council-convention">convened</a> within 45 days of being initiated.
-	Concurrently, it must prepare a report for the [=Council=]
+	Concurrently, it must prepare a report for a [=Council=]
```
TallTed, line 2319: This one should be `the`, as it refers to `the` Council just formed.
```suggestion
	Concurrently, it must prepare a report for the [=Council=]
```

plehegar (approved):

PR #722, opened 2023-03-25
Confidentiality Levels and Redactions

joshco:
Add inline definition of "reasonable efforts" as "including, but not limited to, Confidentiality Levels and Redaction"

***
<a href="https://pr-preview.s3.amazonaws.com/joshco/w3process/pull/722.html" title="Last updated on Apr 15, 2023, 3:00 AM UTC (3716f27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/722/4034ccb...joshco:3716f27.html" title="Last updated on Apr 15, 2023, 3:00 AM UTC (3716f27)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4953,7 +4953,7 @@ Confidentiality Levels</h3>
 			<em class="rfc2119">must</em> treat the information as confidential within W3C,
 
 		<li>
-			<em class="rfc2119">must</em> use reasonable efforts to maintain the proper level of confidentiality, and
+			<em class="rfc2119">must</em> use reasonable efforts to maintain the proper level of confidentiality, including, but not limited to, Confidentiality Levels and Redaction, and
 
 		<li>
 			<em class="rfc2119">must not</em> release this information to the general public or press.
--- a/index.bs
+++ b/index.bs
@@ -4953,7 +4953,7 @@ Confidentiality Levels</h3>
 			<em class="rfc2119">must</em> treat the information as confidential within W3C,
 
 		<li>
-			<em class="rfc2119">must</em> use reasonable efforts to maintain the proper level of confidentiality, including, but not limited to, Confidentiality Levels and Redaction, and
+			<em class="rfc2119">must</em> use reasonable efforts to maintain the proper level of confidentiality, including, but not limited to, applying Confidentiality Levels and using Redaction,
 
 		<li>
 			<em class="rfc2119">must not</em> release this information to the general public or press.


fantasai (changes requested):
"Confidentiality Levels" is a thing and "Redaction" is an action, so it's not really making sense. What are you trying to say here and what's the problem it's trying to solve?

joshco:
Adjusted grammar thing/action. (This PR doesn't seem urgent enough to complicate AB review.  )

The origin of this was on the last CG call, discussing the edge case of a message or other status update that was set to confidentiality level Public or Member, but a contained piece of information needed to remain confidential.  Personal or company identifying information could be redacted within the status update which was set to Public or Member.  

The word redact wasn't present in the document, and there wasn't really a binding between "must use efforts" and what those "efforts" are.  A newcomer might not implicitly make the assumption.

I'm keying off statements I've heard about the doc being long, or difficult to consume.   (Verbosity:  The curse of being a standards wonk)  While I'm new to W3C's process, I was the founding chair of DMTF's Process Committee 2006-2012[1].  We were documenting unwritten rules and adding new processes as the org evolved.  So feel free to let me know to to help best.

[1]: https://www.dmtf.org/about/officers/history

TallTed:
@joshco — Nit in your https://github.com/w3c/w3process/pull/722#issuecomment-1509486443. `2012[1](https://www.dmtf.org/about/officers/history)` would be better `2012 [[1](https://www.dmtf.org/about/officers/history)]` which renders as you will see below, instead of being easily read as a confusing probable typo, `20121`.

2012 [[1](https://www.dmtf.org/about/officers/history)]

TallTed (approved):

fantasai (commented):
Ok, so it's grammatical now, but...

- What does “applying Confidentiality Levels” mean?
- What does “using Redaction” mean? Why is this capitalized?

Altogether it reads very mysteriously.

tantek (approved):

tantek (changes requested):
I'm ok with such examples, yet also support @fantasai's questions about defining these terms. Since this is specifically in the context of Member-only and Team-only, one simplification would be:

s/applying Confidentiality Levels and using Redaction/explicitly labeling "Member-only" or "Team-only", or redacting such information

This avoids duplicating "confidentiality" as well as introducing new Capitalized Terms that imply a formal definition,

joshco:
How about we come to consensus on the next call?

frivoal:
@joshco , I'd like to confirm I understand what you're trying to achieve here. I believe your goal is to state that "use reasonable efforts to maintain the proper level of confidentiality" involves (at least):
* respecting confidentiality levels classification
* if having to expose a document/information to a broader confidentiality level, redacting the parts that wouldn't be appropriate to share

Is that the core of it, or are you trying for something else?

LAZURDFLOWERS (approved):

joshco:
@frivoal You are correct in your assumption of my goal.

@nigelmegitt the issues you raise are good questions.

I'm new to the document, so for me, it wasn't clear what the section means in practice.  Eg, what is someone supposed to do?

frivoal:
Action on me to create an alternative pull request, based on the understanding gained from the discussion as captured in the minutes above.

PR #740, opened 2023-04-25
Don't say that dissent goes away when it's overruled

jyasskin:
***
<a href="https://pr-preview.s3.amazonaws.com/jyasskin/w3process/pull/740.html#ACReviewAfter" title="Last updated on Apr 26, 2023, 12:10 AM UTC (fca1549)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/740/ce50d3d...jyasskin:fca1549.html" title="Last updated on Apr 26, 2023, 12:10 AM UTC (fca1549)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2632,8 +2632,7 @@ After the Review Period</h4>
 	whether there were any [=Formal Objections=],
 	with attention to <a href="#confidentiality-change">changing the confidentiality level</a> of the [=Formal Objections=].
 
-	If there is [=dissent=]
-	(i.e., there were [=Formal Objections=], at least some of which were sustained)
+	If there were [=Formal Objections=], at least some of which were sustained,
 	or if there is not [=consensus=] because of insufficient support,
 	[=W3C Decision=] <em class=rfc2119>must</em> be one of:
 
@@ -4316,7 +4315,7 @@ Abandoning a W3C Recommendation</h5>
 <h5 id="deactivation">
 Process for Rescinding, Obsoleting, Superseding, Restoring a Recommendation</h5>
 
-	The process of rescinding, obsoleting, 
+	The process of rescinding, obsoleting,
 	superseding,
 	or restoring
 	a [=Recommendation=] can be initiated
@@ -5609,7 +5608,7 @@ Acknowledgments</h2>
 	Robin Berjon (W3C),
 	Sally Khudairi (W3C),
 	Sam Ruby (IBM),
-	Sam Sneddon,		
+	Sam Sneddon,
 	Sandro Hawke (W3C),
 	Sangwhan Moon (Odd Concepts),
 	Scott Peterson (Google),


nigelmegitt:
I don't read the previous text as saying that the dissent went away when it's overruled. This change doesn't look meaningful, or particularly harmful to me. Please could you explain the motivation for the change @jyasskin ?

index.bs
```diff
@@ -2632,8 +2632,7 @@ After the Review Period</h4>
 	whether there were any [=Formal Objections=],
 	with attention to <a href="#confidentiality-change">changing the confidentiality level</a> of the [=Formal Objections=].
 
-	If there is [=dissent=]
-	(i.e., there were [=Formal Objections=], at least some of which were sustained)
+	If there were [=Formal Objections=], at least some of which were sustained,
```
TallTed, line 2635: There was discussion earlier, that dissent could exist without Formal Objections, which seems to be overwritten by the PR, as written. This tweak doesn't quite put less-intense dissent back, but it at least doesn't say that sustained Formal Objections are the *only* possible form of dissent....
```suggestion
	If there is [=dissent=]
	(e.g., there were [=Formal Objections=], at least some of which were sustained),
```

jyasskin: This tweak says that one dissenting member can force the proposal to be rejected or returned for more work, even if their formal objection is overruled. I don't think that's what we want the Process to say?

jyasskin (commented):
Re @nigelmegitt, the existing text says that "dissent" can be restated as "there were Formal Objections, at least some of which were sustained". That implies that, if there were formal objections that were all overruled, there isn't dissent. But that's not true, so I suggested this change to avoid that implication. 

We do [define "dissent"](https://www.w3.org/Consortium/Process/Drafts/#def-Dissent) as "At least one individual in the set sustains an objection.", but that's a different meaning of "sustain" than the one used when dealing with formal objections.

index.bs
TallTed: @jyasskin — I don't understand how you get that meaning.

With `i.e.`, it's "If there is dissent (that is, there were Formal Objections, at least some of which were sustained)".  In other words, dissent can *only* take the form of [sustained] Formal Objections.

With `e.g.`, it's "If there is dissent (for example, there were Formal Objections, at least some of which were sustained)".  In other words, dissent *may* take the form of [sustained] Formal Objections, but there are also other forms of dissent which may apply in this section.

(Here we have another reason to avoid `i.e.` and `e.g.,` even though the previous advice from the i18n folks has been reconsidered.)

jyasskin: Right, so your use of `e.g.` is an improvement with respect to the meaning of "dissent", but I don't think it results in the right meaning of the whole paragraph.

The [whole paragraph](https://www.w3.org/Consortium/Process/Drafts/#ACReviewAfter) we're discussing is:

> If there is [dissent](https://www.w3.org/Consortium/Process/Drafts/#def-Dissent) (...) or ..., [W3C Decision](https://www.w3.org/Consortium/Process/Drafts/#def-w3c-decision) must be one of:
> * The proposal is returned for additional work, with a request to the initiator to improve the proposal.
> * The proposal is rejected.

Dissent is created by a single member. There's a statement in the [previous section](https://www.w3.org/Consortium/Process/Drafts/#ACReviewStart) that dissent in this context must be expressed as a Formal Objection, but whether or not we agree with that, the existence of the dissent doesn't depend on whether the FO Council agrees with that member. But the paragraph then says that this single member can force the W3C Decision to be to either reject or return the proposal. That doesn't seem right to me.

The current wording uses `i.e.` to (badly) redefine "dissent" so that the rest of the paragraph works: it says that if some Formal Objections were sustained, then the proposal has to be returned or rejected. Changing the `i.e.` to `e.g.` avoids redefining "dissent" (good), but then gives single dissenters too much power (bad). My suggestion is to remove the use of "dissent" in this paragraph, since the W3C Decision depends on the disposition of the FO, not on whether someone disagrees.

TallTed: > My suggestion is to remove the use of "dissent" in this paragraph, since the W3C Decision depends on the disposition of the FO, not on whether someone disagrees.

I think that's a good direction.

dwsinger:
I think I agree with Nigel. The question of whether the objector sustains their objection lies solely with the objector. Whether they _prevail_ in their objection is a question of what decision is reached, which does not lie with them.

jyasskin:
I note that the [Process's definition of "sustain"](https://www.w3.org/Consortium/Process/Drafts/#sustain) is about formal objections. It would be possible to change that word globally, but probably easier to change the word used for persistent dissent.

dwsinger:
you're right, we could use the word in both senses (does the objector sustain the objection, does the Council sustain or overrule). that's potentially confusing.

fantasai (approved):

fantasai:
Note: We [changed](https://github.com/w3c/w3process/pull/762) "sustain" to "uphold" when referring to a Council Decision, to resolve this and other confusion in the Process.

With that out of the way, I think we should accept this change, because the resulting text is more straightforward.

Issue #749, opened 2023-04-30
Conflict of Interest in Council composition


mnot:
> Note: Since dismissal is individual, when the decision being objected to was made by the [TAG](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#technical-architecture-group) or [AB](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#advisory-board) acting as a body, the entire [TAG](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#technical-architecture-group) or [AB](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#advisory-board) is not expected or required to be dismissed.

The optics on this aren't very good; a body that makes a decision that's being appealed should not have oversight of the appeal.

The straightforward thing to do is to use an alternative process for appeals of TAG or AB decisions -- probably have the AB as the appeal body for the TAG, and the Board as the appeal body for the AB.

plehegar:
I would note that the current Process has the same flaw: nothing forces the Director to delegate his authority, even if he has done so in the past on many occasion, such as the WWW 1.0 architecture TAG document moving to Recommendation. We could certainly add to our guidelines on this subject and potentially add to the Process in a future iteration if it remains unsatisfying.

jyasskin:
For what it's worth, I think the method for [Appealing Council Decisions to the AC](https://www.w3.org/Consortium/Process/Drafts/#council-appeals) is likely to be sufficient in the case that a TAG or AB decision is appealed and upheld, but members think the Council's deliberations were unfair because the decision-makers got to vote on the appeal of their own decision.

For contentious decisions, the body that made the decision isn't likely to have made it unanimously, and if the rest of the council is convinced by the appeal, they're likely to be able to join with the minority of the original body to uphold the appeal.

It's true that the optics aren't perfect, but I also don't like the optics of declaring that the Board ([authority over finances and operations](https://www.w3.org/Consortium/Process/Drafts/#bod)) would be more trusted than the AB ([authority over strategy, management, legal matters, process, and conflict resolution](https://www.w3.org/Consortium/Process/Drafts/#ab-role)), which would be more trusted than the TAG ([authority over architecture](https://www.w3.org/Consortium/Process/Drafts/#tag-role)).

mnot:
[...]

PR #761, opened 2023-05-19
Exclude TAG/AB from voting on their own decisions/proposals. #749

fantasai:
There are two variants possible: see bracketed text, which could be kept or removed.

(Personally, I prefer removing it, for simplicity; but could live with either way.)

***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/761.html" title="Last updated on Oct 11, 2023, 2:55 PM UTC (a000179)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/761/321e97e...fantasai:a000179.html" title="Last updated on Oct 11, 2023, 2:55 PM UTC (a000179)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2520,7 +2520,7 @@ Council Deliberations</h5>
 	However,
 	if the decision or proposal being objected to
 	originated with the TAG or AB,
-	then members of that group [at the time the decision or proposal was made]
+	then members of that group at the time the decision or proposal was made
 	must abstain in such a vote.
 	In case of a vote,
 	if two members of a Council who share the same affiliation cast an identical ballot,
--- a/index.bs
+++ b/index.bs
@@ -2517,6 +2517,11 @@ Council Deliberations</h5>
 	In that case,
 	the decision is made by simple majority,
 	with the [=W3C Council Chair=] breaking any tie.
+	However,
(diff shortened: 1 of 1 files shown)


dwsinger:
The bracketed text I think needs to say (if we keep it) "those who were members of the group when the decision was made are excluded from any vote" (it's possible someone left the TAG and joined the AB, for example). It's complex, but I think it gives the correct appearance – those who made the decision against which the FO was filed can't vote. But it also opens many cans of worms – what if an AB or TAG member was a chair of a group, or the primary editor? It also ignores the possibility that the group was divided on the original decision. Overall, I would go without this bracketed text: there are too many corner cases.

nabilaljordan (approved):

index.bs
```diff
@@ -2517,6 +2517,11 @@ Council Deliberations</h5>
 	In that case,
 	the decision is made by simple majority,
 	with the [=W3C Council Chair=] breaking any tie.
+	However,
+	if the decision or proposal being objected to
+	originated with the TAG or AB,
+	then members of that group [at the time the decision or proposal was made]
```
frivoal, line 2523:
```suggestion
	then members of that group at the time the decision or proposal was made
```

Issue #699, opened 2023-01-23
Mandate default custodianship for Registries if the specified custodian is no longer available


nigelmegitt:
Arising from the [TTWG's call on 2023-01-19](https://www.w3.org/2023/01/19-tt-minutes.html#t05) and discussion ongoing to create boilerplate for TTWG registries at w3c/ttwg#241, where the idea is to specify that TTWG is the custodian of its own registries, but if TTWG ceases to operate then custodianship would fall back to the Team:

The question arose: shouldn't the Process define a fallback custodian for all W3C Registries if the originally specified custodian is unable to operate? More broadly, if the Registry Definition's process for requesting changes to the Registry is defunct (perhaps the email address has been frozen, etc) what is the default fallback?

It seems unhelpful or impractical for a WG to have to define a process for what happens when the WG no longer exists.

frivoal:
Having the Team as a fall back seems useful, though we probably want to let groups explicitly opt out of that if for some reason their registry has update criteria that wouldn't be appropriate to hand off to the Team.

Maybe we also want to give the Team the ability, when it inherits custody of something, to assign it to some other body, possibly with an AC review to confirm.

TallTed:
> give the Team the ability, when it inherits custody of something, to assign it to some other body

Exactly what occurred to me on reading the OP.

> registry has update criteria that wouldn't be appropriate to hand off to the Team

I suggest that any group realizing such should nominate a few (3? 5? 10?) groups that are expected to endure sufficiently, which groups should then resolve their intent to serve as a potential fall back, among which I would think the Team could choose when needed ... or the nominating group could order their list by preference, which guidance I would think could be followed as with any creator's wishes about their output.

dwsinger:
whoops, good catch. 

(A few years back I had a need to contact all the [Registration Authorities approved by ISO](https://www.iso.org/maintenance_agencies.html), and was unhappy to find how many dead pointers there were. We can and should do better.)

I think the ideal route is something like:
* if the operator of a Registration Authority wishes to retire, they tell us, and we manage the transition in conversation; if the authoring WG still exists, it goes back to them to revise the Rec-track document that defines the authority
* if we close a WG and we know that they are the approval/operator, we ask them to work out what to do before closure
* if nonetheless we end up with a 'dead pointer', yes, the team should have the authority both to become the operator and also the authority to find and assign a better if they wish (and if you don't want the team to do that, well, see steps 1 and 2 and it won't happen)

[...]

PR #790, opened 2023-10-11
Enable Team to replace defunct registry custodians when no-one else can

frivoal:
See issue #699

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4646,6 +4646,13 @@ Registry Definitions</h4>
 			The [=custodian=] may be the [=Working Group=], the [=Team=], or a delegated entity.
 			The [=custodian=] for all [=registry tables=] in a single [=registry=]
 			<em class=rfc2119>should</em> generally be the same entity.
+
+			If a [=custodian=] of a [=registry=] becomes defunct
+			(e.g. the relevant group is disbanded),
+			and the [=Working Group=] that owns the [=registry definition=] is itself also defunct,
+			the [=Team=] <em class=rfc2119>should</em> propose replacing the [=custodian=],
+			which <em class=rfc2119>must</em> be confirmed
+			by an [=AC Review=] as a [=W3C Decision=].
 	</ul>
 
 <h4 id=reg-pub>


index.bs
```diff
@@ -4646,6 +4646,13 @@ Registry Definitions</h4>
 			The [=custodian=] may be the [=Working Group=], the [=Team=], or a delegated entity.
 			The [=custodian=] for all [=registry tables=] in a single [=registry=]
 			<em class=rfc2119>should</em> generally be the same entity.
+
+			If a [=custodian=] of a [=registry=] becomes defunct
+			(e.g. the relevant group is disbanded),
```
TallTed, line 4651:
```suggestion
			(e.g., the relevant group is disbanded),
```

index.bs
```diff
@@ -4646,6 +4646,13 @@ Registry Definitions</h4>
 			The [=custodian=] may be the [=Working Group=], the [=Team=], or a delegated entity.
 			The [=custodian=] for all [=registry tables=] in a single [=registry=]
 			<em class=rfc2119>should</em> generally be the same entity.
+
+			If a [=custodian=] of a [=registry=] becomes defunct
```
TallTed, line 4650:
```suggestion
			If the [=custodian=] of a [=registry=] becomes defunct
```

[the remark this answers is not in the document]
TallTed: -- or this --
```suggestion
			If the [=custodian=] of a [=registry=] or [=registry table=] becomes defunct
```

TallTed (changes requested):
I think either of these would address what we discussed today.

index.bs
```diff
@@ -4646,6 +4646,13 @@ Registry Definitions</h4>
...
+			(e.g., the relevant group is disbanded),
```
dwsinger, line 4651: (or the custodian is unresponsive to repeated attempts to make contact)

index.bs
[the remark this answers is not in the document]
dwsinger: not sure I like the word "defunct", how about "ceases to exist or ceases to operate as a custodian"?

index.bs
TallTed:
```suggestion
			(e.g., the relevant group is disbanded, or 
			the custodian is unresponsive to repeated attempts to make contact),
```

index.bs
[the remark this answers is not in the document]
TallTed: tweaked my suggestions above to include @dwsinger's replacement for `defunct`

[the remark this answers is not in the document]
nigelmegitt: I prefer not to use "defunct" too.

index.bs
```diff
@@ -4646,6 +4646,13 @@ Registry Definitions</h4>
...
+			and the [=Working Group=] that owns the [=registry definition=] is itself also defunct,
```
nigelmegitt, line 4652: Feels like we should have a defined term for the concept of a WG that's no longer operating - "defunct" feels wrong here. Maybe this is also a bit stronger than "does not have an active Charter" since as we have seen, WGs may effectively continue to operate for a while during Charter renegotiations that go on beyond the previous Charter's end point, for a while anyway - extensions are not always granted synchronously, for practical reasons.

dwsinger: perhaps we should replace 'defunct' with what we have as examples, "the custodian ceases to exist, or is unresponsive to repeated attempts to make contact"?

nigelmegitt: §4.6 describes ["Chartered Group Closure"](https://www.w3.org/2023/Process-20230612/#GeneralTermination) - we could either define a team for "Closed Group" within that section, or just use link to that section here, e.g.:

```suggestion
			and the [=Working Group=] that owns the [=registry definition=] is itself <a href="#GeneralTermination">closed</a>,
```

TallTed, line 4652: Incorporating all previous suggestions from @dwsinger, @nigelmegitt, and myself —
```suggestion
			If the [=custodian=] of a [=registry=] or [=registry table=]
			ceases to exist or to operate as a custodian
			(e.g., the relevant group is disbanded, or 
			the custodian is unresponsive to repeated attempts to make contact),
			and the [=Working Group=] that owns the [=registry definition=] 
			is itself <a href="#GeneralTermination">closed</a> or 
			unresponsive to repeated attempts to make contact,
```

I still prefer `or portion thereof` over `or [=registry table=]`, but I've not included it in this aggregate suggestion because nobody else has voiced a preference.

TallTed (changes requested):

index.bs
nigelmegitt: I prefer it without "or portion thereof" - I don't think that phrase adds anything useful, unless there are some registries that have specific subsections whose custodians are different to the rest of the document, other than the registry table, and if that is the case, it seems strange and I'd like to know more about the use case.

TallTed: As I understand it, any `[=registry=]` may be defined to have multiple `[=registry tables=]`, and any subset of `[=registry tables=]` (1–_n_) might have a different `[=custodian=]` than any other subset, up to the extreme of each and every `[=registry table=]` within a `[=registry=]` having a different `[=custodian=]`, which might also differ from the `[=registry=]` itself.

I took the `SHOULD` in (lines 4647-4648 above) —
```
The [=custodian=] for all [=registry tables=] in a single [=registry=]
<em class=rfc2119>should</em> generally be the same entity.
```
— as an indication that there was already one or more `[=registries]` known or likely to be defined with different `[=custodians=]` for one or more `[=registry tables=]` therein.

I prefer `or portion thereof` because its the fewest words to allow for any subset (1–_n_) of the `[=registry tables=]` within a `[=registry=]` to be assigned to a new `[=custodian=]` without requiring that _all_ `[=registry tables=]` within that `[=registry=]`, or even the `[=registry=]` itself, be reassigned at the same time. I like this ability to "fix in part" because I can foresee time constraints being in play for the initial `[=registry table(s)=]` where the issue was discovered, while there is more time flexibility on the remaining `[=registry table(s)=]` and/or the `[=registry=]` itself.

TallTed: As discussed in today's call
```suggestion
			If the [=custodian=] of a [=registry table=]
			ceases to exist or to operate as a custodian
			(e.g., the relevant group is disbanded, or 
			the custodian is unresponsive to repeated attempts to make contact),
			and the [=Working Group=] that owns the [=registry definition=] 
			is itself <a href="#GeneralTermination">closed</a> or 
			unresponsive,
```

PR #792, opened 2023-10-12
Limit the involvement of Tim Berners-Lee to the TAG proper

frivoal:
As the Council's purpose is to replace his former role as Director in resolving Formal Objections, this removes him from being part of the Council. This way, he does not take part in the Council's deliberations, and is not on the critical path for Council votes or quorum questions, notably the Council's Unanimous Short Circuit.

This is one possible solution to #784, with #791 and #793 being alternatives.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/792.html" title="Last updated on May 30, 2024, 2:03 PM UTC (8f0f189)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/792/01bd370...frivoal:8f0f189.html" title="Last updated on May 30, 2024, 2:03 PM UTC (8f0f189)">Diff</a>

Issue #784, opened 2023-09-26
Council Composition requirements include Tim Berners-Lee, TAG life member


plehegar:
As part of using [5.6.2.4. Unanimous Short Circuit](https://www.w3.org/2023/Process-20230612/#council-short-circuit), we realize that @timbl was unintentionally left on the critical path of W3C Councils.
We should make sure @timbl isn't interested in being on the critical path and update the Process if that's the case.

frivoal:
Assuming this is compatible with what @timbl  wants, I suspect a solution to this problem is to change the status of Tim's participation in the TAG. Currently, he is listed as a "life member" of the TAG. Instead, I think we could draw inspiration from the status of the CEO in the AB, and go for something like:

> As [Director Emeritus | Founding Director], Tim Berners Lee has, for life, a standing invitation to all regular TAG sessions.

This would enable participation to all standard TAG activities, without being on the critical path for Unanimous Short Circuit, or any other activity that might involve voting thresholds or quorum.

This would also mean that @timbl would not be a member of Councils in general (not just for Unanimous Short Circuit), but given that he has not participated so far, it is not clear to me that he wants to be part of them.

Based on the speeches given at the AC meeting in Sophia Antipolis, I would expect this change to be in accordance with @timbl's view of his role going forward, but getting confirmation would be nice.

plehegar:
[[
plh: the council rules currently place TBL on the critical path for applying the unanimous short circuit. We never intended him to be on the critical path, so we should fix this in 2024. (Presuming TBL is okay with that.)

florian: I've made a specific suggestion, would like to know what TBL thinks.

plh: yes.

<fantasai> I think the alternative path would be to specifically make him an optional member of the Councils.

<fantasai> but remain a formal member of the TAG

plh: ...we will need to ensure the TAG is okay with your resolution as well.
]]
https://www.w3.org/2023/09/27-w3process-minutes.html#t04

frivoal:
Made three potential pull requests to address this issue. From the broadest to the narrowest: #791, #792, #793.

After checking that the wording in each is reasonably phrased, we should check with Tim to see if he has a preference.

hober:
Between the three <abbr title="Pull Requests">PRs</abbr>, I prefer #792. That said, please do see my suggested rewording of it.

frivoal:
AB Resolution:
> RESOLUTION: Adopt the proposed text in Process PR #792, as recommended by the TAG.

The PR has been merged. Closing.

PR #793, opened 2023-10-12
Allow Tim Berners-Lee to abstain in Unanimous Short Circuit votes

frivoal:
This removes him from the being on critical path for the Council's Unanimous Short Circuit votes, without affecting any other role he has as a TAG and Council member.

This is one possible solution to #784, with #791 and #792 being alternatives.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 7, 2024, 6:13 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL]([object Object])

```
Error running preprocessor, returned code: 2.
Your document appears to use tabs to indent, but line 2540 starts with spaces.
FATAL ERROR: Line 2540 isn't indented enough (needs 2 indents) to be valid Markdown:
"      (in addition, Tim Berners-Lee &lt;em bs-line-number=2540 class="rfc2119">may&lt;/em> also abstain)"
LINK ERROR: No 'dfn' refs found for 'advisory committee override'.
[=Advisory Committee Override=]
 ✘  Did not generate, due to errors exceeding the allowed error level.
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/process%23793.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2453,6 +2453,8 @@ Unanimous Short Circuit</h5>
 	the Team recommends a resolution
 	and every potential member of a Council who is not renouncing their seat
 	votes affirmatively (no abstentions) to adopt this resolution.
+	As an exception, Tim Berners Lee <em class=rfc2119>may</em> abstain
+	without blocking the adoption of the resolution.
 
 	This step <em class=rfc2119>may</em> be run concurrently with [[#council-participation]]
 	and prior to choosing a [=W3C Council Chair|Chair=].
--- a/index.bs
+++ b/index.bs
@@ -2453,7 +2453,7 @@ Unanimous Short Circuit</h5>
 	the Team recommends a resolution
 	and every potential member of a Council who is not renouncing their seat
 	votes affirmatively (no abstentions) to adopt this resolution.
-	As an exception, Tim Berners Lee <em class=rfc2119>may</em> abstain
+	As an exception, Tim Berners-Lee <em class=rfc2119>may</em> abstain
 	without blocking the adoption of the resolution.
 
 	This step <em class=rfc2119>may</em> be run concurrently with [[#council-participation]]


index.bs
```diff
@@ -2453,6 +2453,8 @@ Unanimous Short Circuit</h5>
 	the Team recommends a resolution
 	and every potential member of a Council who is not renouncing their seat
 	votes affirmatively (no abstentions) to adopt this resolution.
+	As an exception, Tim Berners Lee <em class=rfc2119>may</em> abstain
```
dwsinger, line 2456: Hyphen missing again

frivoal, line 2456:
```suggestion
	As an exception, Tim Berners-Lee <em class=rfc2119>may</em> abstain
```

index.bs
```diff
@@ -2453,6 +2453,8 @@ Unanimous Short Circuit</h5>
 	the Team recommends a resolution
 	and every potential member of a Council who is not renouncing their seat
 	votes affirmatively (no abstentions) to adopt this resolution.
+	As an exception, Tim Berners-Lee <em class=rfc2119>may</em> abstain
+	without blocking the adoption of the resolution.
```
nigelmegitt, line 2457: If we take #791 as introducing the concept of a standing _invitation_, then this could be a better way to address the general problem of invited members of TAG being on the blocking path:

```suggestion
	and every potential member of a Council who is not renouncing their seat
	or declining their invitation
	votes affirmatively (no abstentions) to adopt this resolution.
```

tantek (changes requested):
I agree with @dwsinger's concern. Absent context/linking, "abstain" implies explicit abstain, and that's insufficient to solve this use-case.

We need different language here, perhaps something about voting or not rather than just absention. 

E.g. something like: 

As an exception, the voting affirmatively requirement does not apply to Tim Berners-Lee, who may omit his vote or abstain.

frivoal:
I guess that's the risk when you let a French speaker word things :) The French meaning of abstain is "omit to vote", while voting an explicit non-opinion would be called something like "voting blank". Now that I know English is different, I support @tantek 's proposed rephrasing.

index.bs
frivoal: This PR and #791 are alternatives to eachother. We should take or or the other (or #792), but there's not need to take both.

frivoal:
I suspect that this PR no longer serves a useful purpose given that https://github.com/w3c/process/pull/885 has been adopted. #791 or #792 may remain of interest, but I think we can close this one. Agenda+ to confirm.

index.bs
```diff
@@ -2453,6 +2453,8 @@ Unanimous Short Circuit</h5>
 	the Team recommends a resolution
 	and every potential member of a Council who is not renouncing their seat
 	votes affirmatively (no abstentions) to adopt this resolution.
+	As an exception, Tim Berners-Lee <em class=rfc2119>may</em> abstain
```
tantek, line 2456:
```suggestion
	As an exception, the voting affirmatively requirement does not apply to Tim Berners-Lee, 
	who <em class=rfc2119>may</em> omit his vote or abstain.
```

index.bs
tantek, line 2457:
```suggestion
```

tantek (changes requested):
I think I figured out how to propose the wording change I suggested in prose and accepted by @frivoal.

I think even with the adoption of [#885](https://github.com/w3c/process/pull/885), this is still useful. Even with consideration of [#791](https://github.com/w3c/process/pull/791) or [#792](https://github.com/w3c/process/pull/792), this helps to keep this part of the Process less dependent on a piece of clarification somewhere else.

I think this is a positive improvement that stands on its own that we should adopt, without impacting consideration of the other proposals.

index.bs
TallTed, line 2457:
```suggestion
	and every potential member of a Council who is not renouncing their seat
	(with the possible exception of Tim Berners-Lee)
	votes affirmatively (no abstentions,
	again with the possible exception of Tim Berners-Lee)
	to adopt this resolution.
```

frivoal: This edit no longer makes sense, since the underlying text has been changed by #885.

frivoal: This edit no longer makes sense, since the underlying text has been changed by #885.

index.bs
frivoal: This edit no longer makes sense, since the underlying text has been changed by #885.

frivoal:
A few recent rephrasing suggestions no longer makes sense, since the underlying text has been changed by #885. I have rebased on top of the newest text, and included a simple provision that achieves the same effect. Feel free to propose improved rephrasings if you think that is useful, but personnaly, I don't seen the point: we are now allowing up to 20% of the council to abstain. Tim Berners-Lee can be part of these 20%. Regardless of phrasing, I don't find it useful to say that 20% may abstain if Timbl is among those who vote, but 20% + 1 person may abstain if Timbl is among those who abstain. That seems complicated for no good reason.

Therefore, my preference would be to close this as overtaken by events (https://github.com/w3c/process/pull/791, https://github.com/w3c/process/pull/792 remain relevant). But if others see value in this, I have no objection to finding better phrasing.

index.bs
```diff
@@ -2537,13 +2537,15 @@ Short Circuit</h5>
 	and potential members of a Council who are not renouncing their seat
 	confirm it by a vote which results in both of the following:
 	* at least 80% of them vote affirmatively to adopt this resolution
+      (in addition, Tim Berners-Lee <em class=rfc2119>may</em> also abstain)
```
TallTed, line 2540: "In addition ... also" is odd phrasing. This might work --
```suggestion
	* Tim Berners-Lee either abstains or votes affirmatively to adopt this resolution
```

TallTed: Or this, with a bit of future proofing, even if it's an unlikely future, and making it a little less Timbl-specific (even if that is the underlying intent) —
```suggestion
	* Directors Emeriti either abstain or vote affirmatively to adopt this resolution
```

TallTed (changes requested):

index.bs
dwsinger: and if he votes no? I am unclear as to what is trying to be achieved here. If it's that it's 80% of the body without Tim, then say that Tim does not count towards quorum, but may voyte, abstain, or fail to vote.

TallTed: I presume, if he votes "no", then the last bullet is triggered (i.e., one of "the potential members of the Council who did not renounce their seat" voted against adopting the resolution), and the resolution fails.

(As I understand this, there's an intent that only Directors Emeriti can submit an abstention; the other potential Council members can only abstain by not submitting a ballot.)

```suggestion
	* Directors Emeriti may record a vote of abstention, unlike other potential members of the Council, who can only abstain by not voting
```

Clearer formulation may require a more complete rephrasing.

frivoal: > As I understand this, there's an intent that only Directors Emeriti can submit an abstention; the other potential Council members can only abstain by not submitting a ballot.

No. At least that's not my intent. The original intent was that only Timbl can abstain (by not voting), and everybody else has to vote affirmatively, or the short circuit fails. We have since (in #885) relaxed the rule to say that anybody can abstain (by not voting) as long as it's not too many people doing so. I believe this makes this special rule about Timbl redundant and unnecessary, and that we should close this PR.

@tantek has said that he thinks that even with the adoption of #885, he still thinks this PR is useful, which is why I have rebased it on top of the text that allows for abstention (as long as we stay above 80%). The awkwardness of adding in the Timbl clause comforts me in the idea that this is not helpful.

tantek, line 2540:
```suggestion
```
Thanks for rebasing, and I agree with you that afterwards, this reads more awkwardly than helpful.
Also in taking a step back and considering the broader "why" of this PR, I think an overlooked aspect (which may lead to a cleaner solution) is that a short-circuit is in the context of a *Team* recommendation — this implies to me that the folks already on or affiliated in some way with the Team (including TimBL, both historically and possibly in a remaining soft influence way) have already had a chance to contribute their opinions (perhaps even solicited to do so) as part of that Team recommendation, and presumably are at least "ok" with it, and thus don't need to "double-vote" as it were (once as part of crafting the recommendation, and then again as part of the short-circuit).
Thus I'd drop this line entirely (as it is awkward), and I will propose a different, much smaller change that I believe both addressing the implicit intent of this PR, as well as the broader "problem" it is solving a part of.

index.bs
```diff
@@ -2537,13 +2537,15 @@ Short Circuit</h5>
 	and potential members of a Council who are not renouncing their seat
```
tantek, line 2537:
```suggestion
	and potential members of a Council from elected positions who are not renouncing their seat
```
This has effect of subsetting the short-circuit vote to the elected members of the AB + TAG, which I believe is correct, as the CEO and TimBL have likely already had the chance to contribute to the Team recommendation, and furthermore unburdens the appointed TAG members from feeling pressured to vote (or not) in alignment with the Team who appointed them.

tantek (changes requested):
Good iterations here exploring how to improve this section based on good reasoning.

I added further reasoning in the suggested changes which I believe accomplish the intent of this PR and solve the broader challenge of the Team making a short-circuit recommendation and then having the rest of the (non-Team and affiliated) potential council members vote on it.

index.bs
frivoal: The problem statement (from #784) for this pull request is:
> @timbl was unintentionally left on the critical path of W3C Councils.

I think what you're trying to address here is a distinct question (or even two). I'd suggest opening a new issue/PR (or two) to discuss that separately.

As feedback on this: I think treating the CEO and Director emeritus as part of the Team, and therefore assuming their assent is implied in a Team recommendation is reasonable. However, that logic doesn't apply to appointed TAG members, and the rationale there is different, as you noted. I'm a lot less sure I agree with that part, and in any case, I'd suggest a separate discussion.

dwsinger: I agree. I think once they are on teh TAG, they should be treated the same as every other member.

tantek: @dwsinger wrote:
> I agree. I think once they are on the TAG, they should be treated the same as every other member.

I suppose now that TAG appointees cannot be re-appointed (they have to run for election explicitly to continue serving after their appointed term ends), there is less ongoing connection, potential pressure from, or sense of obligation to the Team appointing them. I'll make a minor edit accordingly at least to capture these suggestions before creating a new issue/PR as @frivoal suggests.

frivoal:
I consider this PR of mine to have become moot since https://github.com/w3c/process/pull/885 has been accepted. When everybody had to affirmatively vote in the short circuits, making an exception for Timbl could make some amount of sense. Now that we do allow for a small number of abstentions anyway, allowing for slightly more abstentions if timbl is part of the people abstaining is unnecessarily convoluted, and brings no clear value. Other open question about his role in the council and tag may valid to discuss, but I am no longer pursuing this particular change.

Issue #800, opened 2023-11-08
Are the rules for updating Registry Definitions appropriate?


nigelmegitt:
Following discussion of the practicalities of including Registries in Rec track documents at https://lists.w3.org/Archives/Public/spec-prod/2023OctDec/0003.html it seems like there may be a Process bug regarding the process for updating Registry Definitions.

[§6.5.2](https://www.w3.org/2023/Process-20231103/#reg-pub) says that registry sections are "purely documentational". What does this actually mean? I initially read it as being "informative", or alternatively "can be changed outside the normal Rec track document process" but [§6.5.1](https://www.w3.org/2023/Process-20231103/#reg-def) requires that every registry table has a defined change process.

It would be absurd to require a strict process for how to change a registry data table but then allow that process itself to be modifiable without its own change process; I doubt that's the intention, but it seems to be the effect at the moment.

Another observation is that §6.5.2 refers to a singular Registry Section, but practically, as per w3c/dapt#196 for example, within a Rec track report incorporating registries, it makes sense to split the components of the Registries into different sections. Large parts of the definitions section are best placed away from the registry data tables themselves, which are best placed in the sections of the specification where they're used and relevant, because they're easier to read there.

So some attention is needed to check if the Process gets the cardinality of each type of section right.

fantasai:
The Registry Definition has to follow the same rules as changes to other normative sections of a REC. The tables are subject to the rules in the Definition. The intention of the Process text there is to clarify that the Definition can't be normative on implementations and has no Patent Policy implications, but probably it's not very clearly explained.

fantasai:
Wrt splitting the various parts of the registry for readability... I think that should be fine, and we can clarify that the Registry Section can in fact be split into multiple sections in various parts of the document. (We might want to call it something other than a Registry Section in that case, but I'm unsure what.)

dwsinger:
Let's try to explain the terms here, and I think it will become clear. 

1) There is text in a Rec-track document that _defines_ the registry, its contents, management etc. That's part of the Rec and is subject to rec-track normal change management and approval. That's the Registry Definition.

2) one of the places you can put the registered material, the registry tables, is in the document that is also the Rec. That section needs to be clearly delineated as being the registry contents; from the point of view of the rec-track document it is purely informational and the rules for its revision are not the rules of rec-track documents, but the rules for that Registry. That's a registry section.

I hope I got this right. And that it makes sense.

nigelmegitt:
[...]

PR #807, opened 2023-12-12
Reduce confusion wrt registry sections #800

fantasai:
* Rename “registry section” to “embedded registry”.
* Move paragraph about Patent Policy to its own section.
* Remove the words “purely documentational” because it's easy to confuse with “informative” / “non-normative”.
* Add “(only)” to the sentence making registries non-normative for Patent Policy purposes.

***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/807.html" title="Last updated on Jan 24, 2024, 4:04 PM UTC (2e75ddd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/807/dcd212d...fantasai:2e75ddd.html" title="Last updated on Jan 24, 2024, 4:04 PM UTC (2e75ddd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2844,7 +2844,7 @@ Types of Technical Reports</h3>
 			in order to document collections of values or other data.
 			These are typically published in a separate [=registry report=],
 			although they can also be directly embedded in [=Recommendation Track=] documents
-			as a [=registry section=].
+			as a [=embedded registry=].
 			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
 			and can even be done without a [=Working Group=].
@@ -4672,20 +4672,12 @@ Publishing Registries</h4>
 
 	[=Registries=] can be published either
 	as a stand-alone [=technical report=] on the [=Registry Track=] called a <dfn>registry report</dfn>,
-	or incorporated as part of a [=Recommendation=] as a <dfn>registry section</dfn>.
+	or incorporated as part of a [=Recommendation=] as a <dfn oldids="registry-section">embedded registry</dfn>.
 
-	A [=registry report=] or [=registry section=]
-	is purely documentational,
-	is not subject to the W3C Patent Policy,
-	and <em class=rfc2119>must not</em> contain any requirements on implementations.
-	For the purposes of the Patent Policy [[PATENT-POLICY]],
-	any [=registry section=] in a [=Recommendation track=] document
-	is not a normative portion of that specification.
-
-	The [=registry report=] or [=registry section=] <em class=rfc2119>must</em>:
+	The [=registry report=] or [=embedded registry=] <em class=rfc2119>must</em>:
 	<ul>
 		<li>
-			Clearly label the [=registry report=]/[=registry section|section=],
+			Clearly label the [=registry report=]/[=embedded registry=],
 			its [=registry tables|tables=],
 			and its [=registry definitions=] as such,
 			including a link to [[#registries]] in this Process.
@@ -4864,6 +4856,16 @@ Specifications that Reference Registries</h4>
 		and implement it as defined by section yy of IETF RFC xxxx”.
 		(The Registry should nonetheless contain </code>Basic-Method</code> as an entry.)
 	</div>
+
+<h4 id=reg-pat>
+Registries and Patents</h4>
+
+	A [=registry report=] or [=embedded registry=]
+	is not subject to the W3C Patent Policy,
+	and <em class=rfc2119>must not</em> define any requirements on implementations.
+	For the purposes of the Patent Policy [[PATENT-POLICY]] (only),
+	any [=embedded registry=] in a [=Recommendation track=] document
+	is not a normative portion of that specification.
 
 <h3 id="switching-tracks">
 Switching Tracks</h3>
@@ -5696,6 +5698,11 @@ Changes since the <a href="https://www.w3.org/2023/Process-20231103/">3 November
 		<li>
 			Stop citing the superseded TAG charter.
 			(See <a href="https://github.com/w3c/w3process/issues/794">Issue 794</a>)
+
+		<li>
+			Rename “registry sections” to [=embedded registries=]
+			to avoid confusion over whether they can be split across multiple sections of a [=Recommendation=].
+			(See <a href="https://github.com/w3c/w3process/issues/800">Issue 800</a>)
 	</ul>
 
 
--- a/index.bs
+++ b/index.bs
@@ -4672,7 +4672,7 @@ Publishing Registries</h4>
 
 	[=Registries=] can be published either
 	as a stand-alone [=technical report=] on the [=Registry Track=] called a <dfn>registry report</dfn>,
-	or incorporated as part of a [=Recommendation=] as a <dfn oldids="registry-section">embedded registry</dfn>.
+	or incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
 
 	The [=registry report=] or [=embedded registry=] <em class=rfc2119>must</em>:
 	<ul>
--- a/index.bs
+++ b/index.bs
@@ -2842,8 +2842,8 @@ Types of Technical Reports</h3>
 		<dd>
 			[=Working Groups=] can also publish [=registries=]
 			in order to document collections of values or other data.
-			These can either be published in a [=registry report=],
-			or directly embedded in [=Recommendation Track=] documents
+			A registry can be published either as a distinct [=registry report=],
+			or directly within [=Recommendation Track=] a document
 			as an [=embedded registry=].
 			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
--- a/index.bs
+++ b/index.bs
@@ -4677,7 +4677,7 @@ Publishing Registries</h4>
 	The [=registry report=] or [=embedded registry=] <em class=rfc2119>must</em>:
 	<ul>
 		<li>
-			Clearly label the [=registry report=]/[=embedded registry=],
+			Clearly label the [=registry report=] or [=embedded registry=],
 			its [=registry tables|tables=],
 			and its [=registry definitions=] as such,
 			including a link to [[#registries]] in this Process.
--- a/index.bs
+++ b/index.bs
@@ -2842,8 +2842,8 @@ Types of Technical Reports</h3>
 		<dd>
 			[=Working Groups=] can also publish [=registries=]
 			in order to document collections of values or other data.
-			These are typically published in a separate [=registry report=],
-			although they can also be directly embedded in [=Recommendation Track=] documents
+			These can either be published in a [=registry report=],
+			or directly embedded in [=Recommendation Track=] documents
 			as an [=embedded registry=].
 			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
--- a/index.bs
+++ b/index.bs
@@ -2844,7 +2844,7 @@ Types of Technical Reports</h3>
 			in order to document collections of values or other data.
 			These are typically published in a separate [=registry report=],
 			although they can also be directly embedded in [=Recommendation Track=] documents
-			as a [=embedded registry=].
+			as an [=embedded registry=].
 			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
 			and can even be done without a [=Working Group=].


index.bs
```diff
@@ -2844,7 +2844,7 @@ Types of Technical Reports</h3>
 			in order to document collections of values or other data.
 			These are typically published in a separate [=registry report=],
 			although they can also be directly embedded in [=Recommendation Track=] documents
-			as a [=registry section=].
+			as a [=embedded registry=].
```
chrisn, line 2847:
```suggestion
			as an [=embedded registry=].
```

index.bs
```diff
@@ -4672,20 +4672,12 @@ Publishing Registries</h4>
 
 	[=Registries=] can be published either
 	as a stand-alone [=technical report=] on the [=Registry Track=] called a <dfn>registry report</dfn>,
-	or incorporated as part of a [=Recommendation=] as a <dfn>registry section</dfn>.
+	or incorporated as part of a [=Recommendation=] as a <dfn oldids="registry-section">embedded registry</dfn>.
```
chrisn, line 4675:
```suggestion
	or incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
```

index.bs
```diff
@@ -4672,20 +4672,12 @@ Publishing Registries</h4>
...
-	For the purposes of the Patent Policy [[PATENT-POLICY]],
-	any [=registry section=] in a [=Recommendation track=] document
-	is not a normative portion of that specification.
-
-	The [=registry report=] or [=registry section=] <em class=rfc2119>must</em>:
+	The [=registry report=] or [=embedded registry=] <em class=rfc2119>must</em>:
 	<ul>
 		<li>
-			Clearly label the [=registry report=]/[=registry section|section=],
+			Clearly label the [=registry report=]/[=embedded registry=],
```
chrisn, line 4680:
```suggestion
			Clearly label the [=registry report=] or [=embedded registry=],
```

chrisn (changes requested):

index.bs
```diff
@@ -4865,6 +4857,16 @@ Specifications that Reference Registries</h4>
...
 
+<h4 id=reg-pat>
+Registries and Patents</h4>
+
+	A [=registry report=] or [=embedded registry=]
+	is not subject to the W3C Patent Policy,
+	and <em class=rfc2119>must not</em> define any requirements on implementations.
+	For the purposes of the Patent Policy [[PATENT-POLICY]] (only),
+	any [=embedded registry=] in a [=Recommendation track=] document
+	is not a normative portion of that specification.
```
nigelmegitt, line 4868: I'm not yet convinced this is correct. I think it is true when applied to the Registry Tables and their entries, but it is not clear to me that the Registry Definition is not normative, since my understanding based on the issue thread (https://github.com/w3c/w3process/issues/800#issuecomment-1802979298 especially) is that Registry Definitions are subject to the same document transition requirements as the reports in which they're contained. So if a Rec track technical report contains a Registry Definition, then it is treated as normative and cannot be changed without going through the generic requirements for changing a normative section of a Rec track document.

If that is correct, then it's only the Registry Tables that are non-normative here.

nigelmegitt (changes requested):
Thank you @fantasai for putting this pull request together. I don't think it matches the state of the issue discussion as it stands, so either the issue thread contains some misunderstandings or the pull request needs a bit of adjustment.

I'd like to see the definition part of §6.5 amended as well, specifically the wording "associated components" which seems ambiguous. It is not clear that the registry tables _are_ the registries, and that they must be hard linked to registry definitions, and can be used by reference or inclusion in a Rec track document.

index.bs
frivoal: > So if a Rec track technical report contains a Registry Definition, then it is treated as normative and cannot be changed without going through the generic requirements for changing a normative section of a Rec track document.

Yes, but to pay attention to the first line of this sentence:
> For the purposes of the Patent Policy [[PATENT-POLICY]] (only),
> […]

As far as Rec track transition criteria and the like are concerned, yes, this is treated like normative text, but that's not what that sentence is talking about.

nigelmegitt: OK, then we need to make that point somewhere else.

frivoal: I'm not sure I understand. What point? Where?

dwsinger: Nigel, do you want to use the phrase "embedded registry content" or "embedded registry tables" as we seem to have continuing confusion between that and "registry definition" (which is normative)?

frivoal: I think it is clear that the content of the registry tables are absolutely not meant to be covered by the patent policy. For the registry definitions, it is more ambiguous. It is normative from the point of view of operating the registry (and from the point of view of what motions you have to go through to amend normative text in a REC), but with the phrasing above, it wouldn't be normative from the point of view of the Patent Policy.

About that, I have two thoughts:
* Cases where this would matter are likely vanishingly rare: you'd have to be in a case where the way you need to do things to modify the content of a registry table is not only patentable (rare already), but also not covered elsewhere. Rules like "must be approved by group X", or "must have a title", or "must not be removed once added"… are not particularly patentable in the first place. Arguably, you could have a registry rule which say something like "the third column of each entry must contain a cryptographic hash of ????, computed according to the following algorithm:[…]". That could get technical enough for this to raise the question of patentability (though it is not obvious that pure algorithms are patentable, but let's not get there today). But even that isn't enough: you'd also have to be in the situation where the registry definition not only calls for that specific method/algo, but also defines it inline, which sounds like a questionable choice. If you want to invoke something of reasonable complexity that should be implementable by multiple people (so that it can be used), you're better of writing an actual standalone spec for that (which would get patent coverage), and then invoking it from your definition.
* It is far from obvious to me that the Patent Policy, as currently written, would be able to cover a registry definition, even if we were to claim that it is normative from the Patent Policy's point of view: the RF licensing requirements cover implementations of a spec, and only covers architectural and interoperability requirements. It is far from clear to me that following the rules of a registry in order to maintain part of a specification would count as implementing the specification, nor that those rules would be considered to be architectural and interoperability requirements.

So all in all, I doubt that the Patent Policy can apply to Registry Definitions, nor that there is much value in trying to make it so. Which is why I am comfortable with the Process stating that not only are the Registry Tables not covered by the Patent Policy, but that neither are the Definitions.

fantasai: > So if a Rec track technical report contains a Registry Definition, then it is treated as normative and cannot be changed without going through the generic requirements for changing a normative section of a Rec track document.

Correct. This is covered in https://www.w3.org/2023/Process-20231103/#reg-pub

> If that is correct, then it's only the Registry Tables that are non-normative here.

No, the registry tables are also normative. It's just that changes to them are [Class 5 changes](https://www.w3.org/2023/Process-20231103/#correction-classes), and are allowed according to the rules of the Registry Definition.

The paragraph quoted here is specifically "for the purposes of the patent policy"; it doesn't apply to any other purpose.

dwsinger: I disagree with Florian above. The Registry Definition is part of a Recommendation, and Recommendation rules apply. We should not complicate our lives by writing unneeded exceptions. The Registry Tables are not, but the definitions are ordinary Rec text.

frivoal: @dwsinger I wouldn't object to doing it the way you say, but I don't think it's going to make a meaningful difference. As I said, I am not convinced the Patent Policy can even successfully apply to Registry Definitions, even if we tried, and even if it did apply, I don't think that's going to turn out to be relevant or useful. But I certainly there would be any harm if it did.

I think it doesn't really make a difference, so we should go with what's easiest to describe and understand. My take is that "Patent Policy doesn't apply to Registries" without using any qualifiers is easiest. But that's subjective, and I don't care too strongly.

dwsinger: I agree with going with the easiest, and I agree that it's hard to see how it could apply. But surely the easiest is to say that a registry definition must be in a Recommendation, and stop. Then all the usual rules apply that apply to Recs. Adding "but the Patent Policy doesn't apply" is both a wrinkle and will likely result in people asking "why the specific carve-out?"

frivoal: > Adding "but the Patent Policy doesn't apply" is both a wrinkle and will likely result in people asking "why the specific carve-out?"

Whether we say that about the whole registry (Definitions and Tables), or only about the Tables, we have to say it anyway. Which is why I think it is easiest to make that sentence apply to Registries altogether, rather than single out a specific part of Registries (Tables). But again, not a hill I am particularly insistent about defending :)

frivoal: (Also, for better or worse, that is what the current text says. The diff makes it look like this is new text, but it is merely moved text, aside from the renaming from [=registry section=] to [=embedded registry=])

nigelmegitt: > Nigel, do you want to use the phrase "embedded registry content" or "embedded registry tables" as we seem to have continuing confusion between that and "registry definition" (which is normative)?

There shouldn't be any difference between the rules for registry tables whether they're embedded or in a registry report, so I'm not sure why we'd have this term at all.

nigelmegitt: > I'm not sure I understand. What point? Where?

Sorry, I meant the point that has been discussed in this thread, about which of Registry Definition and Registry Table are a) normative and b) subject to which kind of change management.

frivoal: > There shouldn't be any difference between the rules for registry tables whether they're embedded or in a registry report, so I'm not sure why we'd have this term at all.

It's not particularly deep: registry in a standalone document need a name, so that we can talk about that sort of documents. When discussing publications of such things for example, it's handy to have a name for them. So we called those "Registry Reports", because they're technical reports that contains registries (and nothing else). It also seemed convenient to have a word to talk about a registry which is not in a standalone registry report, but is inside a Recommendation. So we used to call these registry sections, but as you pointed out, there's a singular/plural issue with that term, so we're proposing to call those embedded registries.

In most respects, it doesn't matter if a registry is in a registry report, or is an embedded registry inside a REC, but the publication process is slightly different: RECs, for the sake of their non-registry parts, go through a PR stage, for instance, which is unnecessary for Registry reports.

frivoal: > I meant the point that has been discussed in this thread, about which of Registry Definition and Registry Table are a) normative and b) subject to which kind of change management.

The word "normative" is used in the Process a number of times, it is used in the ordinary sense of the word, and not as a specific defined term within the Process, with clear consequence for being or not being normative. And I don't mean about registries specifically, I mean the Process in general.

As for Registries, I think the Process covers all the relevant aspects already, but I don't think the answer is just "yes it's normative" or "no it's not", because there are multiple facets to that, and it depends which one you are considering.

For instance, from the point of view of being requirements about an implementation of the specification, the Registry Definition is not normative, because it doesn't discuss the implementation in the first place, it discusses the registry tables. (The Process does say “must not contain any requirements on implementations“.)


But it does define binding behavior on people who will maintain the registry, and will use RFC2119 terms for doing that, so in that sense, it is normative.

Registry tables document values, not requirements, as discussed in https://www.w3.org/2023/Process-20231103/#reg-ref-specifications. So they are not a place to use RFC2119, and in that sense they are not normative, since they do not dictate anything. (The Process does say “must not contain any requirements on implementations“.) But they are (when sufficiently mature) authoritative, within the scope defined by the registry definitions.

But from the point of view of the Recommendation track publication process, and of the Registry Track publication process, the usual [classes of changes 1 through 4](https://www.w3.org/2023/Process-20231103/#correction-classes) apply as always, with the first 2 being editorial, and the next 2 being substantive. No difference here between ordinary REC text and Registry Definitions. Changes to Registry tables are defined there to be [their own class](https://www.w3.org/2023/Process-20231103/#class-5).

From the point of view of the Patent Policy, [normative](https://www.w3.org/Consortium/Patent-Policy-20200915/#def-essential-requirements) means:
> For purposes of this definition, the normative portions of the Specification shall be deemed to include only architectural and interoperability requirements.

Registry tables do not contain requirements, so they cannot be normative *for the purposes of the patent policy*, and we call this out. (Plus, we need registry tables not to be covered by the Patent Policy, otherwise we couldn't do the rapid revisions registries are designed for).

The current process goes beyond that, and says that Registry Reports and Registry sections in their entirety are not subject to the patent policy, including Registry Definitions. For Registry Reports, this allows a simplified publication cycle, and for embedded registies, I think it does not make much of a difference (see https://github.com/w3c/w3process/pull/807#discussion_r1424755999), but for now it matters not where they are defined.

And I think that covers it. Is there some other aspect of "being normative" that you'd like an clarification about?

index.bs
```diff
@@ -2842,9 +2842,9 @@ Types of Technical Reports</h3>
 		<dd>
 			[=Working Groups=] can also publish [=registries=]
 			in order to document collections of values or other data.
-			These are typically published in a separate [=registry report=],
-			although they can also be directly embedded in [=Recommendation Track=] documents
-			as a [=registry section=].
+			These can either be published in a [=registry report=],
+			or directly embedded in [=Recommendation Track=] documents
+			as an [=embedded registry=].
```
TallTed, line 2847:
```suggestion
			A registry can be published either as a distinct [=registry report=],
			or directly within [=Recommendation Track=] a document
			as an [=embedded registry=].
```

index.bs
```diff
@@ -2842,9 +2842,9 @@ Types of Technical Reports</h3>
 		<dd>
 			[=Working Groups=] can also publish [=registries=]
 			in order to document collections of values or other data.
-			These are typically published in a separate [=registry report=],
-			although they can also be directly embedded in [=Recommendation Track=] documents
-			as a [=registry section=].
+			A registry can be published either as a distinct [=registry report=],
+			or directly within [=Recommendation Track=] a document
```
TallTed, line 2846: whoops... 
```suggestion
			or directly within a [=Recommendation Track=] document
```

frivoal: Thanks. Fixed in 5908467

Issue #735, opened 2023-04-24
Requirement to publish formal objections should include a timeline


jyasskin:
https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#registering-objections says 

> A record of each [Formal Objection](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#FormalObjection) against a decision regarding a publicly-available document must be made [publicly available](https://www.w3.org/Consortium/Process/Drafts/snapshots/2023-04#confidentiality-change)

It would be good to describe _when_ the record must be made public. This ambiguity caused a problem with the [formal objections to the DID recommendation](https://www.w3.org/2022/06/DIDRecommendationDecision.html), when @tantek posted Mozilla's objection publicly, while we left Google's responses only member-visible. The Team decided to make the record public much later than I'd expected, which meant that @tantek and Mozilla took more of the blame than they should have.

This doesn't need to be fixed for Process 2023.

fantasai:
Proposed some wording at https://github.com/w3c/w3process/pull/808 ; basically adding “Upon registration with the Team” to the publication requirement.

fantasai:
(Note to the Team, it's not required to be announced promptly, just available.)

PR #808, opened 2023-12-12
Clarify FO publication timing #735

fantasai:
***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/808.html" title="Last updated on Dec 14, 2023, 5:08 AM UTC (533cdb8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/808/dcd212d...fantasai:533cdb8.html" title="Last updated on Dec 14, 2023, 5:08 AM UTC (533cdb8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2287,6 +2287,9 @@ Registering Formal Objections</h3>
 	A Call for Review to the Advisory Committee
 	<em class="rfc2119">must</em> identify any [=Formal Objections=]
 	related to that review.
+	This requirement is waived
+	if the [=Formal Objection=] is resolved to the satisfaction of the objector
+	before its <a href="#confidentiality-change">confidentiality is changed</a>.
 
 	Note: [=Formal Objections=] against matter in a [=technical report=]
 	are required to be addressed before <a href="#rec-advance">requesting advancement</a>
--- a/index.bs
+++ b/index.bs
@@ -2279,7 +2279,8 @@ Registering Formal Objections</h3>
 	rationales,
 	and decisions <em class="rfc2119">should</em> also be recorded.
 
-	A record of each [=Formal Objection=] against a decision regarding a publicly-available document
+	Upon registration with the [=Team=],
+	a record of each [=Formal Objection=] against a decision regarding a publicly-available document
 	<em class="rfc2119">must</em> be made <a href="#confidentiality-change">publicly available</a>;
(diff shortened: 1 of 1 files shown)


cwilso:
This seems like it would not allow the Team to attempt to resolve Formal Objections prior to making them public, which might lead to more press cycles than necessary; it also seems like the FO would be made public immediately upon the vote being filed ("upon registration"), disallowing changing the vote later.

index.bs
```diff
@@ -2279,7 +2279,8 @@ Registering Formal Objections</h3>
 	rationales,
 	and decisions <em class="rfc2119">should</em> also be recorded.
 
-	A record of each [=Formal Objection=] against a decision regarding a publicly-available document
+	Upon registration with the [=Team=],
```
jyasskin, line 2282: I guess this implies that registration needs to include the exact wording that'll become public, so the Team doesn't need to spend time going through [7.3. Changing Confidentiality Level](https://www.w3.org/2023/Process-20231103/#confidentiality-change). That's fine with me, but a delay of up to, say, a week, would also be fine with me.

jyasskin (approved):

jyasskin:
@cwilso I almost made your "upon the vote being filed" comment, but then I noticed
> A [Formal Objection](https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/808.html#FormalObjection) filed during an [Advisory Committee Review](https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/808.html#advisory-committee-review) is considered registered at the close of the review period.

at the bottom of the section.

index.bs
fantasai: The intention was to give some time to the Team to do that, so if this seems a bit too instantaneous, we should pick some other wording here. :)

frivoal: Based on the 2023-12-13 Process CG call:
```suggestion
	No later than when the <a href="#team-fo-mediation">relevant council is initiated<a>,
```

frivoal: (More precise guidance will be written up as a /Guide article, with an Action on @plehegar to draft that.)

jyasskin (approved):
I'm glad to see that the Guide will encourage the Team to publish objections somewhat faster than this requires. This amount of detail in the Process is enough for me. Thanks!

tantek (approved):

PR #818, opened 2024-01-25
Clarify what a registry is made of

frivoal:
The specs that use the registry are necessarily for a registry to be useful, but they're not part of it.

See #800

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4569,11 +4569,10 @@ The Registry Track</h3>
 	consisting of one or more associated <dfn export lt="registry table">registry tables</dfn>,
 	each table representing an updatable collection
 	of logically independent, consistently-structured <dfn export lt="registry entry">registry entries</dfn>.
-	A registry has three associated components:
+	It consists of:
 	<ul>
 		<li>the [=registry definition=], defining how the [=registry tables=] are structured and maintained
 		<li>one or more [=registry tables=], holding the data set represented by the [=registry=] (the <dfn>registry data</dfn>)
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
Looks good to me, thank you.

I did wonder if it is worth retaining _something_ about referencing specifications, but I couldn't think of anything useful that isn't generically true for all technical reports.

index.bs
```diff
@@ -4569,11 +4569,10 @@ The Registry Track</h3>
 	consisting of one or more associated <dfn export lt="registry table">registry tables</dfn>,
```
TallTed, line 4569: "associated" is unnecessary here and may cause confusion.
```suggestion
	consisting of one or more <dfn export lt="registry table">registry tables</dfn>,
```

index.bs
```diff
@@ -4569,11 +4569,10 @@ The Registry Track</h3>
...
 	each table representing an updatable collection
 	of logically independent, consistently-structured <dfn export lt="registry entry">registry entries</dfn>.
-	A registry has three associated components:
+	It consists of:
```
TallTed, line 4572: The antecedent is unclear. Better to use the noun than the pronoun, here.
```suggestion
	A registry consists of:
```

frivoal: I wondered about that. I thought it might have been clear enough, but I was unsure. Going back to being explicit.

nigelmegitt (approved):

TallTed (approved):

tantek (approved):
Nice simplification

plehegar (approved):

Issue #836, opened 2024-03-26
TAG appointment ambiguity about ratification by both AB and TAG


frivoal:
https://www.w3.org/Consortium/Process/Drafts/#TAG-appointments has this sentence:

> The [Team](https://www.w3.org/Consortium/Process/Drafts/#team)'s choice of appointee(s) is subject to ratification by secret ballot by both the [AB](https://www.w3.org/Consortium/Process/Drafts/#advisory-board) and the [TAG](https://www.w3.org/Consortium/Process/Drafts/#technical-architecture-group), each requiring a two-thirds approval.

There was some confusions as to what "each" means, the two possible alternative being:
* each appointee must reach 2/3 approval (of the TAG+AB counted as one body)
* the TAG and AB must each separately reach a 2/3 approval threshold for the appointment to be ratified.

Based on the issue and commit that introduced this (https://github.com/w3c/w3process/issues/715 and https://github.com/w3c/w3process/commit/bcc08778105583c1cd8302db14e4af107b51d7c4), I think it is clear that the second meaning is intended.

We should clarify that to avoid confusion.

martinthomson:
> The [Team](https://www.w3.org/Consortium/Process/Drafts/#team)'s choice of appointee(s) is subject to ratification by secret ballot by both the [AB](https://www.w3.org/Consortium/Process/Drafts/#advisory-board) and the [TAG](https://www.w3.org/Consortium/Process/Drafts/#technical-architecture-group), each requiring a two-thirds approval.

The word "each" here implies distribution of the statement following to multiple items in the preceding text.

I see four interpretations:

1. The approval needs to reach two thirds of the *ballot* for each *appointee*.
2. The approval needs to reach two thirds of the *ballot* for each *ratifying body*.
3. The approval needs to reach two thirds of the *sitting members* for each *appointee*.
4. The approval needs to reach two thirds of the *sitting members* for each *ratifying body*.

This issue covers one point of distribution here.  There, the question is whether the "each" applies to the bodies (TAG and AB) rather than the appointees.  I think that points to option 4, but this could be clearer still.

frivoal:
For the "of the ballot vs of the ratifying body" question, see also https://github.com/w3c/w3process/issues/838

TallTed:
Oh dear. There are additional interpretations. I don't think the intent has ever been for such approval votes as these to cover a *roster* of appointees (as interpretations 2 and 4 above imply), but rather to be held for *each* appointee (no matter their number).

[...]

PR #837, opened 2024-03-26
Disambiguate what "each" means in TAG appointment ratification

frivoal:
Addresses #836

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/837.html" title="Last updated on Mar 28, 2024, 9:12 AM UTC (57607a1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/w3process/837/77ccd12...frivoal:57607a1.html" title="Last updated on Mar 28, 2024, 9:12 AM UTC (57607a1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1253,7 +1253,7 @@ Technical Architecture Group Appointments</h5>
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
 	by both the [=AB=] and the [=TAG=],
-	requiring a two-thirds approval from each of these groups counted separately.
+	requiring a two-thirds approval from each of these groups, counted separately.
 	In the case of regularly scheduled elections,
 	the [=TAG=] participants in this ratification are
 	its members for the upcoming term.
--- a/index.bs
+++ b/index.bs
@@ -1253,7 +1253,7 @@ Technical Architecture Group Appointments</h5>
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
 	by both the [=AB=] and the [=TAG=],
-	each requiring a two-thirds approval.
+	requiring a two-thirds approval from each of these groups counted separately.
 	In the case of regularly scheduled elections,
 	the [=TAG=] participants in this ratification are
 	its members for the upcoming term.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1253,7 +1253,7 @@ Technical Architecture Group Appointments</h5>
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
 	by both the [=AB=] and the [=TAG=],
-	each requiring a two-thirds approval.
+	requiring a two-thirds approval from each of these groups counted separately.
```
TallTed, line 1256: `from each of these groups` already covers `counted separately`.
```suggestion
	requiring a two-thirds approval from each of these groups.
```

TallTed (changes requested):

index.bs
cwilso: As there was confusion about this, I think it is better to be explicit, and Florian's original wording is better.

cwilso (approved):

index.bs
martinthomson:
```suggestion
	requiring a two-thirds approval from each of these groups, counted separately.
```

tantek (approved):
This is a good improvement. I support landing this WITHOUT closing #836 as I think @martinthomson identified some additional ambiguities to address there.

index.bs
```diff
@@ -1253,7 +1253,7 @@ Technical Architecture Group Appointments</h5>
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
 	by both the [=AB=] and the [=TAG=],
-	each requiring a two-thirds approval.
+	requiring a two-thirds approval from each of these groups, counted separately.
```
fantasai, line 1256: I think this could be simplified to "from each" without losing the meaning?

TallTed: Simplifying it not quite so far, to `from each of these groups`, [was rejected above](https://github.com/w3c/w3process/pull/837#pullrequestreview-1960870313). I'd be OK with either simplification (though #838 raises more concerns).

TallTed: `from each group.`?

hober (approved):
This clearly addresses an ambiguity in the existing text (which may have other issues).

Issue #864, opened 2024-05-08
Switching tracks **and back**


frivoal:
What happens to a document which switches away from the REC track (e.g., to Note), and then back to the REC track?

Typically, nowadays, you're not really supposed to do that, and instead if you want to discontinue a REC track document, you should use https://www.w3.org/2023/Process-20231103/#abandon-draft and publish a Discontinued Draft without switching tracks. But that provision is new, so there are already a number of documents that have been parked as Notes, some of which could be switching back. There's now a should against switching away from the REC track that, but (1) it's only a should, and (2) this should is pretty new. So we still need to consider the case.

We have:
* https://www.w3.org/2023/Process-20231103/#switching-tracks, which says:
 
    > Given a [Group decision](https://www.w3.org/2023/Process-20231103/#group-decision) to do so, [Working Groups](https://www.w3.org/2023/Process-20231103/#GroupsWG) can republish a [technical report](https://www.w3.org/2023/Process-20231103/#technical-report) on a different track than the one it is on, under the following restrictions:
    > 
    > * A [technical report](https://www.w3.org/2023/Process-20231103/#technical-report) that is or was a [W3C Recommendation](https://www.w3.org/2023/Process-20231103/#RecsW3C), [W3C Statement](https://www.w3.org/2023/Process-20231103/#statement), or [Patent Review Draft](https://www.w3.org/2023/Process-20231103/#patent-review-drafts) cannot switch tracks.
    > * A [technical report](https://www.w3.org/2023/Process-20231103/#technical-report) should not switch away from the [Recommendation Track](https://www.w3.org/2023/Process-20231103/#w3c-recommendation-track) without due consideration of the Patent Policy implications and approval of W3C’s legal counsel if the [Working Group](https://www.w3.org/2023/Process-20231103/#GroupsWG) envisions a likelihood of returning to it later. 
    > 
    > [Technical reports](https://www.w3.org/2023/Process-20231103/#technical-report) that switch tracks start at their new track’s initial maturity stage, while retaining any established identity (url, shortname, etc.).

* and https://www.w3.org/2023/Process-20231103/#maturity-stages, which lists:

    > * Working Draft (WD) […]
    > * Candidate Recommendation (CR) […]
    > * Proposed Recommendation (PR) […]
    > * W3C Recommendation (REC) […]

[...]

PR #877, opened 2024-05-21
Add note reminding that the first stage of REC track is WD

frivoal:
See #864

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4848,6 +4848,13 @@ Switching Tracks</h3>
 	their new track’s initial maturity stage,
 	while retaining any established identity (url, shortname, etc.).
 
+	Note: The initial maturity stage of the [=Recommendation track=] is [=Working Draft=].
+	[=First Public Working Draft=] designates a specific type of [=Working Draft=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4848,6 +4848,13 @@ Switching Tracks</h3>
 	their new track’s initial maturity stage,
 	while retaining any established identity (url, shortname, etc.).
 
+	Note: The initial maturity stage of the [=Recommendation track=] is [=Working Draft=].
+	[=First Public Working Draft=] designates a specific type of [=Working Draft=]
+	and is not a separate maturity stage.
+	A document which switches to the [=Recommendation track=] is only published as a [=First Public Working Draft=]
+	if it was never previously published as such,
+	otherwise it is simply a [=Working Draft=].
```
TallTed, line 4856: Sorry for the timing; I was on holiday.
```suggestion
	if it was never previously published as such;
	otherwise, it is simply a [=Working Draft=].
```

frivoal: Thanks. Suggestion adopted.

PR #883, opened 2024-06-17
Rationalize/streamline/modernize some core terminology

sideshowbarker:
- Technical Report → W3C Publication
- Recommendation Track → Standards Track

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/process/pull/883.html" title="Last updated on Jun 23, 2024, 2:24 AM UTC (ee38ec4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/883/538646f...ee38ec4.html" title="Last updated on Jun 23, 2024, 2:24 AM UTC (ee38ec4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -165,7 +165,7 @@ Introduction</h2>
 	by encouraging <a href="#Consensus">consensus</a>,
 	soliciting reviews (by both Members and public),
 	incorporating implementation and interoperability experience,
-	and requiring Membership-wide approval as part of the <a href="#Reports">technical report development process</a>.
+	and requiring Membership-wide approval as part of the <a href="#publications">W3C Publication development process</a>.
 	[[#group-participation|Participants]] in W3C include
 	<a href="#member-rep">representatives of its Members</a> and the [=Team=],
 	as well as <a href="#invited-expert-wg">Invited Experts</a>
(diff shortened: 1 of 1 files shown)


frivoal:
@sideshowbarker, to facilitate discussions, could you explain the rationale(s) behind this proposal?

sideshowbarker:
> to facilitate discussions, could you explain the rationale(s) behind this proposal?

Sure. Here it is:

Last week, I had the fun experience of once again being on a WG call (and at 1am my time, which is also usually the case with these things…) attempting to explain CR transition to some members and chairs of a working group who had actually read the parts of Process document about transitions.

And for the Nth time, I found myself needing to answer questions from WG members about things like:

- Is a “CR” some different stage of publication than the very-similar- sounding “TR” in https://www.w3.org/TR/ and that they’ve heard me make reference to when I talk to them about “TR space”?

- What is a “Technical Report”? How is it different from the Working Draft publications the group has already done, or the Candidate Recommendation publication that the group is planning to do?

- At what point is a W3C publication actually considered a standard? If publications follow the “Recommendation Track”, than must a publication reach Recommendation status in order become a standard?

- How can they publish a “living standard” or “evergreen standard” as a CR if the W3C Process says standards are published on the “Recommendation Track”?

It was extremely frustrating to see the confusion and consternation on their faces — and for this to be the Nth time in 17 years where I’ve found myself needing to do this — while I struggled to explain:

- Why the Process doc doesn’t just use “publications” or something to refer to what it currently calls “Technical Reports” — since nobody, us included, ever otherwise calls our publications “Technical Reports”.

- Why the Process doc seems to go so far out of its way to avoid just using the term “standard” — since we otherwise talk so much about “standardization” and the “standardization process”, etc.

For better or worse, I ended up telling that they might be better of not trying to read the Process document itself — since, as currently written, it seems to just be causing them to be more confused rather than less confused.

But I feel pretty foolish telling people not to read the Process doc, and would very much prefer to instead be able to tell them they can safely read it without it confusing them even more than might already be confused.

So the “Technical Report → W3C Publication” change and the “Recommendation Track → Standards Track” suggested in the patch in this PR are mitigations that could help eliminate a couple of key sources for the repeated confusion I’ve seen, and significantly alleviate some of the repeated confusion.

preechapon-cmd (approved):

Issue #886, opened 2024-06-18
Adjust AC appeal vote threshold based on participation


frivoal:
(Split from https://github.com/w3c/process/issues/882).

To avoid making momentous decisions based on a slim majority in vote with low participation, the W3C's [bylaws have provisions](https://github.com/w3c/board/blob/main/governing-docs/bylaws.md#section-11-action-by-members) for what they call "Requisite Member Vote" where decisions are confirmed by : 
* a 75% majority when participation is at or below 15%
* a 2/3rds majority when participation is between 15% and 20%
* a 50% majority when participation is at or above 20%

I think it would be reasonable to adopt the same concept for [AC appeals](https://www.w3.org/policies/process/#ACAppeal), which currently can overturn decisions via a 50% majority without quorum or any other consideration for the level of participation.

frivoal:
This would look something like this: https://github.com/w3c/process/pull/888/commits/7bc50584aac354e3a8a7112eadcbc4f4408e2c9d

frivoal:
From the AB:
> RESOLUTION: Adopt supermajority requirements for low-quorum AC Appeals using same ramp as Bylaws; add minimum quorum of 5%.

martinthomson:
I disagree with this decision.  It's needlessly complicated and it results in non-linearity that encourages odd voting patterns.

martinthomson:
Having looked at the bylaws more closely, I want to revise my statement above.

Why would this not instead *refer* to the bylaws.  The change in #901 sets up different thresholds, where it seems like the rationale was to have the rules be congruent.  A reference to the "Requisite Member Vote" rules in the bylaws (which fail to set a minimum threshold, so a single member could affirm a decision) would also benefit from updates to that definition.

frivoal:
This issue has been closed on the basis of an AB resolution. The Process CG follows AB instructions when they exist (and refers to the AB for resolution of any non-consensual situation), so if you would like this question to be revisited, you need to ask the AB. https://github.com/w3c/AB-memberonly/issues would be the right place.

With that said, here are some further thoughts about the substance of your comment:

> it results in non-linearity that encourages odd voting patterns.

I agree it is non-linear / discontinuous. (I too would have preferred a continuous function, but mathematical elegance was not he main criterion followed by the laywers who drafted this concept for the bylaws.)

[...]

Issue #882, opened 2024-06-11
Recall procedures for TAG and AB


mnot:
The Board has a [removal procedure](https://github.com/w3c/board/blob/main/governing-docs/bylaws.md#section-5-resignation-and-removal), but the TAG and AB do not, beyond the CEO's ability to [discipline them](https://www.w3.org/2023/Process-20231103/#discipline).

Given the direct process role that they now play when Councils are convened, it seems appropriate to make the TAG and AB more accountable to the Membership by defining a procedure for recalling them.

Straw-man proposal: a recall can be instigated by a demand signed by 20 Member representatives, and it's executed as a [requisite member vote](https://github.com/w3c/board/blob/main/governing-docs/bylaws.md#section-11-action-by-members).

frivoal:
Only commenting on a side aspect of the proposal for now, rather than on the core question, but invoking Bylaws concepts like _Member representatives_ and _requisite member votes_ for groups like the AB or TAG which are anchored in the Process is weird, and creates inter-dependencies that I'd rather avoid. If anything, we should go with AC Reps.

mnot:
Yes, that bugged me too, but those procedures / definitions seemed more appropriate. Maybe I've just had my head in the bylaws too much lately.

In the longer term, rationalising the bylaws and process should make such things easier (I hope). In the meantime, let's get agreement in principle on the parameters first, and worry about implementation language later.

dwsinger:
one place to look for similarity or grounding might be the appeal procedure in the Process?

frivoal:
> one place to look for similarity or grounding might be the appeal procedure in the Process?

Yes, I was thinking about that too: https://www.w3.org/2023/Process-20231103/#ACAppeal seems like a workable basis. I'm happy to try and draft something based on that. TLDR: it's a two step process, where you need to convince 5% of the voter base that the question is worth taking up to have a vote, and if you do, you hold a vote during which you need to reach majority for the proposal to pass.

Incidentally, both for this recall procedure and for the general AC appeal, I think we could/should borrow the Bylaws's notion of a threshold that varies based on quorum. passing drastic things by simple majority without a quorum requirement worries e a little bit, even if there's not precedent of it being an issue. Maybe I should open a separate issue about that?

frivoal:
See https://github.com/w3c/process/issues/886 for a discussion of adopting voting thresholds along the lines of what the bylaws do for requisite member votes.

chaals:
[...]

PR #888, opened 2024-06-18
Generalizing AC Appeals and using this procedure for recall.

frivoal:
This PR is a first draft attempting to address <del>https://github.com/w3c/process/issues/886 and </del>https://github.com/w3c/process/issues/882. <del>Neither have</del><ins>It has not</ins> been resolved on at this point, but this shows what adopting <del>them</del><ins>it</ins> could look like.

It can be reviewed as a whole, or commit by commit<del>, to distinguish the effects of #886 from those of #882</del>.

**update: #886 has been handled separately, removing discussion of it from this pull request**.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/888.html" title="Last updated on Sep 25, 2025, 6:52 AM UTC (fa87daa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/888/d3d9ab2...frivoal:fa87daa.html" title="Last updated on Sep 25, 2025, 6:52 AM UTC (fa87daa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1350,10 +1350,12 @@ Removing AB or TAG members</h5>
 AC Recall of AB or TAG</h5>
 
 	The [=Advisory Committee=] <em class=rfc2119>may</em>
-	hold a <dfn export>vote of no confidence</dfn>
-	in the [=Advisory Board=] or the [=Technical Architecture Group=].
+	<dfn export>recall</dfn>,
+	as a whole,
+	all [[#AB-TAG-elections|elected]] and [[#TAG-appointments|appointed]] participants
+	from the [=Advisory Board=] or the [=Technical Architecture Group=].
 
-	An [=Advisory Committee representative=] initiates a [=vote of no confidence=]
+	An [=Advisory Committee representative=] initiates a [=recall=]
 	by sending a request to the Team, and <em class=rfc2119>should</em> also share this request with the Advisory Committee.
 	The request <em class=rfc2119>must</em> identify whether the [=AB=] or [=TAG=] is targeted,
 	and <em class=rfc2119>should</em> also include the rationale.
@@ -1379,10 +1381,7 @@ Elected Groups Vacated Seats</h5>
 			the participant resigns, or
 
 		<li>
-			the participant is [=removed=], oor
-
-		<li>
-			a [=vote of no confidence=] in the body they are part of is successful, or
+			the participant is [=removed=] or [=recalled=], oor
 
 		<li>
 			an Advisory Board or TAG participant changes affiliations
@@ -3034,7 +3033,7 @@ Advisory Committee Override</h3>
 	process nor its [[#addressing-fo|escalation path]] is sufficient.
 
 	Note: Currently, this only applies to [=AC Appeals=]
-	and [=votes of no confidence=] in the [=AB=] or [=TAG=].
+	and [=recalls=] of the [=AB=] or [=TAG=].
 
 	First, the [=Team=] <em class="rfc2119">must</em> announce the proposal for an [=Advisory Committee Override=]
 	to the [=Advisory Committee=]
--- a/index.bs
+++ b/index.bs
@@ -1355,7 +1355,7 @@ AC Recall of AB or TAG</h5>
 	all [[#AB-TAG-elections|elected]] and [[#TAG-appointments|appointed]] participants
 	from the [=Advisory Board=] or the [=Technical Architecture Group=].
 
-	An [=Advisory Committee representative=] initiates a [=recall=]
+	An [=Advisory Committee representative=] invokes a [=recall=]
 	by sending a request to the Team, and <em class=rfc2119>should</em> also share this request with the Advisory Committee.
 	The request <em class=rfc2119>must</em> identify whether the [=AB=] or [=TAG=] is targeted,
 	and <em class=rfc2119>should</em> also include the rationale.
@@ -1367,8 +1367,8 @@ AC Recall of AB or TAG</h5>
 	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
 	if the [=Advisory Committee Override=] proposal passes,
 	all [[#AB-TAG-elections|elected]] or [[#TAG-appointments|appointed]] seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
-	if it fails, it cannot be invoked on the same body
-	sooner than six months since the previous invocation.
+	if it fails, [=recall=] cannot be invoked on the same body
+	sooner than six months since the previous invocatoin.
 	The [=Team=] must announce the conclusion of the [=Advisory Committee Override=] to the [=Advisory Committee=].
 
 <h5 id="AB-TAG-vacated">
--- a/index.bs
+++ b/index.bs
@@ -1346,6 +1346,29 @@ Removing AB or TAG members</h5>
 
 	Note: Removal from the [=AB=] or [=TAG=] does not imply removal from a convened [=Council=].
 
+<h5 id="AB-TAG-recall">
+AC Recall of AB or TAG</h5>
+
+	The [=Advisory Committee=] <em class=rfc2119>may</em>
+	hold a <dfn export>vote of no confidence</dfn>
+	in the [=Advisory Board=] or the [=Technical Architecture Group=].
+
+	An [=Advisory Committee representative=] initiates a [=vote of no confidence=]
+	by sending a request to the Team, and <em class=rfc2119>should</em> also share this request with the Advisory Committee.
+	The request <em class=rfc2119>must</em> identify whether the [=AB=] or [=TAG=] is targeted,
+	and <em class=rfc2119>should</em> also include the rationale.
+
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
+	on the proposal.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
+	if the [=Advisory Committee Override=] proposal passes,
+	all [[#AB-TAG-elections|elected]] or [[#TAG-appointments|appointed]] seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
+	if it fails, it cannot be invoked on the same body
+	sooner than six months since the previous invocation.
+	The [=Team=] must announce the conclusion of the [=Advisory Committee Override=] to the [=Advisory Committee=].
+
 <h5 id="AB-TAG-vacated">
 Elected Groups Vacated Seats</h5>
 
@@ -1356,7 +1379,10 @@ Elected Groups Vacated Seats</h5>
 			the participant resigns, or
 
 		<li>
-			the participant is [=removed=], or
+			the participant is [=removed=], oor
+
+		<li>
+			a [=vote of no confidence=] in the body they are part of is successful, or
 
 		<li>
 			an Advisory Board or TAG participant changes affiliations
@@ -1386,7 +1412,7 @@ Elected Groups Vacated Seats</h5>
 		<li>
 			When an elected seat on either the [=AB=] or [=TAG=] is vacated,
 			the seat is filled at the next regularly scheduled election for the group
-			unless the group Chair requests that W3C hold an election before then
+			unless the group Chair requests that W3C hold a [=special election=] before then
 			(for instance, due to the group's workload).
 
 			<ul>
@@ -1400,15 +1426,18 @@ Elected Groups Vacated Seats</h5>
 					including a resignation effective as of a given date in the future.
 			</ul>
 
-			When such an election is held,
-			the minimum number of available seats is such that
-			when added to the number of continuing participants,
-			the minimum total number of elected seats is met
-			(8 for the [=TAG=], 9 for the [=AB=]);
-			and the maximum number corresponds to all unoccupied seats.
-			Except for the number of available seats and the length of the terms,
-			the <a href="#AB-TAG-elections">usual rules for Advisory Board and Technical Architecture Group Elections</a> apply.
+		<li>
+			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
+			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
+			unless the next regularly scheduled election is fewer than three months away.
 	</ul>
+
+	When a <dfn>special election</dfn> is held,
+	vacated seats are filled for the remainder of the term of the vacancy,
+	and the (minimum, maximum) number of available seats is
+	the number of seats required to meet
+	the (minimum, maximum) total defined for the group.
+	The <a href="#AB-TAG-elections">usual rules for Advisory Board and Technical Architecture Group Elections</a> otherwise apply.
 
 <h3 id="GAGeneral" oldids="ChapterGroups, WG-and-IG">
 Chartered Groups: Working Groups and Interest Groups</h3>
@@ -3004,7 +3033,8 @@ Advisory Committee Override</h3>
 	used to resolve certain matters where neither the usual [=consensus=]
 	process nor its [[#addressing-fo|escalation path]] is sufficient.
 
-	Note: Currently, this only applies to [=AC Appeals=].
+	Note: Currently, this only applies to [=AC Appeals=]
+	and [=votes of no confidence=] in the [=AB=] or [=TAG=].
 
 	First, the [=Team=] <em class="rfc2119">must</em> announce the proposal for an [=Advisory Committee Override=]
 	to the [=Advisory Committee=]
--- a/index.bs
+++ b/index.bs
@@ -1796,7 +1796,7 @@ Initiating Charter Refinement</h3>
 	and can be appealed only by 5 or more [=Members=],
 	through their [=Advisory Committee representative=],
 	formally objecting to the decision within 8 weeks of the decision being announced.
-	In this case the [=Team=] <em class=rfc2119>must</em> start an [=appeal vote=]
+	In this case the [=Team=] <em class=rfc2119>must</em> start an [=override vote=]
 	on whether to overturn the [=Team Decision=].
 	(No action is required to be taken when fewer than 5 members object.)
 
@@ -2942,7 +2942,7 @@ Advisory Committee Votes</h3>
 
 	The [=Advisory Committee=] votes in <a href="#AB-TAG-elections">elections for seats on the TAG or Advisory Board</a>,
 	and in the event of an [=Advisory Committee Appeal=]
-	achieving the required support to trigger an [=appeal vote=].
+	achieving the required support to trigger an [=override vote=].
 	Whenever the [=Advisory Committee=] votes,
 	each Member or group of [=related Members=] has one vote.
 
@@ -2976,24 +2976,49 @@ Appeal by Advisory Committee Representatives</h3>
 
 	An [=Advisory Committee representative=] initiates an [=appeal=] by sending a request to the [=Team=],
 	and should also share this request with the [=Advisory Committee=].
-	The request should say “I appeal this Decision”
+	The request <em class=rfc2119>should</em> say “I appeal this Decision”
 	and identify the decision,
-	and may also include their rationale for appealing the decision.
+	and <em class=rfc2119>may</em> also include their rationale for the [=appeal=].
 
 	Note: See [[DECISION-APPEAL inline]] for a recommendation
 	on how to communicate an appeal request to the [=Team=] and the [=AC=].
 
-	Within one week the [=Team=] <em class="rfc2119">must</em> announce the appeal process
+	Within one week, the [=Team=] <em class="rfc2119">must</em> initiate the [=appeal=]
+	in the form of an [=Advisory Committee Override=]
+	on the proposal to overturn the decision.
+
+	The conclusion of the [=Advisory Committee Override=] is final;
+	the same decision cannot be appealed more than once.
+
+	If the [=Advisory Committee Override=]
+	approves the proposal to overturn the decision,
+	those who had initiated the proposal <em class=rfc2119>may</em> revise it
+	to address the causes of rejection
+	and follow the ordinary applicable process
+	to submit the revised proposal.
+
+<h3 id="ac-override">
+Advisory Committee Override</h3>
+
+	An <dfn export>Advisory Committee Override</dfn> is an exceptional two-step procedure
+	used to resolve certain matters where neither the usual [=consensus=]
+	process nor its [[#addressing-fo|escalation path]] is sufficient.
+
+	Note: Currently, this only applies to [=AC Appeals=].
+
+	First, the [=Team=] <em class="rfc2119">must</em> announce the proposal for an [=Advisory Committee Override=]
 	to the [=Advisory Committee=]
 	and provide a mechanism for [=Advisory Committee representatives=]
-	to respond with a statement of positive support for this appeal.
+	to respond with a statement of positive support for holding a vote.
 	The archive of these statements <em class="rfc2119">must</em> be [=member-only=].
+
 	If, within <span class="time-interval">one week</span> of the Team's announcement,
-	5% or more of the [=Advisory Committee=] support the appeal request,
-	the Team <em class="rfc2119">must</em> organize an <dfn>appeal vote</dfn>
+	5% or more of the [=Advisory Committee=] support holding the vote,
+	the Team <em class="rfc2119">must</em> organize an <dfn>override vote</dfn>
 	asking the [=Advisory Committee=]
-	“Do you approve of the Decision?”
-	together with links to the decision and the appeal support.
+	whether they approve of the Advisory Committee Override proposal,
+	including details of the proposal
+	and links to support for holding the vote.
 
 	The ballot <em class="rfc2119">must</em> allow for three possible responses:
 	“Approve”,
@@ -3006,24 +3031,19 @@ Appeal by Advisory Committee Representatives</h3>
 	(including explicit “abstain” ballots) 
 	by [=Advisory Committee Representatives=]:
 	* if fewer than 5% participate,
-		the vote fails.
+		the proposal is rejected.
 	* if at least 5% but no more than 15% participate,
 		and the number of “Approve” ballots exceeds three times (3x) the number of “Reject” ballots,
-		the vote passes.
+		the proposal is approved.
 	* if more than 15% but fewer than 20% participate,
 		and the number of “Approve” ballots exceeds twice (2x) the number of “Reject” ballots,
-		the vote passes.
+		the proposal is approved.
 	* if 20% or more participate,
 		and the number of “Approve” ballots exceeds the number of “Reject” ballots,
-		the vote passes.
+		the proposal is approved.
 
-	If the vote passes,
-	the decision is overturned.
-	Following such rejection,
-	those who had initiated the proposal may revise it
-	to address the causes of rejection
-	and follow the ordinary applicable process
-	to submit the revised proposal.
+	Otherwise,
+	the proposal is rejected.
 
 <h2 id="Reports">
 W3C Technical Reports</h2>
--- a/index.bs
+++ b/index.bs
@@ -1381,7 +1381,7 @@ Elected Groups Vacated Seats</h5>
 			the participant resigns, or
 
 		<li>
-			the participant is [=removed=] or [=recalled=], oor
+			the participant is [=removed=] or [=recalled=], or
 
 		<li>
 			an Advisory Board or TAG participant changes affiliations
--- a/index.bs
+++ b/index.bs
@@ -1428,7 +1428,7 @@ Elected Groups Vacated Seats</h5>
 		<li>
 			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
 			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
-			unless the close of the next regularly scheduled election is fewer than three months away.
+			unless a regular Call for Nominations is scheduled within 1 month.
 	</ul>
 
 	When a <dfn>special election</dfn> is held,
--- a/index.bs
+++ b/index.bs
@@ -1428,7 +1428,7 @@ Elected Groups Vacated Seats</h5>
 		<li>
 			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
 			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
-			unless the next regularly scheduled election is fewer than three months away.
+			unless the close of the next regularly scheduled election is fewer than three months away.
 	</ul>
 
 	When a <dfn>special election</dfn> is held,


index.bs
```diff
@@ -1299,6 +1299,38 @@ Verifiable Random Selection Procedure</h5>
...
+	The request <em class=rfc2119>should</em> say “I call for removal of (an) AB and/or TAG participant(s)”,
+	<em class=rfc2119>must</em> identify the [=AB=] or [=TAG=] [=participant=] or [=participants=] targeted,
+	and <em class=rfc2119>should</em> also include the rationale for the proposed [=recall=].
+
+	Messages sent by [=Advisory Committee representatives=] to the [=Advisory Committee=]
+	with the ostensible intent of garnering support for a recall that has not yet been formally requested
+	<em class=rfc2119>should</em> also be treated by the [=Team=]
+	as a request to initiate the recall.
+
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate a [=Advisory Committee Override=]
```
TallTed, line 1320:
```suggestion
	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
```

index.bs
```diff
@@ -1299,6 +1299,38 @@ Verifiable Random Selection Procedure</h5>
...
+	on the recall proposal.
+
+	If the [=Advisory Committee Override=] proposal passes,
+	the seat(s) of the relevant [=AB=] or [=TAG=] participant(s) are vacated immediately.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=].
+
+	A single [=Member=] or group of [=related Members=] cannot re-invoke this process
+	sooner than 6-month since their previous invocation,
```
TallTed, line 1330: It's unclear when the six-month timer begins. So, this —
```suggestion
	A single [=Member=] or group of [=related Members=] cannot re-invoke the [=recall=] process
	less than six months following the conclusion of their previous invocation,
```

TallTed: — or this —
```suggestion
	A single [=Member=] or group of [=related Members=] cannot re-invoke the [=recall=] process
	less than six months following the initiation of their previous invocation,
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
 
 	An [=Advisory Committee representative=] initiates an [=appeal=] by sending a request to the [=Team=],
 	and should also share this request with the [=Advisory Committee=].
-	The request should say “I appeal this Decision”
+	The request <em class=rfc2119>should</em> say “I appeal this Decision”
 	and identify the decision,
-	and may also include their rationale for appealing the decision.
+	and <em class=rfc2119>may</em> also include their rationale for appealing the decision.
```
TallTed, line 2798:
```suggestion
	and <em class=rfc2119>may</em> also include their rationale for making the [=appeal=].
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
 
 	Note: See [[DECISION-APPEAL inline]] for a recommendation
 	on how to communicate an appeal request to the [=Team=] and the [=AC=].
 
-	Within one week the [=Team=] <em class="rfc2119">must</em> announce the appeal process
+	Within one week the [=Team=] <em class="rfc2119">must</em> initiate the [=appeal=]
```
TallTed, line 2803:
```suggestion
	Within one week, the [=Team=] <em class="rfc2119">must</em> initiate the [=appeal=]
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	in the form of a [=Advisory Committee Override=]
+	on the proposal to overturn the decision.
+
+	The conclusion of the [=Advisory Committee Override=] is final:
```
TallTed, line 2807:
```suggestion
	The conclusion of the [=Advisory Committee Override=] is final;
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	the same decision cannot be appealed more than once.
+
+	When the [=Advisory Committee Override=]
```
TallTed, line 2810:
```suggestion
	If the [=Advisory Committee Override=]
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	approves the proposal to overturn the decision,
+	the proposal <em class=rfc2119>may</em> be returned for additional work,
+	after which the applicable decision process is re-initiated.
+
+<h3 id="ac-override">
+Advisory Committee Override</h3>
+
+	An <dfn export>Advisory Committee Override</dfn> is an exceptional two-step procedure
+	used to resolve certain matters where the usual [=consensus=] process
+	and its [[#addressing-fo|escalation path]] are insufficient.
```
TallTed, line 2820:
```suggestion
	and its [[#addressing-fo|escalation path]] have been proven to be insufficient.
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
 	to the [=Advisory Committee=]
 	and provide a mechanism for [=Advisory Committee representatives=]
-	to respond with a statement of positive support for this appeal.
+	to respond with a statement of positive support for holding a vote.
 	The archive of these statements <em class="rfc2119">must</em> be [=member-only=].
+
 	If, within <span class="time-interval">one week</span> of the Team's announcement,
-	5% or more of the [=Advisory Committee=] support the appeal request,
-	the Team <em class="rfc2119">must</em> organize an appeal vote
+	5% or more of the [=Advisory Committee=] support the holding the vote,
```
TallTed, line 2832:
```suggestion
	5% or more of the [=Advisory Committee=] support holding the vote,
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	(including those who cast an explicit “abstain” ballot):
+	* if no more than 15%,
+		the vote passes
+		if the number of “Approve” ballots exceeds three times the number of “Reject” ballots.
+	* if more than 15% but less than 20%,
+		the vote passes
+		if the number of “Approve” ballots exceeds twice the number of “Reject” ballots.
+	* if 20% or more,
+		the vote passes
+		if the number of “Approve” ballots exceeds the number of “Reject” ballots.
```
TallTed, line 2856:
```suggestion
	* if 15% or fewer participate,
		and the number of “Approve” ballots exceeds triple (3x) the number of “Reject” ballots,
		the vote passes.
	* if more than 15% but fewer than 20% participate,
		and the number of “Approve” ballots exceeds double (2x) the number of “Reject” ballots,
		the vote passes.
	* if 20% or more participate,
		and the number of “Approve” ballots exceeds the number of “Reject” ballots,
		the vote passes.
```

TallTed (changes requested):

index.bs
TallTed: or
```suggestion
	The conclusion of the [=Advisory Committee Override=] is final, and
```

frivoal: Not sure what's wrong with the colon. The part after the colon explains what it means for the conclusion to be final.

index.bs
frivoal: I guess the suggested works too, but I don't seen what's wrong with the original (this line predates this PR by the way, I merely added an rfc2119 class on "may").

index.bs
frivoal: I meant the second one, but it seems to me that the existing phrasing already is unambiguous (and shorter). Can you help me understand why you think it's not clear?

index.bs
frivoal: I think that phrasing doesn't work well for the AB/TAG recall case. Make me wonder about my phrasing too, to be honest.

index.bs
frivoal: I think I like the rephrasing in terms of the first line of each bullet, but I prefer the original for the rest of the text. Any third opinion?

index.bs
TallTed: > why you think it's not clear

Uh... because it wasn't clear to me, after reading several times. Hence my comment, and alternate suggestions.

It matters, because if the process initiated on 2025-06-01 takes 2025-09-15, re-invocation request might not be accepted until 2025-12-01 *or* until 2026-03-15. It's important to be able to calculate which date applies.

index.bs
TallTed: I'll live with either wording, though I'd suggest an added cross-reference (which might need different markup, to handle `appeal` vs `appealing`)
```suggestion
	and <em class=rfc2119>may</em> also include their rationale for [=appealing=] the decision.
```

index.bs
TallTed: There's nothing *wrong* with the colon; it's just that I find colons often confuse readers. The meaning of the sentence shouldn't be changed by any of the constructions above.

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	approves the proposal to overturn the decision,
+	the proposal <em class=rfc2119>may</em> be returned for additional work,
+	after which the applicable decision process is re-initiated.
+
+<h3 id="ac-override">
+Advisory Committee Override</h3>
+
+	An <dfn export>Advisory Committee Override</dfn> is an exceptional two-step procedure
+	used to resolve certain matters where the usual [=consensus=] process
+	and its [[#addressing-fo|escalation path]] are insufficient.
```
TallTed, line 2820: Maybe?
```suggestion
	used to resolve certain matters where neither the usual [=consensus=]
	process nor its [[#addressing-fo|escalation path]] is sufficient.
```

index.bs
```diff
@@ -2758,44 +2793,69 @@ Appeal by Advisory Committee Representatives</h3>
...
+	(including those who cast an explicit “abstain” ballot):
+	* if no more than 15%,
+		the vote passes
+		if the number of “Approve” ballots exceeds three times the number of “Reject” ballots.
+	* if more than 15% but less than 20%,
+		the vote passes
+		if the number of “Approve” ballots exceeds twice the number of “Reject” ballots.
+	* if 20% or more,
+		the vote passes
+		if the number of “Approve” ballots exceeds the number of “Reject” ballots.
```
TallTed, line 2856: Also possible... and `more than 15% but fewer than 20%` could become `between 15% and 20%` --
```suggestion
	(including those who cast an explicit “abstain” ballot). The vote passes in any of the following cases:
	* if 15% or fewer participate,
	  and the number of “Approve” ballots exceeds triple (3x) the number of “Reject” ballots
	* if more than 15% but fewer than 20% participate,
	  and the number of “Approve” ballots exceeds double (2x) the number of “Reject” ballots
	* if 20% or more participate,
	  and the number of “Approve” ballots exceeds the number of “Reject” ballots
```

index.bs
frivoal: I think the key to me is that "invoking" is a point in time, at the start, not a period, just the same way "initiating is" (with the nuance that invoking is more specific: it's initiating by saying something).

So, your second phrasing matches what I meant, but to me it feels redundant: at the since the start of the beginning of the initiation of the invocation… ;)

index.bs
frivoal: Resolved by https://github.com/w3c/process/pull/888#discussion_r1657569297

index.bs
TallTed: "If, then, if" construction — i.e., having the "if" portions of each bullet broken by the "then" portion — makes little sense to me. I feel strongly that the simpler "if, and, then" construction is clearer. (All "then" are implicit, and I think they do not need to be made explicit.)

frivoal: Ok, I think I agree with this latest point of yours. I'd be in favor of taking in your overall wording, but the one part I a still not convinced about is replacing
> the number of “Approve” ballots exceeds _twice_ the number of “Reject” ballots

with

> the number of “Approve” ballots exceeds _double (2x)_ the number of “Reject” ballots

I think both are correct, but the first one sounds better to my ear. But then again, I'm not a native speaker. Am I wrong here?

dwsinger: 'twice' sounds better to me but I am not sure why. (and 'thrice' would be the matching archaic word)

TallTed: `Twice` and `thrice` are certainly options. But I think both are more likely to be misunderstood by non-native English readers than `double` and `triple`.

I added the 2x and 3x parentheticals to help clarity for those non-native English readers.

A mid-ground might be `twice (2x)` and `thrice (3x)`.

dwsinger: I would not use the word "thrice", it's considered somewhat archaic. Twice (2x) and three times (3x) would be fine by me.

chaals:
An alternative mid-ground would be stating a "supermajority" threshold: 
twice => > 2/3 or 67%, thrice => > 3/4 or 75%

index.bs
TallTed: `Twice (2x)` and `three times (3x)` applied to my change suggestion.

frivoal:
@chaals, I'd rather not phrase it this way, because when you just say "supermajority of 2/3" or some such phrasing, it's ambiguous how you treat abstain ballots. You can make it clear, but that usually make the phrasing longer and clunkier, which is why I think "x times as many ballots for as against" or that sort of phrasing is better.

index.bs
TallTed: I think that your phrasing is ambiguous enough that it could be read as starting the six month timer at conclusion of the previous invocation, which would be certain to prevent overlapping recall process runs, which seems a likely reason for this timer. 

Starting the timer at "invocation time" leaves the potential of overlap, if an early run takes longer than 6 months (which seems possible if unlikely).

It might help to include a reason for this required time lapse?
```suggestion
	A single [=Member=] or group of [=related Members=] cannot re-invoke the [=recall=] process
	less than six months following the start of the process based on their previous invocation,
```

index.bs
```diff
@@ -1318,6 +1350,9 @@ Elected Groups Vacated Seats</h5>
 		<li>
 			the participant resigns, or
 
+		<li>
+			the participant is [=recalled=], or
```
TallTed, line 1354: These "or" list items should be rewritten, with a new lead-in —
`	An [=Advisory Board=] or [=TAG=] participant's seat is vacated when any of the following occurs:`
— and all of the `, or` should be removed from the `<li>` that follow. GitHub won't let me make a suggestion that reaches so far above and below the lines changed by this PR. If need be, I'll submit a PR for this, but I'm hoping this comment will be sufficient for you to adjust #888.

index.bs
```diff
@@ -2841,24 +2876,48 @@ Appeal by Advisory Committee Representatives</h3>
 
 	An [=Advisory Committee representative=] initiates an [=appeal=] by sending a request to the [=Team=],
 	and should also share this request with the [=Advisory Committee=].
-	The request should say “I appeal this Decision”
+	The request <em class=rfc2119>should</em> say “I appeal this Decision”
 	and identify the decision,
-	and may also include their rationale for appealing the decision.
+	and <em class=rfc2119>may</em> also include their rationale for [=appealing=] the decision.
```
TallTed, line 2881:
```suggestion
	and <em class=rfc2119>may</em> also include their rationale for the [=appeal=].
```

index.bs
```diff
@@ -2841,24 +2876,48 @@ Appeal by Advisory Committee Representatives</h3>
...
 
 	Note: See [[DECISION-APPEAL inline]] for a recommendation
 	on how to communicate an appeal request to the [=Team=] and the [=AC=].
 
-	Within one week the [=Team=] <em class="rfc2119">must</em> announce the appeal process
+	Within one week, the [=Team=] <em class="rfc2119">must</em> initiate the [=appeal=]
+	in the form of a [=Advisory Committee Override=]
```
TallTed, line 2887:
```suggestion
	in the form of an [=Advisory Committee Override=]
```

index.bs
```diff
@@ -2841,24 +2876,48 @@ Appeal by Advisory Committee Representatives</h3>
...
+Advisory Committee Override</h3>
+
+	An <dfn export>Advisory Committee Override</dfn> is an exceptional two-step procedure
+	used to resolve certain matters where neither the usual [=consensus=]
+	process nor its [[#addressing-fo|escalation path]] is sufficient.
+
+	Note: Currently, this only applies to [=AC Appeals=]
+	and [=recalls=] of [=AB=] or [=TAG=] participants.
+
+	First, the [=Team=] <em class="rfc2119">must</em> announce the proposal for a [=Advisory Committee Override=]
```
TallTed, line 2908:
```suggestion
	First, the [=Team=] <em class="rfc2119">must</em> announce the proposal for an [=Advisory Committee Override=]
```

index.bs
```diff
@@ -2841,24 +2876,48 @@ Appeal by Advisory Committee Representatives</h3>
...
-	5% or more of the [=Advisory Committee=] support the appeal request,
-	the Team <em class="rfc2119">must</em> organize an appeal vote
+	5% or more of the [=Advisory Committee=] support holding the vote,
+	the Team <em class="rfc2119">must</em> organize a vote
 	asking the [=Advisory Committee=]
-	“Do you approve of the Decision?”
-	together with links to the decision and the appeal support.
+	“Do you approve of the Advisory Committee Override proposal?”
+	together with details about the proposal
+	and links to support for holding the vote.
```
TallTed, line 2920:
```suggestion
	whether they approve of the Advisory Committee Override proposal,
	including details of the proposal
	and links to support for holding the vote.
```

TallTed (changes requested):

index.bs
frivoal: > It might help to include a reason for this required time lapse?

Generally, if you ask for a recall and the motion does not pass, being able to ask again and again has more to do with harassment than with accountability, even if you target different subsets of the elected body every time, so we should try to prevent it. However, if new facts come to light, asking for another recall might be appropriate.

> if an early run takes longer than 6 months (which seems possible if unlikely).

I don't think it is or should be possible:
- first someone invokes the recall
- Within one week, the Team must open a poll to see if there's support
- Withing one week of that, if we don't reach 5%, it's over, and if we do, we launch the full recall vote
- after the vote, the procedure ends, and there's no recourse

It is true that the Process doesn't specify how long the vote should stay open, but a 6 month balloting period is unheard of and would be unreasonable. Maybe we should encode into the Process that the ballot stay open for 28 days, which is a typical duration for that sort of things (e.g., that's what was used for the [EME AC Appeal](https://www.w3.org/2002/09/wbs/33280/eme-appeal-vote/results))

Dlbandy04 (approved):

frivoal:
I've updated this draft PR to reflect my current take on this issue, as expressed in https://github.com/w3c/AB-memberonly/issues/237#issuecomment-2354297741.

frivoal:
The AB has not reached a conclusion on this topic. Temporarily removing agenda+

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
+	A chair of the [=AB=] or [=TAG=] <em class=rfc2119>must</em> hold a hearing
+	on the potential [=removal=] of a participant
+	if requested by at least three of the participants in the group.
+	After giving the individual in question
+	an opportunity to defend themselves,
+	a vote on the proposed [=removal=] is held.
+	If at least three quarters of the participants in the group,
+	excluding the individual who is the subject of such vote,
+	then vote in favor of the proposal,
+	the individual's seat on [=AB=] or [=TAG=] is [=vacated=] immediately.
```
martinthomson, line 1333: I would say that a lower threshold is sufficient.  A simple majority might suffice in this case.

Consider this: if there are three people who are willing to raise this, then that alone is indicative of dysfunction in need of remediation.  Removal should not be the first option there, but if you need it, then there is a chance that some people will be forced to recuse or some will be absent, making it quite difficult to reach 3/4.

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
+	on the proposal.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
+	if the [=Advisory Committee Override=] proposal passes,
+	all seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
+	if it fails, it cannot be invoked on the same body
+	sooner than 6-month since their previous invocation.
+
```
martinthomson, line 1356: I think that in both cases, you want to include some sort of reporting responsibility.  That is, there is an announcement that the seat is vacated.

There is no way to hide what is happened here and you don't want to have people learn via rumor mills.  Information should be publicly communicated to the AC by the chairs or Team as promptly as possible.

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
+Collective removal</h6>
+
+	The [=Advisory Committee=] <em class=rfc2119>may</em>
+	hold a <dfn export>vote of no confidence</dfn>
+	in the [=Advisory Board=] or the [=Technical Architecture Group=].
+
+	An [=Advisory Committee representative=] initiates a [=vote of no confidence=]
+	by sending a request to the Team, and <em class=rfc2119>should</em> also share this request with the Advisory Committee.
+	The request <em class=rfc2119>must</em> identify which of the [=AB=] or [=TAG=] is targeted,
+	and <em class=rfc2119>should</em> also include the rationale.
```
martinthomson, line 1345: I think that you want a three member threshold for this too.  Otherwise, this is open to trolling and DoS.

I'd be OK with a higher threshold than three, but not a lower one.

If that takes the form of one AC member initiating an override that has to be seconded by two other members in all cases, that would be ideal.  I know that this mechanism hasn't been activated, but it's an organizational vulnerability.

index.bs
```diff
@@ -1368,6 +1419,12 @@ Elected Groups Vacated Seats</h5>
 			and the maximum number corresponds to all unoccupied seats.
 			Except for the number of available seats and the length of the terms,
 			the <a href="#AB-TAG-elections">usual rules for Advisory Board and Technical Architecture Group Elections</a> apply.
+
+		<li>
+			If seats are vacated due to a successful [=vote of no confidence=],
```
martinthomson, line 1424: I think that you want to set a threshold for triggering an election, not just have this for the no confidence vote.  The bodies will likely function with a few gaps, but there is a critical point at which the groups are non-viable.  I might suggest 2/3 of the target size as that threshold.

```suggestion
			If vacancies reduce
			the number of [=AB=] participants to 6 or fewer
			or the number of [=TAG=] participants to 7 or fewer,
```

index.bs
```diff
@@ -1368,6 +1419,12 @@ Elected Groups Vacated Seats</h5>
...
+			the [=Team=] <em class=rfc2119>must</em> organize an election,
+			under the same condition as for individually vacated seats,
```
martinthomson, line 1426:
```suggestion
			the [=Team=] <em class=rfc2119>must</em> organize an election.
			Vacated seats are filled for the remainder of the term of the vacancy,
```

index.bs
```diff
@@ -1368,6 +1419,12 @@ Elected Groups Vacated Seats</h5>
...
+			unless the next regularly scheduled election is fewer than three months away.
```
martinthomson, line 1427: I'm not clear on the three month choice.  That means that you might win an election for a 3 month and a day term.  That's pretty pointless, I'd go with six.

```suggestion
			unless the next regularly scheduled election for the vacancy is less than six months away,
			in which case the remainder of the term is extended by two years.
```

index.bs
```diff
@@ -2887,24 +2944,50 @@ Appeal by Advisory Committee Representatives</h3>
...
 	to the [=Advisory Committee=]
 	and provide a mechanism for [=Advisory Committee representatives=]
-	to respond with a statement of positive support for this appeal.
+	to respond with a statement of positive support for holding a vote.
 	The archive of these statements <em class="rfc2119">must</em> be [=member-only=].
+
 	If, within <span class="time-interval">one week</span> of the Team's announcement,
-	5% or more of the [=Advisory Committee=] support the appeal request,
-	the Team <em class="rfc2119">must</em> organize an appeal vote
+	5% or more of the [=Advisory Committee=] support holding the vote,
```
martinthomson, line 2985: This is basically like having a vote to have a vote.  I don't see the point.

It reduces some of my concerns about DoS, engaging this mechanism is exactly the sort of thing I'd like to avoid, but it also makes it nearly infeasible to use this process.  Even though 5% seems small, that's a pretty significant movement given the activity levels in a pretty large consortium.  I suggested one objector and two seconds, but you could make the two into five or something like that.

index.bs
```diff
@@ -2917,24 +3000,19 @@ Appeal by Advisory Committee Representatives</h3>
 	(including explicit “abstain” ballots) 
 	by [=Advisory Committee Representatives=]:
 	* if fewer than 5% participate,
-		the vote fails.
+		the proposal is rejected.
 	* if at least 5% but no more than 15% participate,
 		and the number of “Approve” ballots exceeds three times (3x) the number of “Reject” ballots,
```
martinthomson, line 3005: This seems pretty silly.  I understand the goal, but it encourages tactical voting due to it being non-contiguous.  Opponents can withhold participation to get a better outcome.

martinthomson (commented):
I don't spend a lot of time with this part of the process and there are a few comments here that reflect that.  Feel free to defer those to issues if you would prefer not to engage with them.

index.bs
martinthomson: It might be worth an example.  For simplicity, let's assume 100 members.

At the 3x to 2x threshold (15%), four "reject" votes blocks any number of "approve" votes up to 11, but one more "reject" vote causes the motion to pass.  This is because at 16 participating, five "reject" votes is overridden by 11 "approve" votes.

Worse, prior to the 2x to 1x threshold (20%), a motion is blocked by seven "reject" votes at 13:7 in favor.  Up to six more "reject" votes causes the motion to pass.  If opponents want to have their say and retain the same outcome, they need to find seven more "reject" votes (with no more "approve" votes).

There are ways to address this, but it involves math.  I wonder if this goal is worth that.

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
+
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
+	on the proposal.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
+	if the [=Advisory Committee Override=] proposal passes,
+	all seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
+	if it fails, it cannot be invoked on the same body
+	sooner than 6-month since their previous invocation.
```
TallTed, line 1355:
```suggestion
	sooner than six months since their previous invocation.
```

index.bs
frivoal: I disagree. I think it should take a strong supermajority to eject someone. We want to stay clear of the possibility that someone gets kicked out (or simply fears getting kicked out) simply because half the group disagrees with them (and lacks restraint).

Making an allowance for people who need to recuse could make sense, but I don't think it should be possible to kick out someone just because some of their supports didn't show up that day. It is an extraordinary measure, and I find it appropriate for it to need an extraordinary level of support.

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
+<h6 id="collective-removal">
+Collective removal</h6>
+
+	The [=Advisory Committee=] <em class=rfc2119>may</em>
+	hold a <dfn export>vote of no confidence</dfn>
+	in the [=Advisory Board=] or the [=Technical Architecture Group=].
+
+	An [=Advisory Committee representative=] initiates a [=vote of no confidence=]
+	by sending a request to the Team, and <em class=rfc2119>should</em> also share this request with the Advisory Committee.
+	The request <em class=rfc2119>must</em> identify which of the [=AB=] or [=TAG=] is targeted,
```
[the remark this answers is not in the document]
frivoal: Does this help?
```suggestion
	The request <em class=rfc2119>must</em> identify whether the [=AB=] or [=TAG=] is targeted,
```

index.bs
frivoal: It already is a 3 stage thing:
1. someone calls for a vote of no confidence
2. we check if >=5% of the membership agree that we should run a vote of no confidence (within a time limit of 1 week)
3. if so, we run the actual vote of no confidence

Adding an "at least two people need to agree" seems redundant with step two.

index.bs
frivoal: Does this help?
```suggestion
	The [=Team=] must announce the conclusion of the [=Advisory Committee Override=] to the [=Advisory Committee=].

```

index.bs
frivoal: That seems like a separate discussion. The previous bullet point in this list already deal with the case of seats being vacated for reasons other than a successful vote of no confidence, by letting the chair decide whether to run an special election or not. Your proposal would make it no longer a choice by the chair, which may or may not be a good idea, but is a separate idea. If you want to pursue that, could I ask you to open a separate issue?

index.bs
frivoal: Works for me. I'll let chairs decide if there's consensus on this point or not.

index.bs
frivoal: This is the pre-existing mechanism and threshold used for AC appeals. This attempts to reuse this system, not to change this.

And you're right, I don't think a 5% threshold is small. It is actually a non trivial barrier to meet. But I don't think it should be trivial to launch votes of no confidence (or AC appeals), so I'm personally comfortable with that.

If you do want to insist on a lower threshold, could you clarify if you want that change both for AC appeals AND votes of no confidence (in which case I'd suggest opening a separate issue), or only for votes of no confidence (in which case this is indeed the right place to discuss it)?

index.bs
frivoal: We're not defining this mechanism here, just reusing it (and adjusting the phrasing slightly to make it work in new contexts). I suggest filing a separate issue if you dislike the mechanism itself. For context, this was adopted through PR https://github.com/w3c/process/pull/901, based on issue https://github.com/w3c/process/issues/886 (the prior state was a simple 50% majority, independent of the level of participation).

Now, as to the substance of the question, in case you do decide to pursue it, here's a little extra information / opinion:

* It is true that this is discontinuous, and has slightly weird properties because of that, but I don't think it actually matters in practice, because it's very hard to operationalize: it's only reliably advantageous for "no" voters to refrain voting when close to the threshold if they can (a) know what the current tally is, and (b) count on no more "yes" vote coming in. Depending on how we set up the vote, (a) might be true, but we could easily guard against that by deciding that this must always be a secret ballot. And even if we don't, (b) won't be true, so people cannot really count on staying below the threshold.

* I think there is value in having something of that nature in general to avoid consequence-heavy decisions being taken by accident just because not enough people were paying attention. Especially in the case of votes of no confidence, I consider them more likely than other kind of votes to be at risk of being invoked repeatedly by disgruntled parties. I think it would be unfortunate if the nth instance of a vote of no confidence passed despite wide opposition, simply because people got tired of answering largely the same question over and over again, or even on the first try because we're just past an election, and too many don't bother answer because they think it's obvious. So having a high bar when participation is low seems appropriate.

* Though it is indeed a little odd, I think there's value in matching the same system used by the bylaws rather than coming up with something similar but different, even if better. Or we should change both to match.

index.bs
martinthomson: That would be fine for the latter, but there is no requirement on the TAG/AB to announce when they act.

index.bs
martinthomson: I think that the second step is not feasible (you allow more time for the final vote, which only needs to clear the same threshold).  So I am suggesting a replacement for that stage.

index.bs
martinthomson: I sort of agree, but I don't think that you can separate that out completely.  You are introducing new ways in which the membership of these groups can drop to critically low levels.  Though we don't expect this process to be used often, in combination with existing stuff, you have increased the risk of the groups being understaffed.  Staffed to zero is just one special case.

index.bs
martinthomson: I'm more concerned about the time limit.  Not only do you have to muster 5% of the membership, you have to do so within a week.

I was viewing this with a lens of the vote of no confidence, but I can't think of a case where this wouldn't be a problem for other such appeals.  I don't want this sort of thing to be easy to trigger, but right now it seems to be akin to impossible.  A healthy system might have this appeal run once every few years, but I can't remember a single occurrence, partly because this requires the moving of mountains.  An accountability system that does not operate is probably one that cannot function, so it might as well not exist.

index.bs
wareid: What might be more appropriate is that for the vote to succeed, all members of the elected body must participate (avoiding the issue of "my supporters didn't show up), and thus for the proposal to carry a simple majority is sufficient. Not requiring every member to participate is where risk is introduced. To be additionally clear, abstaining is acceptable.

index.bs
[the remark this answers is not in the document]
wareid: Perfect.

index.bs
frivoal: Exclusion by simple majority seems inappropriate to me for something that amounts to a disciplinary procedure. In cases where it is warranted, reaching a significant super majority (everybody but a friend or two) should be doable.

Simple majority seems far too easy to reach for mere differences of opinion, which is not what this should be about.

martinthomson: I can make a different argument: those who initiate this process risk the wrath of the AC (and the other process whereby the AC loses faith in all the participants).  They do so because *they believe that they cannot continue to work with the identified person*.  They have to convince a whole bunch of other people -- half the body -- that this extraordinary disciplinary action is warranted.  The result -- success or failure -- is then made public.

I've seen some dysfunction in groups, but never at this level.  If it were the case that half the body wants someone gone, that's enough for me.

I would instead say that removal on the basis of convincing half a body is already too high a bar to clear.

cwilso: I would object to a simple majority (even with all voting; in practice, those tend to mean the same thing).  I could be agreeable to a 2/3rds supermajority (instead of 3/4), but simple majority is not sufficient, in my strongly-held opinion.

index.bs
```diff
@@ -1309,15 +1309,66 @@ Verifiable Random Selection Procedure</h5>
...
 
+		<li>
+			the participant is [=removed=], or
+
+		<li>
+			a [=vote of no confidence=] in the body they are part of is successful, or
+
 		<li>
 			an Advisory Board or TAG participant changes affiliations
 			such that the <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a> are no longer met,
```
TallTed, line 1374: I can only suggest through line 1374, but similar changes should be made through line 1384.
```suggestion
	An [=Advisory Board=] or [=TAG=] participant's seat is <dfn lt="vacated|vacant">vacated</dfn> when any of the following occurs:

	<ul>
		<li>
			The participant resigns.
		</li>
		<li>
			The participant is [=removed=].
		</li>
		<li>
			A [=vote of no confidence=] in the body they are part of is successful.
		</li>
		<li>
			An Advisory Board or TAG participant changes affiliations
			such that the <a href="#AB-TAG-constraints">Advisory Board and TAG participation constraints</a> are no longer met.
		</li>
```

index.bs
martinthomson: We're clearly in disagreement then, because I am convinced that this is not usable.  Even at a simple majority.

It's possible that a mechanism like this is not the one that would be engaged in more extreme cases (obvious and overt bullying, for instance).  But if we don't make this a usable process, then people who find themselves in a difficult situation will choose to stop participating rather than to initiate this process.

Even 2/3 is virtual unanimity in groups this size.  Consider an 11 person group (the standard maximum size of both AB and TAG.  With one person eliminated, that means 10 voting participants.  To clear 3/4, you need 8 votes in support.  For 2/3, you need to get to 7.  A simple majority would be 6.  For something like this, where recusal or failure to vote is counted the same as a vote against the motion, the margins are thin.

If I'm in the rough, then so be it.

frivoal: > Even 2/3 is virtual unanimity in groups this size.

Yes, and I think virtual unanimity is appropriate. That's what we have on the Board, and I've seen (and used) similar provisions in other entities. Needing virtual unanimity to kick a elected member out is desirable. With simple majority, it is far too easy to gang up on someone just because their views don't align with a big chunk of the group.

Leaving room for a couple of contrarians/friends refusing the vote a justified dismissal avoids unnecessary blockage, but if you've got 3 or 4 people out of 10 in support of the potential dismissal, what you have on your hands is a political problem, not a simple disciplinary problem, and those aren't appropriate to solve by dismissal.

wareid: Gaming out this scenario in my head based on the current process, I see one major issue with forcing a two thirds or three quarters majority for these groups in a few delicate scenarios. The process already has a provision for the CEO removing a member of the AB/TAG for either not conducting themselves to the standards of the body or to the organization (while not defined this way in the process, it's fairly clear this is to the standard of the code of conduct). With that already in place, I see this mechanism coming into play only in scenarios where there is a difference in opinion between the AB/TAG and CEO on the reason for dismissal. Why I find it concerning to have such a high bar as 7-8 out of a max of 11 vs simple majority (which would be 5-6) is the potential nature of a case where the CEO refuses to act on an issue. 

I can see this becoming an issue in cases of the more grey areas of the code of conduct: bullying, continued harassment (especially microaggressions), or interpersonal misconduct. The code of conduct directs us to focus on impact over intent, but I can still see a scenario where a member of an elected body continually mistreats another member, and the impacted member goes to the CEO to have it addressed and is dismissed as it is viewed as "not so bad" or "just ignore them". In cases where this situation gets to the point where the person needs to take it to a dismissal process, having the 2/3 or 3/4 majority would require potentially convincing 7-8 people of the issue occurring, which puts significant strain on the person making the accusation. I worry we'd more likely see people choosing to leave the elected body for their own protection over having to put themselves through this kind of process.

cwilso: There is a serious problem with enabling a simple majority of members to the rest of the entire body.  If you enable a simple majority to eject members,  those 5-6 members could eject _**everyone else off the body**_.  This may seem like an extreme, and like "this would never be done" - but if that were possible in US politics, I think we all know it 100% would have been used by now.  The bar for ejecting members _**should not**_ be so low - it *should* be a strong consensus of the rest of the body.

The CEO is not the first port of call to address a case where there is mistreatment.  The Chair should be.  After that, perhaps the CEO, or perhaps first the ombuds.  The entire body should be involved if necessary, but having to convince 2/3 of your colleagues is not the first step; and we should be selecting Chairs to be the necessary arbiters of fairness and justice.

We HAVE diversity in our elected bodies; and as deeply as I agree that we should view our processes through the lens of the marginalized, I want to ensure that we also protect against abuse of tools we put in place.  If we don't believe that censure by the Chair will have effect, or that 2/3rds of the body won't care about abuse, then frankly we are likely to have a Bad Time regardless; but even then, we should not enable the body to be taken over by a simple majority.

martinthomson: I find that scenario less plausible than the one Wendy described.  If the W3C ends up as badly dysfunctional as current US politics, I'll have taken my toys elsewhere long past.

Consider that the group needs to report the action to the community.  Do that often and it will be abundantly clear to the entire membership.

frivoal: > we need to view this process through the lens of the marginalized, we want diversity in our bodies, to achieve that we need processes that ensure the most vulnerable are protected.

Well, that makes me double down on super-majority even more. Unless we are content with marginalized people being tokens of diversity who are welcome to be there as long as they don't say anything that might upset the status quo, we need to make sure that they are comfortable speaking up even if they're going to go against the grain.

If you have to weight everything you say against the risk of being kicked out, then you'll say much less, even when what you have to say is legit. There should be a very high bar before kicking out someone who has been duly elected, and protecting the rights of minorities is among the reasons for that. Let's not risk a tyranny of the majority.

I also don't think it's appropriate in a governance document to assume that people will be well-behaved, and that what happens in other polities cannot happen to us. We're just as human as anywhere else, and the whole point of having formal governance is to deal with cases where things can go wrong. «I'll just move to [IETF | Canada] if things end up too dysfunctional» may be a comforting thought, but it isn't good governance.

frivoal: This discussion is quite distinct from the rest of the points made about this PR. To try and keep things contained, I have split the PR in two, with this particular aspect being moved to its own PR: https://github.com/w3c/process/pull/986

fantasai:
I think this PR needs some updating; and we should clean up the discussion to focus on the current proposal at hand, rather than on previous variations of trying to address this problem. @frivoal Would it make sense to do the clean-up here, or to summarize the open points of discussion into the issue and open a new PR?

index.bs
frivoal: Merging in the proposed tweak above. The other half of this question is handled by https://github.com/w3c/process/pull/1023

index.bs
fantasai: What threshold should trigger an automatic election is an interesting question, but probably should be discussed as its own issue. What we can say for sure is that a threshold of zero needs to trigger an election. :) So how about we write the PR here for the case of zero, and we can open an issue about possibly raising that number higher.

frivoal: I think that works. We do have a means to deal with a low (non-zero) number of participants (chair discretion) but it seems worth discussing whether an automatic (non-zero) threshold would be better. I'm landing the zero threshold change, let's discuss the rest a https://github.com/w3c/process/issues/1032

```suggestion
			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
```

index.bs
fantasai: Three months is not the term being elected for, it's the time remaining until the next election: the term might extend for another year on top of those three months.

Three months is the existing threshold we have for not scheduling a special election because it takes approximately that long to organize the regular election anyway.

So I think we keep this threshold at three months, and if we want something to handle excessively short terms, then that would be a separate point.

martinthomson: There are two things: when the election is held (at vacancy vs. on the regular schedule) and what the term is (the remainder of the vacated term, or something else).  Those need to both be specified clearly.

I would prefer that vacancies not exist for too long, but maybe that is better addressed in #1032, rather than here.

As for the term, the assumption will be that the term is for the period of vacancy, unless otherwise specified.  Personally, I don't like the idea that you might get term+N unless N is sufficiently small (3 months is maybe small enough).

fantasai: > As for the term, the assumption will be that the term is for the period of vacancy, unless otherwise specified.  Personally, I don't like the idea that you might get term+N unless N is sufficiently small (3 months is maybe small enough).

Agreed. Filed as https://github.com/w3c/process/issues/1033

index.bs
fantasai: @martinthomson Can you file a separate issue about increasing the time limit? (And fwiw, an AC appeal has occurred at least once, since I recall there was one during the last decade or so.)

index.bs
frivoal: Good suggestion. Taking it in.

index.bs
fantasai: I think this conversation is off-topic. Hopefully also resolved, but @martinthomson if you want to pursue this, please file a separate issue.

index.bs
```diff
@@ -1309,15 +1309,41 @@ Verifiable Random Selection Procedure</h5>
 			The incomplete terms are assigned in result order.
 	</ol>
 
+<h5 id="AB-TAG-removal">
+Removing AB or TAG members</h5>
```
fantasai, line 1313:
```suggestion
<h5 id="AB-TAG-recall">
AC Recall of the AB or TAG</h5>
```

index.bs
martinthomson: None of this appears in the current process.  I don't know how to manage that.

index.bs
frivoal: Filed at https://github.com/w3c/process/issues/1034

index.bs
martinthomson: I see.  I disagree with that decision, but I'll take it up there.

index.bs
```diff
@@ -1344,6 +1344,31 @@ Removing AB or TAG members</h5>
...
+	The request <em class=rfc2119>must</em> identify whether the [=AB=] or [=TAG=] is targeted,
+	and <em class=rfc2119>should</em> also include the rationale.
+
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
+	on the proposal.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
+	if the [=Advisory Committee Override=] proposal passes,
+	all [[#AB-TAG-elections|elected]] or [[#TAG-appointments|appointed]] seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
```
TallTed, line 1367:
```suggestion
	all [[#AB-TAG-elections|elected]] and [[#TAG-appointments|appointed]] seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
```

index.bs
```diff
@@ -1398,16 +1423,19 @@ Elected Groups Vacated Seats</h5>
...
-			when added to the number of continuing participants,
-			the minimum total number of elected seats is met
-			(8 for the [=TAG=], 9 for the [=AB=]);
-			and the maximum number corresponds to all unoccupied seats.
-			Except for the number of available seats and the length of the terms,
-			the <a href="#AB-TAG-elections">usual rules for Advisory Board and Technical Architecture Group Elections</a> apply.
+		<li>
+			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
+			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
+			unless the next regularly scheduled election is fewer than three months away.
```
jyasskin, line 1429: I'm worried about there being no TAG or AB at all for the 3 months + the normal election schedule. #1033 suggests extending terms if they're "too short". Riffing on that, what if this said something like

```suggestion
			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
			unless the next regularly scheduled election is fewer than three months away.
			If the next regularly scheduled election is fewer than three months away, that
			election must be held as soon as practical, with the newly elected members
			serving until the dates their terms would have ended if the election had
			been run at its normal schedule.
```

TallTed: Riffing on @jyasskin's suggestion... "fewer than three months" doesn't read right to me. I think "less than three months" is the proper wording. Also, the repeated phrase feels clumsy, hence "in which case".
```suggestion
			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
			unless the next regularly scheduled election is less than three months away,
			in which case, that
			election must be held as soon as practical, with the newly elected members
			serving until the dates their terms would have ended if the election had
			been run at its normal schedule.
```

martinthomson: That's a really long sentence @TallTed.

The terms of newly elected members is not specified for the case where the regular election is three months or more away.  Also, given that this is a case where the election will be replacing people who would not ordinarily have been up for election at the next cycle, some more care is required in specifying terms.

TallTed: @martinthomson — 
```suggestion
			If vacancies reduce the number of [=AB=] or [=TAG=] participants to zero,
			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=],
			unless the next regularly scheduled election is less than three months away.
			In that case, the schedule is to be accelerated, and the next
			election held as soon as practical. In all cases, the newly elected member(s)
			serve until the date(s) their term(s) would have ended had the election
			been run at its normal schedule.
```

frivoal: > I'm worried about there being no TAG or AB at all for the 3 months + the normal election schedule.

That wouldn't be 3 months plus the election schedule, that would be up to 3 months, including the normal election schedule, which is itself 2 months long.

I am unsure this change is necessary. An election takes 2 months to run if started instantly: 1 month for nominations, 1 for the actual election. If we add a little bit of inertia (say, one week) for setting things up (writing/sending the announcements, preparing the nomination system…), this anticipated election clause speeds up things by at most 3 weeks, in the "best" case when the recall happened at 3 months minus epsilon. If the recall happens later, the speed up is even smaller, and it happens within 2 months (plus a bit) of an election, no speed up is possible.
I'm not convinced that an potential speed up of a couple weeks, on a rare event, if it happens to be timed just right, is worth the extra rule.

timscharfenort8 (commented):
Good work

index.bs
fantasai: Closing this thread as out-of-scope. If anything needs follow-up, please _file a new issue_.

index.bs
fantasai: I think we should be running the call for nominations in parallel with the recall vote--maybe have it close a week after the recall vote closes. Then the election will only take roughly a month.

index.bs
fantasai: @TallTed I'm going to close this discussion, since the text is pre-existing and this is an editorial change to it that is not tied to this PR. Feel free to open a separate PR. :)

index.bs
fantasai: @TallTed I think that's worth handling as a separate PR. This discussion is complicated enough already, let's keep the diffs as simple as we can.

index.bs
```diff
@@ -1346,6 +1346,31 @@ Removing AB or TAG members</h5>
...
+
+	Within one week, the [=Team=] <em class=rfc2119>must</em> initiate an [=Advisory Committee Override=]
+	on the proposal.
+
+	The conclusion of the [=Advisory Committee Override=] is final,
+	and cannot be the subject of a [=Formal Objection=] nor of an [=AC Appeal=]:
+	if the [=Advisory Committee Override=] proposal passes,
+	all [[#AB-TAG-elections|elected]] or [[#TAG-appointments|appointed]] seats on [=AB=] or [=TAG=] are [=vacated=] immediately;
+	if it fails, it cannot be invoked on the same body
+	sooner than six months since the previous invocation.
```
frivoal, line 1371:
```suggestion
	sooner than six months since the previous initiation.
```

index.bs
frivoal: Addressed using: https://github.com/w3c/process/pull/888/commits/fe62b416672b69a201d8616f9368313960c4b9e2

index.bs
frivoal: I just realized where the confusion/disagreement may be stemming from. When talking about 3 months until the next election, I had assumed 3 months until the *close* of the next election. But the text didn't say that. If we were to add that, would you think (as I do) that that's enough to address the issue you raised, or would you still want to do something to address the case of hitting 0 members with the <1 month until the regularly scheduled call for nominations?

frivoal: @fantasai I do not think we should communicate during a recall in a way that normalizes the idea that it will likely be successful, by getting people to already prepare for the next steps as if it had succeeded already.

jyasskin: @frivoal Yes, I think that specifying that it's 3 months until the _end_ of the planned election solves the problem. Thanks!

frivoal:
```suggestion
			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
			unless the close of the next regularly scheduled election is fewer than three months away.
```

index.bs
```diff
@@ -1400,16 +1425,19 @@ Elected Groups Vacated Seats</h5>
...
-			when added to the number of continuing participants,
-			the minimum total number of elected seats is met
-			(8 for the [=TAG=], 9 for the [=AB=]);
-			and the maximum number corresponds to all unoccupied seats.
-			Except for the number of available seats and the length of the terms,
-			the <a href="#AB-TAG-elections">usual rules for Advisory Board and Technical Architecture Group Elections</a> apply.
+		<li>
+			If vacancies reduce the number of [=AB=] or [=TAG=] participants to 0,
+			the [=Team=] <em class=rfc2119>must</em> organize a [=special election=]
+			unless the close of the next regularly scheduled election is fewer than three months away.
```
jyasskin, line 1431: This is nitpicky, but "close" [isn't well-defined before the call for nominations is finished](https://www.w3.org/policies/process/drafts/#AB-TAG-elections:~:text=Calls%20for%20Nominations%20are%20issued%20until%20a%20sufficient%20number%20of%20people%20have%20been%20nominated), and it's ambiguous even after that between the "deadline for votes" and the day the results are announced. The other precise deadline, in https://www.w3.org/policies/process/drafts/#TAG-appointments, is "unless a regular Call for Nominations is scheduled within 2 months", and I suggest we use the same phrasing here, except with a tighter bound to match the 3-month upper bound on having-no-TAG-or-AB:

```suggestion
			unless a regular Call for Nominations is scheduled within 1 month.
```

Note that this can result in having two active elections running at the same time, so it might still be a good idea to take [my suggestion](https://github.com/w3c/process/pull/888#discussion_r2058785961) to move the normal election earlier if it's 1-2.5 months until the regular call for nominations. But that's not a critical part of this change, so it's reasonable to handle it in a separate PR.

frivoal: Good point. I agree this phrasing is more precise.

index.bs
```diff
@@ -1346,6 +1346,31 @@ Removing AB or TAG members</h5>
...
+<h5 id="AB-TAG-recall">
+AC Recall of AB or TAG</h5>
+
+	The [=Advisory Committee=] <em class=rfc2119>may</em>
+	<dfn export>recall</dfn>,
+	as a whole,
+	all [[#AB-TAG-elections|elected]] and [[#TAG-appointments|appointed]] participants
+	from the [=Advisory Board=] or the [=Technical Architecture Group=].
+
+	An [=Advisory Committee representative=] invokes a [=recall=]
```
cwilso, line 1358: It seems like the bar for triggering this should be much higher than a single AC representative request.

frivoal: read on. This starts with a single person (how else), but then you need to meet the same criteria as an AC appeal, which means finding 5% needs to support having the vote, or it stops right there.

index.bs
```diff
@@ -1356,7 +1381,7 @@ Elected Groups Vacated Seats</h5>
 			the participant resigns, or
 
 		<li>
-			the participant is [=removed=], or
+			the participant is [=removed=] or [=recalled=], oor
```
hober, line 1384:
```suggestion
			the participant is [=removed=] or [=recalled=], or
```

typo

index.bs
```diff
@@ -1346,6 +1346,31 @@ Removing AB or TAG members</h5>
 
 	Note: Removal from the [=AB=] or [=TAG=] does not imply removal from a convened [=Council=].
 
+<h5 id="AB-TAG-recall">
+AC Recall of AB or TAG</h5>
+
+	The [=Advisory Committee=] <em class=rfc2119>may</em>
+	<dfn export>recall</dfn>,
+	as a whole,
+	all [[#AB-TAG-elections|elected]] and [[#TAG-appointments|appointed]] participants
```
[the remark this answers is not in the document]
hober: That said, it seems like the Team ought to have the power to recall people from appointed seats. It put them there for reasons. If it turns out their presence is not conducive to advancing those reasons, the Team ought to be able to remove them and appoint someone else.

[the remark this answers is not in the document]
frivoal: > Why can the AC recall appointed seats?

Individually, it should not (and in this proposal, doesn't). As a whole, that's indeed debatable. Here, the idea is that if specific individuals, appointed or elected, need to be removed the group itself should deal with that, but the AC finds the group as a whole to be dysfunctional / harmful beyond redemption, it can call for a clean slate. Why kick the appointed people out too, even though they're not elected by the membership? My idea would be that (a) if the group is a whole is broken, they're part of that, and (b) they're appointed by the Team in order to complement the elected positions. If we're starting from scratch on that side, they need reevaluation too.

That said, while it makes sense to me to do it this way, I don't think I'd object to the recall only affecting the elected seats.

As for the Team having the power to recall the people it put it place, I'd need to think about it. First, it's not purely a Team appointment: the team proposes, but the TAG+AB decides. TAG members are supposed to act indepentently, as individuals, not in the interests or on behalf of any particular organization, but for the common good of the web. If the Team can give you the job and take it away, I think that might be a bit too direct of a power dynamic for my taste, making the appointees feel that they work for the Team, which I don't think would be appropriate. Also, "out their presence is not conducive to advancing those reasons" seems too low a bar to eject people. Removal from office is a pretty drastic measure, and I don't think being mediocre, or being capable but not in the expected way, is a justifiable reason.

Jorodimitrov85 (approved):

brentzundel:
The Revising W3C Process CG just discussed `https://github.com/w3c/process/pull/888`, and agreed to the following:

<details><summary>The full IRC log of that discussion</summary>
&lt;brent> subtopic: https://github.com/w3c/process/pull/888<br>
&lt;Ian> Brent: Say "merge", "continue", or "close"<br>
&lt;brent> Github: https://github.com/w3c/process/pull/888<br>
&lt;Ian> PLH: We have a "Needs AB feedback" label<br>
&lt;Ian> [We review the background of the pull request]<br>
&lt;Ian> Brent: I'm hearing it's not ready to merge; we need more AB and TAG feedback.<br>
</details>

frivoal:
Merging into the [ab-tag-discipline](https://github.com/w3c/process/tree/ab-tag-discipline) branch, as decided on the 2025-09-24 Process CG call. That branch itself is not being merged into main, and does not represent consensus, merely work in progress.

index.bs
ID-iot-git: I don't actually fancy doing all this.

PR #909, opened 2024-08-20
Deal with procedural disagreements within the Council

### LINKED ISSUE #843: Dealing with procedural disagreements within the Council
While discussing a case, if there are disagreement within the Council about the Process that governs the Council and the processing of formal objections, how do we move forward?

This is not about disagreement about the substance of the case at hand, as this is already handled (“However, if despite careful deliberation a W3C Council is unable to reach consensus, the W3C Council Chair may instead resort to voting”). This is about the rules that apply to the Council itself: “aren't we required to …?”, “Isn't this out of scope for us?”, “Do we even have the ability to …?”, etc.

There isn't clearly any higher authority to escalate to, and besides, the deliberations of the council being confidential, it's hard to go ask someone about something you cannot discuss.

Hopefully, the Process doesn't have that many ambiguities, and we don't run into such a case too often, but we need an escape valve to avoid a situation where the Council is stuck but because it cannot agree on what it can do.

I can think of a few variants, but the one that makes the most sense to me is:
a) The chair can propose a vote to resolve any such disagreement, which passes if there are more ballots for than against (chair breaks ties?).
b) The nature of the disagreement, vote totals, and resulting decision _must_ be included in the Council report.
c) If the AC is unhappy with the outcome of the Council, including because it thinks the Council got it wrong about this procedural disagreement, AC Appeal may be used to overturn the Council's decision (no need for a new rule for this point, this is already available).

frivoal:
Closes #843

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/909.html" title="Last updated on Aug 20, 2024, 8:53 AM UTC (4592393)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/909/bcbfcdf...frivoal:4592393.html" title="Last updated on Aug 20, 2024, 8:53 AM UTC (4592393)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2639,6 +2639,27 @@ Council Deliberations</h5>
 	and of the status of the discussions.
 	The [=W3C Council Chair=] <em class=rfc2119>may</em> additionally make this report <a href="#confidentiality-levels">public</a>.
 
+<h5 id=procedural-disagreements>
+Procedural Disagreement Within the Council</h5>
+
+	<dfn for="W3C Council">Procedural disagreement</dfn> occurs
+	when there is [=dissent=] between members of a [=Council=]
+	about the meaning of those sections of the Process
+	that govern the [=Council=] itself,
+	the handling of [=Formal Objections=],
+	or the interaction with [=AC Reviews=],
+	in a way that it interfers with the [=Council=]'s ability to deliberate.
+
+	In such cases,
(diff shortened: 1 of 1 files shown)


frivoal:
Closing in favor of https://github.com/w3c/Guide/pull/244

Issue #638, opened 2022-09-22
REC track diagram does not reflect normative text


gsnedders:
The diagram in https://www.w3.org/Consortium/Process/Drafts/#rec-track doesn't seem to be fully explainable by normative text.

Neither the [W3C Recommendation Track](https://www.w3.org/Consortium/Process/Drafts/#w3c-recommendation-track) definition, nor the [Candidate Recommendation](https://www.w3.org/Consortium/Process/Drafts/#RecsCR) definition, nor the [Transitioning to Proposed Recommendation](https://www.w3.org/Consortium/Process/Drafts/#transition-pr) section seem to actually clarify that the final CR prior to advancement has to be a CR Snapshot, which I believe (see #636) is the intention of the diagram.

plehegar:
In [Transitioning to Candidate Recommendation](https://www.w3.org/2021/Process-20211102/#transition-cr), we have:
[[
The first Candidate Recommendation publication after approval of a [Transition Request](https://www.w3.org/2021/Process-20211102/#trans-req) is always a [Candidate Recommendation Snapshot](https://www.w3.org/2021/Process-20211102/#candidate-recommendation-snapshot).
]]

So, the first CR published is also a CR Snapshot even if it doesn't say so explicitly.

Then, in [Transitioning to Proposed Recommendation](https://www.w3.org/2021/Process-20211102/#transition-pr), we have:
[[
must not have made any [substantive changes](https://www.w3.org/2021/Process-20211102/#substantive-change) to the document since the most recent [Candidate Recommendation Snapshot](https://www.w3.org/2021/Process-20211102/#candidate-recommendation-snapshot), other than dropping features identified [at risk](https://www.w3.org/2021/Process-20211102/#at-risk).
]]

So, your last CR prior to advancement may be a CR Draft, as long as you're not trying to introduce substantive changes since the last CR Snapshot.

plehegar:
The diagram could be improved indeed. The risk being to make it too complicated to read.

fantasai:
Yeah, we'd have to introduce a bubble in the middle of the arrow to PR that says “CRD with only editorial changes since previous CRS” ... which seems maybe a bit too complicated to put in the diagram? There's no harm in publishing a new CRS in between, if someone interprets it that way anyway; that can be done (since it's only editorial changes) very easily.

fantasai:
(To be honest, I'd almost prefer to require the extra CRS than complicate the diagram further, since editorial CRS updates are automatic).

[...]

PR #911, opened 2024-08-20
Explicitly mark figures as non-normative

frivoal:
See #638


tantek (approved):

nigelmegitt (approved):
When reviewing this I noticed that we use both "non-normative" and "informative", and raised #914.

PR #912, opened 2024-08-29
Reference w3c-ac-forum policy which defines list access policy

tobie:
Addresses concerns raised on the w3c-ac-forum mailing list https://lists.w3.org/Archives/Member/w3c-ac-forum/2024JulSep/0143.html

***
<a href="https://pr-preview.s3.amazonaws.com/tobie/process/pull/912.html" title="Last updated on Sep 13, 2024, 10:51 AM UTC (be85616)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/912/bcbfcdf...tobie:be85616.html" title="Last updated on Sep 13, 2024, 10:51 AM UTC (be85616)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -785,10 +785,10 @@ Advisory Committee Mailing Lists</h4>
 			One for discussion among [=Advisory Committee representatives=].
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion,
-			<em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM]] for this list,
+			<em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]] for this list,
 			<em class="rfc2119">should</em> participate in discussion when appropriate,
                         and <em class="rfc2119">may</em> grant access to this list to relevant individuals or groups, 
-                        which <em class="rfc2119">must</em> be documented in the list's <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM]].
+                        which <em class="rfc2119">must</em> be documented in the list's <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]].
 			Ongoing detailed discussions <em class="rfc2119">should</em> be moved to other appropriate lists
 			(new or existing, such as a mailing list created for a <a href="#GAEvents">Workshop</a>).
 	</ol>
@@ -6019,7 +6019,7 @@ Changes since earlier versions</h3>
 		"href": "https://www.w3.org/about/council/",
 		"title": "Council Report Index"
 	},
-        "AC-FORUM": {
+        "AC-FORUM-POLICY": {
 		"href": "https://www.w3.org/2015/10/ac-forum-pol",
 		"title": "Policy for w3c-ac-forum Mailing List"
         }
--- a/index.bs
+++ b/index.bs
@@ -711,16 +711,17 @@ Resignation from a Group</h4>
 	See “Exclusion and Resignation from the Working Group” in the W3C Patent Policy [[PATENT-POLICY]]
 	for information about obligations remaining after resignation from certain groups.
 
-<h4 id="emeritus">
+<h4 id="emeritus-status">
 Emeritus status</h4>
-        Emeritus status is granted to
+        <dfn export lt="Emeritus Status|Emeritus">Emeritus status</dfn> is granted to
         elected or appointed participants
         of the <a href="#AB">Advisory Board</a> and <a href="#TAG">Technical Architecture Group</a>
         and [=Board=] directors
         once their term has completed.
-        Individuals with emeritus status <em class="rfc2119">may</em> 
-        join the <a href="#ACCommunication">Advisory Committee Mailing List</a> for discussions
-        and <em class="rfc2119">may</em> participate in discussions.
+        Individuals with [=Emeritus Status=]
+        are authorized to access <a href="#Member-only">Member-only</a> information and
+        <em class="rfc2119">may</em> join the <a href="#ACCommunication">Advisory Committee Mailing List</a> for discussions
+        to participate in discussions.
 
 <h3 id="AC">
 The Advisory Committee (AC)</h3>
@@ -797,7 +798,7 @@ Advisory Committee Mailing Lists</h4>
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion
 			and <em class="rfc2119">should</em> participate in discussion when appropriate.
-                        Additionally, indivudals with <a href="#emeritus">emeritus status</a>
+                        Additionally, indivudals with [=Emeritus Status=]
                         <em class="rfc2119">may</em> join the list and also participate in discussion.
 			Ongoing detailed discussions <em class="rfc2119">should</em> be moved to other appropriate lists
 			(new or existing, such as a mailing list created for a <a href="#GAEvents">Workshop</a>).
@@ -4962,7 +4963,8 @@ Confidentiality Levels</h3>
 	<a href="#invited-expert-wg">Invited Experts</a>,
 	the Advisory Board,
 	the TAG,
-	and the Team.
+	the Team,
+        and individuals with [=Emeritus Status=].
 	For example,
 	the <a href="#WGCharter">charter</a> of some Working Groups
 	<em class="rfc2119">may</em> specify a [=Member-only=] confidentiality level for group proceedings.
--- a/index.bs
+++ b/index.bs
@@ -711,6 +711,17 @@ Resignation from a Group</h4>
 	See “Exclusion and Resignation from the Working Group” in the W3C Patent Policy [[PATENT-POLICY]]
 	for information about obligations remaining after resignation from certain groups.
 
+<h4 id="emeritus">
+Emeritus status</h4>
+        Emeritus status is granted to
+        elected or appointed participants
+        of the <a href="#AB">Advisory Board</a> and <a href="#TAG">Technical Architecture Group</a>
+        and [=Board=] directors
+        once their term has completed.
+        Individuals with emeritus status <em class="rfc2119">may</em> 
+        join the <a href="#ACCommunication">Advisory Committee Mailing List</a> for discussions
+        and <em class="rfc2119">may</em> participate in discussions.
+
 <h3 id="AC">
 The Advisory Committee (AC)</h3>
 
@@ -786,6 +797,8 @@ Advisory Committee Mailing Lists</h4>
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion
 			and <em class="rfc2119">should</em> participate in discussion when appropriate.
+                        Additionally, indivudals with <a href="#emeritus">emeritus status</a>
+                        <em class="rfc2119">may</em> join the list and also participate in discussion.
 			Ongoing detailed discussions <em class="rfc2119">should</em> be moved to other appropriate lists
 			(new or existing, such as a mailing list created for a <a href="#GAEvents">Workshop</a>).
 	</ol>
--- a/index.bs
+++ b/index.bs
@@ -798,8 +798,8 @@ Advisory Committee Mailing Lists</h4>
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion
 			and <em class="rfc2119">should</em> participate in discussion when appropriate.
-                        Additionally, individuals with [=Emeritus Status=]
-                        <em class="rfc2119">may</em> join the list and also participate in discussion.
+                        Individuals with [=Emeritus Status=]
+                        <em class="rfc2119">may</em> also join the list and participate in discussions.
 			Ongoing detailed discussions <em class="rfc2119">should</em> be moved to other appropriate lists
 			(new or existing, such as a mailing list created for a <a href="#GAEvents">Workshop</a>).
 	</ol>
--- a/index.bs
+++ b/index.bs
@@ -711,18 +711,6 @@ Resignation from a Group</h4>
 	See “Exclusion and Resignation from the Working Group” in the W3C Patent Policy [[PATENT-POLICY]]
 	for information about obligations remaining after resignation from certain groups.
 
-<h4 id="emeritus-status">
-Emeritus status</h4>
-        <dfn export lt="Emeritus Status|Emeritus">Emeritus status</dfn> is granted to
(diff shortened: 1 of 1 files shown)


chaals:
If the W3C decides to do this I suggest they extend generic "member" access. Otherwise it's a bit odd to be in discussions where it regularly isn't possible to find some key information.

tobie:
@chaals addressed in https://github.com/w3c/process/pull/912/commits/b6f377681a17806d3f07113798a4cf5d13546c1c.

frivoal:
Note: this matches existing practice, was recommended by the AB in 2019, and applied by CEOs (or under their delegation) since. https://www.w3.org/2019/11/20-ab-minutes.html#resolution05

Making it more visible is probably a good idea, though I wonder if /Guide isn't enough. The more formal we make the rule, the more we have to worry about getting the details right.

If we do make it a formal rule:
* this should be subordinated to acceptance by the people who get this access granted of some rules (such as the invited expert agreement) that at the very least subjects them to the confidentiality rules and the Code of Conduct. AB and TAG alumni may otherwise already be bound by such an agreement, but not necessarily.
* This is implicit true, so it may not need restatement, but this access should be subject to https://www.w3.org/policies/process/#discipline
* It would be good to be able to opt out of "authorized to access Member-only information". People should be allowed to leave W3C if they want to, and not be held responsible for an access they may no longer want.

index.bs
```diff
@@ -786,6 +798,8 @@ Advisory Committee Mailing Lists</h4>
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion
 			and <em class="rfc2119">should</em> participate in discussion when appropriate.
+                        Additionally, indivudals with [=Emeritus Status=]
```
tantek, line 801:
```suggestion
                        Additionally, individuals with [=Emeritus Status=]
```

tantek (changes requested):
minor fix to help this along

tobie:
> this should be subordinated to acceptance by the people who get this access granted of some rules (such as the invited expert agreement) that at the very least subjects them to the confidentiality rules and the Code of Conduct. AB and TAG alumni may otherwise already be bound by such an agreement, but not necessarily.

This seems like a good reasons to formalize this. As @chaals pointed out above, w3c-ac-forum participation requires member-level access. Only downside is this requires a new agreement and therefore some legal expenses.

> This is implicit true, so it may not need restatement, but this access should be subject to https://www.w3.org/policies/process/#discipline

Yeah, I don't see a reason to restate this. Could be part of the agreement, though.

> It would be good to be able to opt out of "authorized to access Member-only information". People should be allowed to leave W3C if they want to, and not be held responsible for an access they may no longer want.

I used "is authorized" on purpose, here. I think this gives the necessary flexibility.

jandrieu:
I'm confused.

Is this a new feature that AB/TAG/Board members get lifetime access? 

Or did I just not realize that was how the W3C does things?

chrisn:
I'm not convinced that people should have lifetime access. I'd prefer to see this status reviewed periodically, similar to what should happen with Invited Experts.

index.bs
```diff
@@ -711,6 +711,18 @@ Resignation from a Group</h4>
...
+Emeritus status</h4>
+        <dfn export lt="Emeritus Status|Emeritus">Emeritus status</dfn> is granted to
+        elected or appointed participants
+        of the <a href="#AB">Advisory Board</a> and <a href="#TAG">Technical Architecture Group</a>
+        and [=Board=] directors
+        once their term has completed.
+        Individuals with [=Emeritus Status=]
+        are authorized to access <a href="#Member-only">Member-only</a> information and
+        <em class="rfc2119">may</em> join the <a href="#ACCommunication">Advisory Committee Mailing List</a> for discussions
+        to participate in discussions.
```
TallTed, line 724:
```suggestion
        <em class="rfc2119">may</em> join the <a href="#ACCommunication">Advisory Committee Mailing List</a>
        to participate in discussions.
```

index.bs
```diff
@@ -786,6 +798,8 @@ Advisory Committee Mailing Lists</h4>
 			Though this list is primarily for Advisory Committee representatives,
 			the [=Team=] <em class="rfc2119">must</em> monitor discussion
 			and <em class="rfc2119">should</em> participate in discussion when appropriate.
+                        Additionally, individuals with [=Emeritus Status=]
+                        <em class="rfc2119">may</em> join the list and also participate in discussion.
```
TallTed, line 802:
```suggestion
                        Individuals with [=Emeritus Status=]
                        <em class="rfc2119">may</em> also join the list and participate in discussions.
```

chrisn:
Given that @koalie has explained existing policy over in w3c-ac-forum, I don't think Process changes are needed.

koalie:
> Given that @koalie has explained existing policy over in w3c-ac-forum, I don't think Process changes are needed.

The [policy for the w3c-ac-forum mailing list](https://www.w3.org/2015/10/ac-forum-pol) was [revised in November 2022](https://lists.w3.org/Archives/Member/w3c-ac-members/2022OctDec/0022.html) to add to the classes of participants "past members of the AB, at the discretion of the CEO"

chaals:
> I'm not convinced that people should have lifetime access. I'd prefer to see this status reviewed periodically, similar to what should happen with Invited Experts.

In principle, I agree. I have no problem with explicitly stating that status should be reviewed periodically. (To be clear, I also don't mind if the AC decides that the status of "formerly relevant person" should not grant ongoing access to the AC's recorded discussions).

In practice, the current approach is that Seth, or his successor with respect to managing this, can decide that I have become a doddering fool, or racist hate-monger, or spokes-person for an organisation trying to destabilise W3C, or a political inconvenience, or a waste of the AC's time, and she/he/they can shut off my mailing list access as they see fit, leaving me to consider what to do. 

Everyone knows that I (and others in my position) likely have access to many AC reps' email if I want a case to be argued. Likewise, nobody is really required to read what I write.

I'm skeptical that pushing a particular "disenfranchisement" power to e.g. the AC would improve anything.

(I note that I don't have access to https://www.w3.org/2019/11/20-ab-minutes.html#resolution05 - so it is somewhat difficult to comment on how this situation arose, despite being on the AB at the time).

index.bs
tobie: There are actually two AC mailing lists per process. So this text, as clumsy as it sounds, isn’t actually redundant. /cc @frivoal

tobie:
Bringing together the different threads for the sake of coming to a resolution here:

1. I opened this PR to address the concern expressed by @mnot that the practice of allowing past AB members to stay on the list wasn't the result of a formal decision.
2. It turns out that it was the result of an [AB resolution]( https://www.w3.org/2019/11/20-ab-minutes.html#resolution05) and was documented as part of the [policy for the w3c-ac-forum mailing list](https://www.w3.org/2015/10/ac-forum-pol).

We can either consider that the existing solution is good enough and close this PR or we can keep that PR open to address some of the additional points that were brought up:

1. Formalize this practice a bit more and make it more discoverable by including it in the process document (thereby addressing @mnot's further comment on the mailing list).
2. Extend this practice to past TAG members and board directors (which could also be done simply by an AB resolution).
3. Also grant access to membership-level information (this would require a formalized agreement and seems a little more involved).
4. Limit the duration in time of this access (which @koalie points out is challenging to implement in practice and is to some degree covered by the existing resolution of making the access at the discretion of the CEO).

I note in passing that current board directors aren't listed at all in the mailing list policy (which should probably be fixed).

My inclination would be to keep things simpler and more flexible by formally delegating who access to w3c-ac-forum to the policy, e.g.: "Though this list is primarily for Advisory Committee representatives, access might be granted to other groups or individuals through the list's policy."

index.bs
frivoal: Ah, I see. How about "may join and participate in the AC mailing list for discussions"?

jandrieu:
For our perspective, this is an unfortunate policy that should be reversed.

If we want certain individuals to be granted some form of emeritus status it should be explicit and formally decided, not automatically granted to anyone who ever gets elected to one of the groups that qualifies.

The policy as currently exists is the equivalent of those elected becoming  "senators for life".

While I respect the intentions of folks trying to figure out how to maintain continuity given the massive organizational transition we are necessarily going through, I can't, for the life of me, see how permanent, automatic, lifetime enfranchisement of existing leadership is at all compatible with the ideals of this organization.

tobie:
Revamped PR as per [my last comment](https://github.com/w3c/process/pull/912#issuecomment-2322887626).

@jandrieu just to clarify, this practice isn't granting any voting powers of any kind to anyone concerned by it, so while I hear your broader concerns, I don't think they're justified in this case. That said, if you think that this is an issue, I suggest filing an issue with the AB about it, as at this point, this PR is just documenting an existing practice.

mnot:
@tobie it's not just about voting powers, it's ability to influence discussion. It's well recognised that the AC has a few long-term participants who make most of the noise, and many who are relatively quiet. Giving the noisemakers permanent access is a step backwards.

Extending special access to _more_ people is also a step backwards. We should be working towards incorporating the views of a broader selection of the community of people actually performing and being affected by W3C's current work, not perpetuating access for a selected few.

You say:

> which could also be done simply by an AB resolution

This is not within the authority of the AB; it is (as discussed) a decision of the CEO. The AB can of course give its advice, but there's no requirement to route suggestions and advice from Members _through_ the AB, and doing so doesn't give the advice any authority.

index.bs
```diff
@@ -6015,7 +6018,11 @@ Changes since earlier versions</h3>
 	"COUNCIL-REPORT-INDEX": {
 		"href": "https://www.w3.org/about/council/",
 		"title": "Council Report Index"
-	}
+	},
+        "AC-FORUM": {
```
frivoal, line 6022:
```suggestion
        "AC-FORUM-POLICY": {
```

index.bs
```diff
@@ -784,8 +784,11 @@ Advisory Committee Mailing Lists</h4>
 		<li>
 			One for discussion among [=Advisory Committee representatives=].
 			Though this list is primarily for Advisory Committee representatives,
-			the [=Team=] <em class="rfc2119">must</em> monitor discussion
-			and <em class="rfc2119">should</em> participate in discussion when appropriate.
+			the [=Team=] <em class="rfc2119">must</em> monitor discussion,
+			<em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM]] for this list,
```
frivoal, line 788:
```suggestion
			<em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]] for this list,
```

index.bs
```diff
@@ -784,8 +784,11 @@ Advisory Committee Mailing Lists</h4>
...
+			<em class="rfc2119">should</em> participate in discussion when appropriate,
+                        and <em class="rfc2119">may</em> grant access to this list to relevant individuals or groups, 
+                        which <em class="rfc2119">must</em> be documented in the list's <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM]].
```
frivoal, line 791:
```suggestion
                        which <em class="rfc2119">must</em> be documented in the list's <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]].
```

index.bs
frivoal: I'd suggest changing the shorthand of this bibliographical reference, since it isn7t referring to AC Forum itself, but to a policy about it.

frivoal (approved):
The revamped PR looks good to me, as it simply documents what exists already, and makes it more discoverable. Lack of discoverability seems to be an issue, as very few people know about the existing documentation about AC Forum.

I'm suggesting a minor editorial tweak, but other than that, this works for me.

frivoal (approved):

chrisn:
In as much as this PR describes existing practice, I approve. But I do share some of the concerns being expressed, and would encourage a rethink of the policy.

index.bs
```diff
@@ -784,8 +784,11 @@ Advisory Committee Mailing Lists</h4>
 		<li>
 			One for discussion among [=Advisory Committee representatives=].
 			Though this list is primarily for Advisory Committee representatives,
-			the [=Team=] <em class="rfc2119">must</em> monitor discussion
-			and <em class="rfc2119">should</em> participate in discussion when appropriate.
+			the [=Team=] <em class="rfc2119">must</em> monitor discussion,
+			<em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]] for this list,
+			<em class="rfc2119">should</em> participate in discussion when appropriate,
+                        and <em class="rfc2119">may</em> grant access to this list to relevant individuals or groups, 
+                        which <em class="rfc2119">must</em> be documented in the list's <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]].
```
fantasai, line 791: Trying to reduce the redundancy somewhat...

```suggestion
			the [=Team=] <em class="rfc2119">must</em> monitor discussion,
			<em class="rfc2119">should</em> participate in discussion when appropriate,
			<em class="rfc2119">may</em> grant temporary or indefinite access to this list to relevant individuals or groups,
			and <em class="rfc2119">must</em> maintain a <a href="https://www.w3.org/2015/10/ac-forum-pol">policy<a> [[AC-FORUM-POLICY]] for this list.
```

fantasai:
> (I note that I don't have access to https://www.w3.org/2019/11/20-ab-minutes.html#resolution05 - so it is somewhat difficult to comment on how this situation arose, despite being on the AB at the time.)

The issue title, fwiw, is about making better use of AB alumni and taking advantage of their institutional knowledge and perspective.

tantek (approved):
LGTM for documenting existing practice (making reality more discoverable), and also +1 to @fantasai's improvements.

frivoal:
As per the minutes of the last call [above](https://github.com/w3c/process/pull/912#issuecomment-2402519980), the reasons for proposing to reject this PR are:

The Team already has general authority to manage communications, including mailing lists, how they're set up, and who has access to them. They're using this authority appropriately (and in this case did write an explicit policy about how that particular mailing list works). The Process is already often decried as being too long, so we shouldn't add things that aren't necessary. Besides, having the Process confer to the Team the right to grant access to certain people to this mailing list based on a written policy may imply that they would not have that right otherwise, and it's not obvious that this is a useful precedent to set. The Process does not need to manage every aspect of the Team's functions.

koalie:
> As per the minutes of the last call [above](https://github.com/w3c/process/pull/912#issuecomment-2402519980), the reasons for proposing to reject this PR are:
> 
> The Team already has general authority to manage communications, including mailing lists, how they're set up, and who has access to them. They're using this authority appropriately (and in this case did write an explicit policy about how that particular mailing list works). The Process is already often decried as being too long, so we shouldn't add things that aren't necessary. Besides, having the Process confer to the Team the right to grant access to certain people to this mailing list based on a written policy may imply that they would not have that right otherwise, and it's not obvious that this is a useful precedent to set. The Process does not need to manage every aspect of the Team's functions.

What @frivoal wrote.

Issue #921, opened 2024-10-01
Forbid holding multiple leadership positions


martinthomson:
It would appear that the process permits a single individual to hold multiple positions in leadership.  A position is not [vacated](https://www.w3.org/policies/process/drafts/#AB-TAG-vacated) when someone is appointed to a second (or third) leadership position.

This is not a feature, but a process vulnerability.  Holding multiple positions vests too much power in a single individual.

The additional demands on the time of the individual who serves in multiple position is a problem, albeit a lesser one.

## Recommendation

I recommend that we modify the Process to expressly prohibit anyone from concurrently holding:

* A W3C Board Director position.
* An AB member position.
* A TAG member position.
* Employment with the W3C or any of its affiliate organizations or contractors.

Should a person be elected or appointed to a position that creates a conflict, they should be required to resign from one or other position.  As the Process only governs AB and TAG membership, it can only insist on resignation from one of those positions.

## Regarding W3C Bylaws

The [bylaws](https://www.w3.org/2022/10/w3c-bylaws-20221019.pdf) of the W3C also do not include any such provision.  However, that can be addressed in Process, avoiding the need for a bylaws revision.  Such a revision would still be welcome, but does not appear to be essential.

The last point forbids members of the Team from holding leadership positions other than those roles that are expressly specified as being for the Team.  The same restriction on Board members might be covered in Delaware law, which might avoid a bylaws revision.  If not, that should be taken up there.

chrisn:
Speaking as AC rep, and *not* as a current BoD member, I fully agree. But note that the Process does already [prevent](https://www.w3.org/policies/process/drafts/#AB-TAG-constraints) the same person being on both the AB and TAG.

chaals:
I'm not sure about restricting positions for people who are employed by W3C (i.e. I don't yet have a firm opinion either way and think it's worth discussing and making a conscious decision).

I agree that there is an important issue here, and I agree that an individual holding a formal position as a member of more than one of the three elected bodies is a problem. In order to allow for communication, it helps to have people liaising between them. Like the custom of giving ex-AB members access to the W3C AC mailing lists, this already provides those individuals with a greater ability than anyone outside the bodies' formal members to influence the relevant body.

[...]

PR #928, opened 2024-10-08
Require resignation from AB/TAG/BoD as a condition of service

martinthomson:
This is part 1 of a solution for #921.
This expands the restriction against concurrent AB/TAG participation to include the BoD.
It does so in a bit of an awkward way,
because the Process cannot tell the BoD how to operate. It is therefore double-ended:

1. If you are appointed to any listed body, you need to resign from the AB or TAG.
2. If you are appointed to the AB or TAG, you cannot accept unless you resign from the other body.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1118,7 +1118,14 @@ Elected Groups Participation Constraints</h5>
 			the <a href="#random">verifiable random selection procedure</a> described below
 			to choose one person for continued participation and declare the other seat(s) vacant.
 		<li>
-			An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.
+			The same individual <em class="rfc2119">must not</em> participate on any two of:
+			the TAG,
+			the AB,
+			or the [=W3C Board of Directors=].
+			An individual who is appointed or selected for AB or TAG,
+			<em class=rfc2119>must</em> either <a href=#resignation>resign</a> from any other position
+			on the other listed bodies or
+			refuse the position.
 	</ul>
 
 <h5 id="AB-TAG-elections">


index.bs
```diff
@@ -1118,7 +1118,14 @@ Elected Groups Participation Constraints</h5>
...
 		<li>
-			An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.
+			The same individual <em class="rfc2119">must not</em> participate on any two of:
+			the TAG,
+			the AB,
+			or the [=W3C Board of Directors=].
+			An individual who is appointed or selected for AB or TAG,
+			<em class=rfc2119>must</em> either <a href=#resignation>resign</a> from any other position
+			on the other listed bodies or
+			refuse the position.
```
TallTed, line 1128:
```suggestion
			The same individual <em class="rfc2119">must not</em> simultaneously serve on any two of
			the TAG,
			the AB,
			or the [=W3C Board of Directors=].
			An individual who is appointed or selected for the AB or TAG
			<em class=rfc2119>must</em> either <a href=#resignation>resign</a> from any position
			on the other listed bodies or
			refuse the new position.
```

index.bs
```diff
@@ -1118,7 +1118,14 @@ Elected Groups Participation Constraints</h5>
 			the <a href="#random">verifiable random selection procedure</a> described below
 			to choose one person for continued participation and declare the other seat(s) vacant.
 		<li>
-			An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.
+			The same individual <em class="rfc2119">must not</em> simultaneously serve on any two of
```
martinthomson, line 1121:
```suggestion
			The same individual <em class="rfc2119">must not</em> concurrently serve on any two of
```
The two are synonyms, but simultaneity implies instantaneous overlap, whereas concurrent is looser.

cwilso:
I would like to go on the record that I oppose this condition.

index.bs
TallTed: OK by me.

PR #929, opened 2024-10-08
Forbid the Team from AB/TAG service

martinthomson:
The Team already has a clearly defined role in the process. Being responsible for appointments to the TAG, for instance, would make service on the TAG a bit of a conflict. Another example is the role of the Team in Councils, for which clearer rules would be clarifying.

I believe that this was a bit of a problem in the distant past, but during my involvement with the organization,
this has never come up.
This is offered as a clarification,
not as a reaction to a particular situation.

This is part 2 of the recommendations in #921.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/w3c-process/pull/929.html" title="Last updated on Oct 8, 2024, 3:40 AM UTC (2f862fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/929/c165854...martinthomson:2f862fb.html" title="Last updated on Oct 8, 2024, 3:40 AM UTC (2f862fb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1119,6 +1119,11 @@ Elected Groups Participation Constraints</h5>
 			to choose one person for continued participation and declare the other seat(s) vacant.
 		<li>
 			An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.
+		<li>
+			Members of the [=Team=] <em class=rfc2119>must not</em> participate on the AB or TAG,
+			except as defined by this process.
+			In particular, the role of Team Contact is defined for both
+			[[#ABParticipation|AB]] and [[#tag-participation|TAG]].
 	</ul>
 
 <h5 id="AB-TAG-elections">


tantek (approved):
I believe this reflects current policy, and setting these expectations explicitly in the Process may help build more trust.

cwilso (changes requested):
I really think this wording is wrong.  This should not be "the Team must not participate" - it should focus on "being on the Team is incompatible with being an elected or appointed member of".  The way this pr is worded, team members could not be invited to participate in meetings in any way except as perhaps Team contact.

(I am fine with the core ask here, I think, just don't think this is how to express is.  Will try to work up a suggested change later this week.

dwsinger:
I think we have to think of a situation where this rule would be needed. At the moment, the team wouldn't think of running, and we wouldn't think of voting for them. To be useful, you have to conceive of an instance where they think they should run and the electorate thinks that they should be elected, and we need a rule to stop that happening. Can anyone?

cwilso:
Many unthinkable things have happened in the political worlds in recent years, so I'm not sure I'm against thinking it through.  But I don't think the roles of the AB, TAG or Board should be filled by Team members - they should be working together - but I do think that we need to navigate carefully because I do think there are some roles that COULD be filled by Team members in some extraordinary circumstances (e.g. Chair of a CG or WG).

I do think it's really important to not characterize this prohibition as "cannot participate", though.

index.bs
```diff
@@ -1119,6 +1119,11 @@ Elected Groups Participation Constraints</h5>
 			to choose one person for continued participation and declare the other seat(s) vacant.
 		<li>
 			An individual <em class="rfc2119">must not</em> participate on both the TAG and the AB.
+		<li>
+			Members of the [=Team=] <em class=rfc2119>must not</em> participate on the AB or TAG,
+			except as defined by this process.
+			In particular, the role of Team Contact is defined for both
+			[[#ABParticipation|AB]] and [[#tag-participation|TAG]].
```
martinthomson, line 1126:
```suggestion
			Members of the [=Team=] <em class=rfc2119>must not</em> be elected or appointed as an AB or TAG participant.
```

I *think* that this addresses @cwilso's concern about word choice.  My choice of "participate" was intended to evoke "participant", rather than the natural meaning of the word.  That's unfortunate.

With the rewording, the note isn't really necessary.

martinthomson (commented):
In many ways, the Process sets up the Team, TAG, and AB as checks on the other bodies.  Having one person hold a position in both is just incompatible with the basic design.

I think that the most plausible bad situation is where a TAG/AB participant takes a job with the W3C such that they then join the Team.

A unreasonable person in that position might be able to exercise executive Team powers (TAG appointments in particular, but also the special function that the Team has for Councils and other procedural matters) to abuse that position.

A reasonable person would recuse as appropriate, but that would mean relying on the reasonableness of every person.  That's something that institutions are designed to counteract: though their formation might depend on exemplary actions of exceptional individuals, their long-term sustainability depends on the strength of structures, like process and culture.

dwsinger:
Ah, OK, agreed, that a change of affiliation to the team should cause something to happen.

fantasai:
@dwsinger Something already does happen: we trigger an affiliation change election. I think that’s good enough for the change of affiliation case.

index.bs
chaals: I agree that this change addresses a bunch of the comments.

brentzundel:
Closing this PR in accordance with the conversation during the lost Process CG meeting with the intention that conversation on this topic continue in Issue #921

Issue #902, opened 2024-07-16
Open up Registries to IG, TAG, AB


frivoal:
Currently, I believe only Working Groups can issue registries. Given that Registries do not contain normative statements and are not subject to the patent policy, I wonder if we could also allow IGs, the AB, and TAG to issue registries as well. I don't really see any reason to restrict them to WGs.

The Process changes to achieve that would be minimal.

For instance, I was wondering if the list of reusable boilerplate text for charters (as discussed in https://github.com/w3c/process/issues/425) could be an AB-maintained repo.

dwsinger:
I don't see how, as the _definition_ of a registry _is_ confined to a rec and hence WGs. But anyone can then _operate_ a registry.

I don't see repos of useful stuff as being a Registry in the formal sense.

frivoal:
> [Registries](https://www.w3.org/policies/process/#registry) can be published either as a stand-alone [technical report](https://www.w3.org/policies/process/#technical-report) on the [Registry Track](https://www.w3.org/policies/process/#registry-track) called a registry report, or incorporated as part of a [Recommendation](https://www.w3.org/policies/process/#RecsW3C) as a registry section.

frivoal:
The registry definition is handled in a way that is similar to a REC, and can be hosted in a REC, but doesn't have to.
https://www.w3.org/policies/process/#reg-pub

dwsinger:
I am sure we decided that Registry Definitions must be in a Technical Report; indeed that seems cited but not stated in the Process.

dwsinger:
We wanted AC approval of the definition of a Registry

frivoal:
> I don't see how, as the definition of a registry is confined to a rec and hence WGs.

> I am sure we decided that Registry Definitions must be in a Technical Report

> We wanted AC approval of the definition of a Registry

Yes, the definitions must be in a technical report, and yes the AC does get to approve the definitions of a registry. However this technical report can *either* be a REC (which get us the requirement for AC review) *or* use the "registry report" type of technical report, created for that purpose, which has equivalent requirements, including the AC review. Some early iterations of the registry process did not have this alternative, but the one we landed on does. I

So, while we cannot enable groups other than WGs to do Registries-in-a-REC since only WGs can do RECs, we could enable other groups to do Registries-in-a-Registry-Report.

> I don't see repos of useful stuff as being a Registry in the formal sense.

It'd be narrower than that:
[...]

PR #972, opened 2025-01-08
Extend the ability to publish Registries to IG/AB/TAG

frivoal:
See #902

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/972.html" title="Last updated on Jan 20, 2025, 7:19 AM UTC (5aa1dea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/972/7ce45be...frivoal:5aa1dea.html" title="Last updated on Jan 20, 2025, 7:19 AM UTC (5aa1dea)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4814,7 +4814,7 @@ Updating Registry Tables</h4>
 	(i.e. [[#correction-classes|Class 5 changes]])
 	can be made by re-publishing the [=technical report=] that contains the affected table,
 	without needing to satisfy any other requirements for the publication
-	(not even consensus, unless this is required by the [=registry definition=]).
+	(not even a [=group decision=], unless this is required by the [=registry definition=]).
 	Such [=registry changes=] do not trigger new [=Advisory Committee Reviews=],
 	nor Exclusion Opportunities,
 	and do not require verification via an [=update request=],
--- a/index.bs
+++ b/index.bs
@@ -3015,14 +3015,14 @@ Types of Technical Reports</h3>
 
 		<dt>Registries
 		<dd>
-			[=Working Groups=] can also publish [=registries=]
+			[=Chartered groups|chartered=] and [=elected groups=] can also publish [=registries=]
 			in order to document collections of values or other data.
 			A registry can be published either as a distinct [=registry report=],
 			or directly within a [=Recommendation Track=] document
 			as an [=embedded registry=].
 			[=registry definition|Defining a registry=] requires [=wide review=] and [=consensus=],
 			but once set up, changes to registry entries are lightweight
-			and can even be done without a [=Working Group=].
+			and can even be done without the initiating group's involvement.
 			See [[#registries]] for details.
 	</dl>
 
@@ -4702,7 +4702,7 @@ Registry Definitions</h4>
 			and which is responsible for evaluating whether such requests
 			satisfy the criteria defined in the [=registry definition=].
 
-			The [=custodian=] may be the [=Working Group=], the [=Team=], or a delegated entity.
+			The [=custodian=] may be the initiating [=chartered group|chartered=] or [=elected group=], the [=Team=], or a delegated entity.
 			The [=custodian=] for all [=registry tables=] in a single [=registry=]
 			<em class=rfc2119>should</em> generally be the same entity.
 
@@ -4710,7 +4710,7 @@ Registry Definitions</h4>
 			ceases to exist or to operate as a custodian
 			(e.g., the relevant group is disbanded, or 
 			the custodian is unresponsive to repeated attempts to make contact),
-			and the [=Working Group=] that owns the [=registry definition=] 
+			and the [=chartered group|chartered=] or [=elected group=] that owns the [=registry definition=]
 			is itself <a href="#GeneralTermination">closed</a> or 
 			unresponsive,
 			the [=Team=] <em class=rfc2119>should</em> propose replacing the [=custodian=],
@@ -4721,9 +4721,9 @@ Registry Definitions</h4>
 <h4 id=reg-pub>
 Publishing Registries</h4>
 
-	[=Registries=] can be published either
+	[=Registries=] can be published by [=chartered group|chartered=] or [=elected groups=] either
 	as a stand-alone [=technical report=] on the [=Registry Track=] called a <dfn>registry report</dfn>,
-	or incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
+	or, in the case of [=Working Groups=], incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
 
 	The [=registry report=] or [=embedded registry=] <em class=rfc2119>must</em>:
 	<ul>
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
 	(i.e. [[#correction-classes|Class 5 changes]])
 	can be made by re-publishing the [=technical report=] that contains the affected table,
 	without needing to satisfy any other requirements for the publication
-	(not even Working Group consensus, unless this is required by the [=registry definition=]).
+	(not even consensus, unless this is required by the [=registry definition=]).
 	Such [=registry changes=] do not trigger new [=Advisory Committee Reviews=],
 	nor Exclusion Opportunities,
 	and do not require verification via an [=update request=],
 	even for [=technical reports=] at maturities where this would normally be expected.
 	Such publications can be made
-	even in the absence of a [=Working Group=] chartered to maintain the registry
+	even in the absence of a group chartered to maintain the registry
 	when the [=custodian=] is another entity.
 
 	Note: The custodian is only empowered to make [=registry changes=].
-	If the Working Group establishing the registry wishes
+	If the [=chartered group|chartered=] or [=elected group=] establishing the registry wishes
 	to empower the custodian to add commentary on individual entries,
 	this needs to be part of the registry table’s definition.
 	If other changes are desired,
-	they need to be requested of the responsible Working Group--
-	or in the absence of a Working Group, of the Team.
+	they need to be requested of the [=chartered group|chartered=] or [=elected group=] responsible for maintaining the [=registry definition=]--
+	or in the absence of such a group, of the Team.
 
 	Changes to the [=registry tables=]
 	made in accordance with [=candidate amendment|candidate=] or [=proposed amendments=] to the [=registry definition=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -3015,14 +3015,14 @@ Types of Technical Reports</h3>
 
 		<dt>Registries
 		<dd>
-			[=Working Groups=] can also publish [=registries=]
+			[=Chartered groups|chartered=] and [=elected groups=] can also publish [=registries=]
```
TallTed, line 3018:
```suggestion
			[=Chartered groups|Chartered=] and [=elected groups=] can also publish [=registries=]
```

TallTed (changes requested):
Tiny typo

index.bs
```diff
@@ -4702,15 +4702,15 @@ Registry Definitions</h4>
 			and which is responsible for evaluating whether such requests
 			satisfy the criteria defined in the [=registry definition=].
 
-			The [=custodian=] may be the [=Working Group=], the [=Team=], or a delegated entity.
+			The [=custodian=] may be the initiating [=chartered group|chartered=] or [=elected group=], the [=Team=], or a delegated entity.
```
fantasai, line 4705: I think we can simplify this to "[=group=]", no?

index.bs
```diff
@@ -4721,9 +4721,9 @@ Registry Definitions</h4>
 <h4 id=reg-pub>
 Publishing Registries</h4>
 
-	[=Registries=] can be published either
+	[=Registries=] can be published by [=chartered group|chartered=] or [=elected groups=] either
```
fantasai, line 4724: I think we can just delete the "by" phrase, it's not adding anything.

index.bs
```diff
@@ -4721,9 +4721,9 @@ Registry Definitions</h4>
...
 	as a stand-alone [=technical report=] on the [=Registry Track=] called a <dfn>registry report</dfn>,
-	or incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
+	or, in the case of [=Working Groups=], incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
```
fantasai, line 4726:
```suggestion
	or, in the case of those owned by [=Working Groups=],
	incorporated as part of a [=Recommendation=] as an <dfn oldids="registry-section">embedded registry</dfn>.
```

index.bs
```diff
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
 	(i.e. [[#correction-classes|Class 5 changes]])
 	can be made by re-publishing the [=technical report=] that contains the affected table,
 	without needing to satisfy any other requirements for the publication
-	(not even Working Group consensus, unless this is required by the [=registry definition=]).
+	(not even consensus, unless this is required by the [=registry definition=]).
```
fantasai, line 4817:
```suggestion
	(not even a [=group decision=], unless this is required by the [=registry definition=]).
```

index.bs
```diff
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
...
 	Such [=registry changes=] do not trigger new [=Advisory Committee Reviews=],
 	nor Exclusion Opportunities,
 	and do not require verification via an [=update request=],
 	even for [=technical reports=] at maturities where this would normally be expected.
 	Such publications can be made
-	even in the absence of a [=Working Group=] chartered to maintain the registry
+	even in the absence of a group chartered to maintain the registry
```
fantasai, line 4823:
```suggestion
	even in the absence of a [=group=] chartered to maintain the registry
```

index.bs
```diff
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
...
 	when the [=custodian=] is another entity.
 
 	Note: The custodian is only empowered to make [=registry changes=].
-	If the Working Group establishing the registry wishes
+	If the [=chartered group|chartered=] or [=elected group=] establishing the registry wishes
```
fantasai, line 4827:
```suggestion
	If the [=group=] establishing the registry wishes
```

index.bs
fantasai: And so on throughout.

index.bs
frivoal: `[=group=]` on its own links to a definition that also includes the AC. I don't think the AC should be in the business of publishing anything

index.bs
frivoal: To me, this seems like the place in the Process which establishes who can publish a Registry. The alternative is line 3018, but that feels more like a reference to something established elsewhere than the place that establishes it.

index.bs
frivoal: Indirectly, I think which phrasing we pick here kind of depends on what we do on line 4727: depending on how we right the two parts of this contrast, we might suggest that doing a registry in a REC is the only way for WGs to do it, while we actually want to say that only WGs have the ability to do so, but that they can also do it standalone like everyone else.

index.bs
frivoal: `[=groups=]` also refers to the AB, TAG, and AC, which I don't think is appropriate in the context of this sentense

index.bs
frivoal: Ok with dropping to group, but I don't think the hyperlinking helps

tantek (approved):
Seems like a reasonable improvement, and I agree with the replies from @frivoal to suggested changes.

index.bs
```diff
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
...
 	to empower the custodian to add commentary on individual entries,
 	this needs to be part of the registry table’s definition.
 	If other changes are desired,
-	they need to be requested of the responsible Working Group--
-	or in the absence of a Working Group, of the Team.
+	they need to be requested of the [=chartered group|chartered=] or [=elected group=] responsible for maintaining the [=registry definition=]--
```
nigelmegitt, line 4831:
```suggestion
	they need to be requested of the group responsible for maintaining the [=registry definition=]--
```

index.bs
```diff
@@ -4859,12 +4859,13 @@ Registry Data Reports</h4>
 	is that of the [=technical report=] holding the [=registry definition=].
 
 	Anytime a change is made to a [=registry definition=],
-	the Working Group <em class=rfc2119>must</em> update and republish
+	the [=chartered group|chartered=] or [=elected group=] maintaining the [=registry definition=]
```
nigelmegitt, line 4862:
```suggestion
	the group maintaining the [=registry definition=]
```

index.bs
```diff
@@ -4859,12 +4859,13 @@ Registry Data Reports</h4>
...
+	<em class=rfc2119>must</em> update and republish
 	any document holding the corresponding [=registry tables=]
 	to make it consistent with these changes.
 
 	Given a recorded [=group decision=] to do so,
-	the [=Working Group=]
+	the [=chartered group|chartered=] or [=elected group=] maintaining the [=registry definition=]
```
nigelmegitt, line 4868:
```suggestion
	the group maintaining the [=registry definition=]
```

nigelmegitt (commented):
Looks good to me, have made suggestions to remove probably unnecessary wording.

index.bs
```diff
@@ -4814,22 +4814,22 @@ Updating Registry Tables</h4>
...
 	and do not require verification via an [=update request=],
 	even for [=technical reports=] at maturities where this would normally be expected.
 	Such publications can be made
-	even in the absence of a [=Working Group=] chartered to maintain the registry
+	even in the absence of a group chartered to maintain the registry
 	when the [=custodian=] is another entity.
 
 	Note: The custodian is only empowered to make [=registry changes=].
-	If the Working Group establishing the registry wishes
+	If the [=chartered group|chartered=] or [=elected group=] establishing the registry wishes
```
frivoal, line 4827: So I'll go with this:
```suggestion
	If the group establishing the registry wishes
```

index.bs
frivoal: So I'll go with this:
```suggestion
	If the group establishing the registry wishes
```

index.bs
frivoal, line 4705:
```suggestion
			The [=custodian=] may be the initiating group, the [=Team=], or a delegated entity.
```

index.bs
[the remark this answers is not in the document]
frivoal: I think I like this option best. @fantasai , @TallTed , what do you think? I'd like to take this rephrasing instead of both https://github.com/w3c/process/pull/972/files#r1908141004 and https://github.com/w3c/process/pull/972/files#r1908141327

index.bs
frivoal: @fantasai , are you OK with closing this for the reason given above?

index.bs
frivoal: @fantasai , are you OK with closing this for the reason given above?

index.bs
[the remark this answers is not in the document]
TallTed: Getting there. I suggest some tweaks --
```suggestion
    A [=chartered group|chartered=] or [=elected group=] can publish a [=registry=]
    as a stand-alone [=technical report=] on the [=Registry Track=],
    where it is called a <dfn>registry report</dfn>.
    [=Working Groups=] have the additional option of
    incorporating a [=registry=] as part of a [=Recommendation=],
    as an <dfn oldids="registry-section">embedded registry</dfn>.
```

index.bs
TallTed: Can a Community Group charter not include maintaining a registry?

frivoal: Community groups are not covered by the process at all, consensus doesn't apply (at least not necessarily, though they can opt into it), AC reviews don't apply, charters can be re-written on the fly without external validation… In general, CGs don't do any publication of any kind of technical report.

Now, if some other group has established a registry, and got the registry definitions properly approved, those definitions can designate anyone to be a custodian, and following the rules laid out in the definitions, make updates to the registry data. "Anyone" can be a CG. But a CG itself cannot establish the registry.

nigelmegitt: Does it matter if a custodian is a group with a Charter but the Charter doesn't mention the custodianship role? I would say it's a "nice to have" that the Charter mentions it, but not essential.

frivoal: I would agree it doesn't matter, but I'm not sure why that's relevant.

nigelmegitt: The only relevance is: if we thought it _is_ important that custodians with charters include their custodianship in their charter, then I would propose opening an additional issue to address that; since we seem to be in agreement that it is not important, then I would suggest that we don't need to do anything in response to @TallTed 's [question](https://github.com/w3c/process/pull/972#discussion_r1924307185). But @TallTed may have been going somewhere else with that, I don't know.

Issue #982, opened 2025-01-22
Abandoning Charter Refinement should require Team to agree


fantasai:
The current draft requires only a Chair Decision from a facilitator to abandon efforts on a new charter. During internal review at Apple, it was asked if this decision shouldn't also consider the Team's position.

Suggestion: Require both a chair (chartering facilitator) decision and a Team decision in order to abandon the chartering effort.

frivoal:
Basically I agree. In terms of how to describe it, I'm not 100% sure about whether to require a double decision as you suggest, or to make it a proposal by the facilitator + a decision by the Team. But then again, maybe this is a distinction without a difference.

frivoal:
#983, which solves this issue, was accepted and merged a while back, which resolves this issue. Closing.

PR #983, opened 2025-01-23
Require decision by both the facilitator and team to abandon chartering

frivoal:
See #982

In terms of phrasing, I opted to keep the idea of a decision by the facilitator, confirmed by a decision of the team, rather than something like a team decision to accept a proposal by facilitator, because hooking into the notion of Chair Decision gives us an explicit statement that this is something they do in the own judgement, not based on the consensus of the group. And given that a key reason for giving up on a charter could be that the relevant community is proving incapable of consensus (which could possibly even extend to the question of whether to give up), that seems relevant.


index.bs
```diff
@@ -1775,7 +1775,7 @@ Charter Refinement</h3>
 	The [=charter refinement=] phase concludes when there is either:
```
TallTed, line 1775:
```suggestion
	The [=charter refinement=] phase concludes when there is either of the following:
```

PR #986, opened 2025-02-12
Enable AB/TAG to remove one of their members

frivoal:
See #882

Though they serve related purposes, individual removal by members of the same body and recall of the entire body by its electorate are procedurally distinct. Discussions in #888, which attempted to do both, show debates about each part. In particular, #888 has many comments about various part of collective removal.

To simplify discussion, this pull request isolates the individual removal by members of the same body, since it does not depend on the rest. Refer also to the comment thread at https://github.com/w3c/process/pull/888#discussion_r1936635552, which discussed this aspect in #888.

Discussion of collective removal can continue in #888

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1322,14 +1322,37 @@ Verifiable Random Selection Procedure</h5>
 			The incomplete terms are assigned in result order.
 	</ol>
 
+<h5 id="AB-TAG-removal">
+Removing AB or TAG members</h5>
+
+	Individual participants of the [=Advisory Board=] or the [=Technical Architecture Group=]
+	can be <dfn export lt="remove|removal">removed</dfn> from those groups
+	if they are found by their peers
+	to be grossly neglecting their duties,
+	or to be acting in a way that seriously hampers the group's ability to function normally.
+
+	A chair of the [=AB=] or [=TAG=] <em class=rfc2119>must</em> hold a hearing
+	on the potential [=removal=] of a participant
+	if requested by at least three of the participants in the group.
+	After giving the individual in question
+	an opportunity to defend themselves,
+	a vote on the proposed [=removal=] is held.
+	If at least three quarters of the participants in the group,
+	excluding the individual who is the subject of such vote,
+	then vote in favor of the proposal,
+	the individual's seat on [=AB=] or [=TAG=] is [=vacated=] immediately.
+
 <h5 id="AB-TAG-vacated">
 Elected Groups Vacated Seats</h5>
 
-	An [=Advisory Board=] or [=TAG=] participant's seat is vacated when:
+	An [=Advisory Board=] or [=TAG=] participant's seat is <dfn lt="vacated|vacant">vacated</dfn> when:
 
 	<ul>
 		<li>
 			the participant resigns, or
+
+		<li>
+			the participant is [=removed=], or
 
 		<li>
 			an Advisory Board or TAG participant changes affiliations


swickr:
Two concerns:
1. Process is explicit that the [Advisory Board does not have decision-making authority](https://www.w3.org/policies/process/#ac-role).
2. A primary role of the AB is to provide [ongoing guidance to the Team on issues of strategy, management, legal matters, process, and conflict resolution](provides ongoing guidance to the Team on issues of strategy, management, legal matters, process, and conflict resolution.)

The combination of these two lead me to conclude that the authority of the CEO to remove an elected member is sufficient. The CEO is always entitled to solicit advice from any set of individuals and a duly elected member of the AB who obstructs the progress of that body need not become a showstopper such that Process has to grant special decision-making authority to the AB.

index.bs
```diff
@@ -1322,15 +1322,38 @@ Verifiable Random Selection Procedure</h5>
...
+	A chair of the [=AB=] or [=TAG=] <em class=rfc2119>must</em> hold a hearing
+	on the potential [=removal=] of a participant
+	if requested by at least three of the participants in the group.
+	After giving the individual in question
+	an opportunity to defend themselves,
+	a vote on the proposed [=removal=] is held.
+	If at least three quarters of the participants in the group,
+	excluding the individual who is the subject of such vote,
+	then vote in favor of the proposal,
+	the individual's seat on [=AB=] or [=TAG=] is [=vacated=] immediately.
```
nigelmegitt, line 1343: Is "holding a different opinion" or "preventing a Council from reaching a decision by vote" potentially something that could be considered "hampering the group's ability to function"?

If so, I'd be (a tiny bit) concerned about an _in extremis_ potential abuse of this mechanism to evict an AB or TAG member in order to remove them from a Council (assuming that the opportunity for [dismissal](https://www.w3.org/policies/process/#dismissed) had passed) in order to unlock a Council vote.

It's complex to do the maths on this but I suspect that there could be a scenario in which a majority of a Council could not be found but a three-quarters majority of the AB or the TAG could be found to remove one of them.

For practical purposes, this is a highly theoretical, i.e. unlikely to ever happen, matter of minutia, so doesn't seem like reasonable grounds for blocking the PR, but it would be good to know if anyone did work through the voting numbers given this three quarter majority requirement, and other contexts in which AB or TAG might participate in votes.

wareid: @nigelmegitt I'd refer you to the documentation on Councils, as the situation you describe would not really make sense in the context of how councils are run. It's entirely possible for a council to proceed if someone (or multiple people but not enough to amount to overturning a vote) holds a minority opinion. You can find the details here: https://www.w3.org/policies/process/20231103/#council-deliberations

If someone decided to be so disruptive to any AB/TAG process, council or otherwise, that the bodies couldn't conduct their expected responsibilities, that seems to be a valid premise to initiate this procedure. For it to pass, there is still the requirement of 3/4 of the group voting in favour, which is a significant bar. I'd add that in the case where someone is this disruptive, there is an impact on the membership as well, even if the membership did vote for them. 

The minutes for today aren't available yet, but I will share what I said in support of this change: this is an edge case, and hopefully the most edgy of cases. There are a million steps that come before a decision like this, and I hope we never have to use it, but it is good to have it in the situation where we need it.

frivoal: Another observation:
> Membership of each [Council](https://www.w3.org/policies/process/#w3c-council) instance is fixed at formation,

Making someone not be a AB/TAG member anymore doesn't remove them from the council. Now, it would possible to form a new council without the excluded individual(s) by using the following provision:
> However, if participation in a [Council](https://www.w3.org/policies/process/#w3c-council) falls so low as to hinder effective and balanced deliberations, the [W3C Council Chair](https://www.w3.org/policies/process/#w3c-council-chair) should dissolve the [Council](https://www.w3.org/policies/process/#w3c-council) and call for a new one to be convened.

but if that's just to get rid of a person or two, that wouldn't be a case of "falls so low as to hinder effective and balanced deliberations", and would be pretty solid grounds for escalation / appeal.

index.bs
```diff
@@ -1322,15 +1322,38 @@ Verifiable Random Selection Procedure</h5>
...
+<h5 id="AB-TAG-removal">
+Removing AB or TAG members</h5>
+
+	Individual participants of the [=Advisory Board=] or the [=Technical Architecture Group=]
+	can be <dfn export lt="remove|removal">removed</dfn> from those groups
+	if they are found by their peers
+	to be grossly neglecting their duties,
+	or to be acting in a way that seriously hampers the group's ability to function normally.
+
+	A chair of the [=AB=] or [=TAG=] <em class=rfc2119>must</em> hold a hearing
```
cwilso, line 1334: There should be explicit detail here about who is invited to that hearing; both who must be invited (I'd expect the relevant body, including the individual) and who is allowed.  Also, whether this is publicized or not.  (I would expect not, but it shouild be explicit that it's confidential, perhaps?

index.bs
```diff
@@ -1322,15 +1322,38 @@ Verifiable Random Selection Procedure</h5>
...
-	An [=Advisory Board=] or [=TAG=] participant's seat is vacated when:
+	An [=Advisory Board=] or [=TAG=] participant's seat is <dfn lt="vacated|vacant">vacated</dfn> when:
 
 	<ul>
 		<li>
 			the participant resigns, or
 
+		<li>
+			the participant is [=removed=], or
+
```
cwilso, line 1356: Is it not the intent that this would replace the ability below (in lines 1362-1366) for the CEO to dismiss the participant?  I would have expected those lines to be removed.

koalie: I strongly object.

It's one thing to create a supplemental mechanism for one body to remove one of their members, to prevent a situation that has yet to happen, as a way to afford multiple paths to resolving problems so you aren't bottlenecked on one single path, and it's another thing entirely to strip the existing authority. This was presented in the 20 March AB meeting as a supplemental mechanism. 

Furthermore, W3C has a Board of Directors who oversees the CEO and their actions.

cwilso: I'm not sure how this was presented as a supplemental mechanism - the minutes haven't been released yet so I can't link to them, but I said "I am def in favor of **switching** this to AB itself managing", and there were several others who I thought agreed.

How is the Board (who have precisely this 3/4-vote-of-its-own-membership-to-eject rule) and their oversight of the CEO relevant?

Or maybe more to the point: in what case would it be critical for the CEO to be able to eject an AB member, without the support of 3/4 of the rest of the AB?

index.bs
fantasai: Agree this is worth clarifying.

frivoal: Let's follow up about this aspect in https://github.com/w3c/process/issues/1006

index.bs
frivoal: created https://github.com/w3c/process/issues/1007 to follow up on that

plehegar:
to make it clean:
- this PR was merged with NO changes to the CEO abilities
- there is a difference of opinions on what was implied during the AB discussion
- this PR was merged based on the recorded of the AB and may get reverted if additional information comes from the AB
- if additional changes have to be made impacting the CEO, it would have be done in a new separate pull request

plehegar:
Note that the TAG did not get asked about this pull request:
[[
Jeffrey: other comment I heard is that AB should have waited for [TAG's] review before merging this.

consensus on this point
]]
https://github.com/w3ctag/meetings/blob/gh-pages/2025/03-31-minutes.md#ab-decision-process

Bruno-bot-stack (commented):
<html>
<head>
    <title>
    The title of your page
    </title>
</head>

<body>
    Your page content goes here.
</body>
</html>

index.bs
amyvdh: On the idea of elected bodies removing their own members, Team noted at the AB meeting and above, the conflict inherent in and pushed back on the idea that elected members of the AB would give themselves the power to remove other elected members. 

In the Process under Role of the Advisory Board, it states: "The Advisory Board also serves the Members by…”  - serves. I have also recently heard several AB members talk about representing the Members. I am not certain it would be serving or representing Members to undo Member election choices. Some AB members have even said they feel a certain mandate in being elected by Members. Wouldn’t the AB giving themselves the power to remove another member be in opposition to this choice or mandate of the Members if so? I feel there is thus an element of contradiction and conflict inherent here.

This kind of voting would also mean that a small group of members (also potential business competitors), meant to work together over years for the good of the web, would need to also not just negotiate amongst each other for projects and priorities but also be willing and able to build, bring a case against and move to remove one of their body from a powerful, possibly career advancing position. The conflict of interest and potential for factionalism, threat, distrust and disruption could be immense.

amyvdh: cultural differences and liability 
Our Members come from a wide range of organizations but we know, often from business or global competitors. Our members also come from very different cultural and geographic backgrounds. How could the work or contributions of fellow members be judged by each other fairly when everyone comes with different skills or a different sense of duty? For fellow Members to remove someone from their elected role might open them to accusations of bias or competitor or even national retaliation. 

It is because this that it is possible that someone who is removed by other members might even consider taking legal action for something like bias or career harm so there is also an issue of liability.  

The neutral role of the CEO, as the Process currently assigns the responsibility, does mean questions of bias are greatly reduced (no burden of re-election, not a business competitor, acting for the W3C as a whole, not paid by one company, etc.).

amyvdh: I strongly feel this is a radical, contentious and weighty proposal for a problem that has not happened.

No CEO has yet needed to remove an AB or TAG member so the problem this proposal seeks to address has not actually happened. However, we know that there are longstanding and repeated examples of business competition within W3C or attempts to stop or slow our work. Any new ability for factions to be able to block or remove potential competitors or to prevent differing interests from also serving on elected bodies should be tread very carefully. New methods which would make our governing bodies more contentious, factional or political should absolutely be avoided. Creating new processes to overturn elected decisions of Members in an increasingly disengaged Membership also seems injudicious at this time. 

I would suggest that any addition to allow AB to vote on removing other AB members should be added to the Process, only after it has been shown 1. to be a problem that has actively happened that needs such a heavy solution and 2. with the very clear and direct support of Members for this removal their elected choices and 3. after a review which can identify the problem, what worked and what didn’t, legal risks (eg: wrongful termination or career harm) etc to find the best solutions- not before.

amyvdh: Finally, and I think most concerning, there should be great care taken in using access to the levers of power. Judges and those in power are expected to recuse themselves from conflicts of interest. In other countries it might be said as a group "must not be judge and party". The AB has immense power due to their access to W3C Process through control of the Process CG. 

It seems to me, any proposal by the AB to use it’s access to and functional control over the Process to increase its own power, change its role, or to suggest any adjustment which removes power from any other group or person must absolutely be done rarely, addressed with great seriousness and I would hope, great prudence. I do not see a compelling need for either adding to AB power (moving from "advising" to "deciding" in a more Board-like role) nor to remove from them, a CEO power which impacts all other Groups. 

In this instance, there seems to be great pressure to impose this change by the next revision.  I feel this the benefits of this change are not at all demonstrated as urgently needed.  This power has never been used, I would question the urgency to change it.

chaals:
I am likewise deeply concerned that giving this power to the AB is a bad 
idea with very little upside (the AB can work around inattentive or 
obstructive members, this is just adding a high-effort alternative 
approach) and a lot of downside.

plehegar:
Update from the TAG:
[[
Resolved by consensus: We're unhappy that this was merged without more extensive consultation. We don't think this removal process should be in the Process without more detail on the reasons for removal, and until there is an appeal process to prevent it from being used to marginalize people, for example an appeal to the AC or a way for the AC to remove the body that inappropriately removed someone. Many of us don't think it should be in the process at all, even with such an appeal process.
]]
https://github.com/w3ctag/meetings/blob/gh-pages/2025/telcons/04-21-minutes.md#recall-procedures-for-tag-and-ab---proposed-by-sarven

chrisn:
I tend to agree with @chaals comment above, and am also sympathetic to the TAG's response.

index.bs
fantasai: To respond briefly to your last questions:
1. One of the questions here isn't whether it should be possible to un-do the election choices of the AC -- we already have that ability in the Process, as the CEO has that power -- but by whom, how, and for what purposes that power is applied.
2. We know there has been at least one instance of such a problem in the past. Further discussion of such cases is not a topic for a public forum, however.
3. Agreed 100% that it should be clearly defined, and to be defined in a way so as to avoid abuse of the power.
4. We were told that the TAG and AB are covered under W3C's insurance policies (since such questions have already come up regarding the Council).

frivoal:
As discussed in https://github.com/w3c/process/issues/882#issuecomment-2879403306, a dedicated topic branch was created for this issue and related ones: [ab-tag-discipline](https://github.com/w3c/process/tree/ab-tag-discipline)
Although this PR had previously been merged into the main branch, this has been reverted. See [ab-tag-discipline](https://github.com/w3c/process/tree/ab-tag-discipline), and [related issues](https://github.com/w3c/process/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22Topic%3A%20AB%2FTAG%20discipline%22) or [PRs](https://github.com/w3c/process/issues?q=is%3Apr%20state%3Aopen%20label%3A%22Topic%3A%20AB%2FTAG%20discipline%22) for more on that topic.

PR #994, opened 2025-03-11
Pick a number of weeks

frivoal:
We merged earlier pull requests leaving "X weeks" as an unspecified number. Make that 8.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/994.html" title="Last updated on Mar 12, 2025, 5:18 AM UTC (305f567)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/994/0d6e9c4...frivoal:305f567.html" title="Last updated on Mar 12, 2025, 5:18 AM UTC (305f567)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1767,7 +1767,7 @@ Initiating Charter Refinement</h3>
 	This rejection is a [=Team Decision=],
 	and can be appealed only by 5 or more [=Members=],
 	through their [=Advisory Committee representative=],
-	formally objecting to the decision within X weeks of the decision being announced.
+	formally objecting to the decision within 8 weeks of the decision being announced.
 	In this case the [=Team=] <em class=rfc2119>must</em> start an [=appeal vote=]
 	on whether to overturn the [=Team Decision=].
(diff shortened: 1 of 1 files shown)


ianbjacobs:
I would much rather it be 4 weeks than 8. The goal is to end up with a process that does not require as much time as a Council process. We use 4 weeks for other charter review processes; why 8 weeks here?

frivoal:
Nothing happens during that time, so whether it's 4, 6, 8, or 13.5 doesn't cost us anything different. I think there should be a bound, so that it doesn't count if 4 people agree in quick succession, and the 5th one comes around 17 years later. But while 4 is fine most of the time, if it overlaps with TPAC, end-of-year holidays, or some other thing that take many people away from work for a long time, 4 feels a bit short. I'm proposing 8 as something reasonably short that works even at times where something else is happening.

ianbjacobs:
> Nothing happens during that time, so whether it's 4, 6, 8, or 13.5 doesn't cost us anything different

This slows down the Consortium. That's my concern.

PR #999, opened 2025-03-12
Drop unnecessary detail

frivoal:
Every email explains what the email is about. We can make the Process a tiny be shorter by not restating the obvious.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/999.html" title="Last updated on Mar 12, 2025, 4:32 AM UTC (9dabf68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/999/0d6e9c4...frivoal:9dabf68.html" title="Last updated on Mar 12, 2025, 4:32 AM UTC (9dabf68)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1749,7 +1749,6 @@ Initiating Charter Refinement</h3>
 	and, in the case of rechartering, to the affected [=Group=].
 
 	This <dfn>charter review notice</dfn> must include:
-	* A short summary of the proposal.
 	* The location of the <dfn>charter draft</dfn>, which must be public.
 	* How to participate in the discussion of this [=charter draft=] and where to file issues.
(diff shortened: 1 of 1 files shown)


fantasai:
-1 to this change. It's a summary of the proposal, not a summary of the fact that we're starting Charter Refinement. I don't think we want the template to say

Charter Refinement for Foo charter

Here's the draft.
Here's where to comment.
This person is facilitating.

And nobody knows what Foo is or whether they should care.

tantek (approved):
I support this change. Requirements in the Process MUST be restricted to ONLY that which is truly required rather than "it would be good if" type things. Leave such things to the Guide please.

PR #1000, opened 2025-03-12
Move details of how the Team evaluates charters to the guide

frivoal:
This is similar to how we move the Team's evaluation criteria for Member submissions to the Guide.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1000.html" title="Last updated on Apr 23, 2025, 3:13 PM UTC (7c152b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1000/cf784bd...frivoal:7c152b3.html" title="Last updated on Apr 23, 2025, 3:13 PM UTC (7c152b3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1763,7 +1763,8 @@ Initiating Charter Refinement</h3>
 	An [=Advisory Committee representative=] <em class=rfc2119>may</em> formally request
 	that the [=Team=] initiate [=charter refinement=].
 	The Team <em class=rfc2119>may</em> deny such a request
-	if it thinks the proposal is insufficiently mature or does not align with W3C's scope and mission,
+	if it thinks the proposal does not meet the charter assessment criteria
+	described in the Guide,
 	and <em class=rfc2119>must</em> reply with its rationale.
 	This rejection is a [=Team Decision=],
 	and can be appealed only by 5 or more [=Members=],
--- a/index.bs
+++ b/index.bs
@@ -1764,7 +1764,7 @@ Initiating Charter Refinement</h3>
 	that the [=Team=] initiate [=charter refinement=].
 	The Team <em class=rfc2119>may</em> deny such a request
 	if it thinks the proposal does not meet the charter assessment criteria
-	described in the Guide,
+	described in the Guide (see [[CHARTER inline]]),
 	and <em class=rfc2119>must</em> reply with its rationale.
 	This rejection is a [=Team Decision=],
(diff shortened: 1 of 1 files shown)


frivoal:
This is the minimalist version of this PR. We could choose to keep a little more in the Process (for instance, for Member submissions we kept the IPR clauses), but this is what it looks like if we move it all to the guide. Of course, we actually need to populate that guide article.

fantasai:
I think we should follow the model we did for [Member Submissions](https://www.w3.org/policies/process/drafts/#Submission), which is to keep the high-level ideas of what the Team should be evaluating in the Process (scope, quality, mission-alignment, etc.), and leave the details to the Guide.

TallTed (approved):

chrisn (approved):

index.bs
```diff
@@ -1763,7 +1763,8 @@ Initiating Charter Refinement</h3>
 	An [=Advisory Committee representative=] <em class=rfc2119>may</em> formally request
 	that the [=Team=] initiate [=charter refinement=].
 	The Team <em class=rfc2119>may</em> deny such a request
-	if it thinks the proposal is insufficiently mature or does not align with W3C's scope and mission,
+	if it thinks the proposal does not meet the charter assessment criteria
```
frivoal, line 1766:
```suggestion
        if it thinks the proposal is insufficiently mature,
        does not align with W3C's scope and mission,
	or otherwise thinks the proposal does not meet the charter assessment criteria
```

Here's a slightly less minimalist version, which still defers the details to the guide, but keeps the high level principles in the Process.

fantasai: Are there charter assessment criteria that don't fit under either of the other two requirements? CC @ianbjacobs @plehegar

ianbjacobs: @fantasai,

Yes. @plehegar and I came up with these others to start:

* Does the charter align with established Web architecture principles?
* Is there sufficient support for the charter?
* Is there significant consensus for the charter? One example of lack of consensus is when there are two competing charters and the community has not yet converged.
* Are there clear signals within the W3C community that the W3C Membership will participate in the work, or that organizations will join to do the work?
* Does the Team have adequate resources for the group in light of current W3C priorities?

We think it is useful to have a list (for predictability), that we will want to make it easy for the list to evolve with practice, and that a long list should not weigh down the process doc.

PR #1001, opened 2025-03-12
Drop wide review of charters from the Process

frivoal:
It's been suggested that this could be left to the guide. This PR implements that possible change.

Personally, I'm not convinced, and I think it is worth hard-coding this expectation into the Process, but I'm offering the PR to anchor discussion of that aspect.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1001.html" title="Last updated on Mar 26, 2025, 2:38 AM UTC (5d41c27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1001/0d6e9c4...frivoal:5d41c27.html" title="Last updated on Mar 26, 2025, 2:38 AM UTC (5d41c27)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1788,7 +1788,6 @@ Charter Refinement</h3>
 	Note: The [=Chartering Facilitator=] is not necessarily the [=Chair=] of the [=chartered group|group=] being chartered.
 
 	During [=charter refinement=]:
-	* [=Wide review=] of the [=charter draft=] is initiated and completed.
 	* All issues filed against the [=charter draft=] must be [=formally addressed=],
 		and their resolutions tracked in a disposition of comments.
 	* Decisions are made as [=group decisions=], where the group is made up of all individuals participating in this process,


tantek (approved):
I am a +1 to this change. Having recently led a wide-review (which I think was the right thing to do), I must acknowledge it takes time, and we should be more flexible for this initial "charter refinement" feature to allow us to use human judgment (guide) and experiment and learn. It's premature to make this kind of requirement on a new Process feature which is completely handled informally currently.

chaals:
I think this change is not a good idea. There has been an expectation of this for many years, and many discussions about how chartering is an important place to be looking at scope. Often these are explaining things that people are unhappy about later, so it seems worthwhile clarifying in advance.

nigelmegitt (commented):
For completeness, this change would need to be expanded to remove  "widely reviewed" from line 1729.

nigelmegitt:
Perhaps it should be theoretically possible for W3C to charter a group without wide review (I have not come to my own conclusion about that yet), but in that case, it surely should not be possible to avoid horizontal review of the proposed charter. 

The minimum change that _might_ therefore be acceptable would be to change "wide review" to "horizontal review", rather than removing the provision altogether.

ianbjacobs:
"Wide review" is in the guide today and it works well.

TallTed:
Sometimes it's good to say similar or identical things in multiple places. I think it is not problematic to include mention of wide review in both Guide and Process documents. Lacking a strong reason to remove it from the Process, i would leave it in both.

PR #997, opened 2025-03-11
Allow Team to step in when consensus cannot be found during Charter Refinement.

fantasai:
Ian's proposal describes

> The Team must seek community [consensus](https://www.w3.org/policies/process/#consensus-building) among those participating in the refinement process, and must [formally address all issues](https://www.w3.org/policies/process/#formal-address) (raised via the advertised mechanism). The Team is responsible for managing situations where the W3C community cannot achieve consensus and supports multiple related (and competing) charters.

This proposal preserves the current mode of the Chartering Facilitator, specifically, acting as the chair in determining the consensus of the community, but allows them to ask the Team to step in when consensus cannot be found.

***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/997.html" title="Last updated on Mar 11, 2025, 4:24 PM UTC (a286c2f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/997/0d6e9c4...fantasai:a286c2f.html" title="Last updated on Mar 11, 2025, 4:24 PM UTC (a286c2f)">Diff</a>

PR #1002, opened 2025-03-12
Facilitator not explicitly chair

frivoal:
This is a suggestion by @ianbjacobs to work around the lack of a clearly defined group.

----

I think the explicit reference in the opening sentence to decisions (plural) being made based on that consensus is helpful.

This does lack clarity about what happens when there isn't consensus, though we may be able to combine it with https://github.com/w3c/process/pull/997 to deal with that case…

I am less certain about removing the the different types of decisions a that can conclude the phase. This is certainly simpler, but there distinction in the original text were deliberate.

Still, offering this PR for discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1002.html" title="Last updated on Mar 26, 2025, 4:02 PM UTC (240fb3a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1002/cf784bd...frivoal:240fb3a.html" title="Last updated on Mar 26, 2025, 4:02 PM UTC (240fb3a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1785,6 +1785,9 @@ Charter Refinement</h3>
 	who is chosen (and <em class=rfc2119>may</em> be replaced) by the [=Team=]--
 	is responsible for seeking community [=consensus=] among those participating in the refinement process
 	and making decisions reflecting that [=consensus=].
+	In cases where [=consensus=] cannot be found,
+	the [=Chartering Facilitator=] <em class=rfc2119>may</em> ask the [=Team=] to make a [=Team Decision=],
+	and <em class=rfc2119>must</em> document the rationale for the decision.
 
 	Note: The [=Chartering Facilitator=] is not necessarily the [=Chair=] of the [=chartered group|group=] being chartered.
 
--- a/index.bs
+++ b/index.bs
@@ -1783,8 +1783,8 @@ Charter Refinement</h3>
 	with the goal of achieving [=consensus=] on the proposal.
 	The <dfn>Chartering Facilitator</dfn>--
 	who is chosen (and <em class=rfc2119>may</em> be replaced) by the [=Team=]--
-	is responsible for seeking community consensus among those participating in the refinement process
-	and making decisions reflecting that consensus.
+	is responsible for seeking community [=consensus=] among those participating in the refinement process
+	and making decisions reflecting that [=consensus=].
 
 	Note: The [=Chartering Facilitator=] is not necessarily the [=Chair=] of the [=chartered group|group=] being chartered.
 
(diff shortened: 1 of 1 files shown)


ianbjacobs:
Here is my thinking on reusing the term "Chair":

* In previous discussions, the case was made to reuse the Process-defined concept of "formally addressing an issue." That made sense to me in that context because my approach was close enough, and there is value in reuse when fit for purpose.
* In this case, I do not think the concepts of "group" and "Chair" as used throughout the process document today are fit for purpose for this situation. The roles of the Chair in the process document are for chartered groups or for meetings. There's very little in any case about the roles of a Chair in the process document, and the process document directs people to the Guide for a more complete definition. In the Guide, there are many aspects of Chairing that are irrelevant to this situation, hence my comment that the term as used commonly today does not lend itself readily for reuse here.
*  I also think it is unnecessary to add the concepts here in order to make the point that "someone is in charge" and establish roles and responsibilities. 

Here is my thinking regarding conclusion of the charter refinement phase:

* Ultimately the Team makes a decision based on input. Therefore, I found mention of three types of decisions (group, team, chair) to be confusing. It is both simpler and more accurate to say "The Team makes a decision" but it is important to indicate that the decision is based on community input (as we've discussed).
* I suggest deleting "subject to Team verification that the expectations of charter refinement are fulfilled." That is overly prescriptive. Due to its role supporting the W3C mission, it's important that the Team be able to decide not to start refinement even if expectations are fulfilled and that they be able to start refinement even if expectations are not fulfilled (in order to enable the Consortium to make progress). As we've discussed, in these cases the Process Doc will require the Team to provide rationale to the Membership, and define the escalation path for Members if they don't agree with the decision and rationale.

tantek (approved):
Seems reasonable to allow this flexibility.

TallTed (approved):

tantek (approved):
Re-approving with latest activity.

index.bs
```diff
@@ -1783,8 +1783,8 @@ Charter Refinement</h3>
 	with the goal of achieving [=consensus=] on the proposal.
 	The <dfn>Chartering Facilitator</dfn>--
 	who is chosen (and <em class=rfc2119>may</em> be replaced) by the [=Team=]--
-	acts as [=Chair=] for the [=charter refinement=] process,
-	and is responsible for seeking community consensus among those participating in the refinement process.
+	is responsible for seeking community [=consensus=] among those participating in the refinement process
+	and making decisions reflecting that [=consensus=].
```
frivoal, line 1788: If this pull request is accepted, it would remove the text that PR #997 was proposing to amend. Here's a possible amendment to this PR that introduces an analogous change, which clarifies that decisions in the absence of consensus can be made, and how they are made.

```suggestion
	In cases where [=consensus=] cannot be found,
	the [=Chartering Facilitator=] <em class=rfc2119>may</em> ask the [=Team=] to make a [=Team Decision=],
	and <em class=rfc2119>must</em> document the rationale for the decision.

```

PR #1010, opened 2025-03-28
Make cross reference more obvious

frivoal:
The referenced section has a bunch of rules, it would not be good if they were overlooked.

See https://github.com/w3c/process/issues/1008


tantek (approved):

fantasai (approved):
Merging as editorial.

PR #1011, opened 2025-03-28
Refine Expiration of Charter Refinement

frivoal:
* Define extensions explicitly as decisions, which clarifies that they can be appealed if necessary.
* Clarify that failing to extend is de facto abandonment.
* Move paragraph for better cohesion.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1011.html" title="Last updated on Mar 28, 2025, 3:38 AM UTC (cba5c21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1011/5f0a694...frivoal:cba5c21.html" title="Last updated on Mar 28, 2025, 3:38 AM UTC (cba5c21)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1824,6 +1824,13 @@ Charter Refinement</h3>
 	either to initiate [=AC Review=] of the [=charter draft=],
 	or to abandon the proposal.
 
+	The [=Team=] <em class=rfc2119>may</em> [=Team Decision|decide=]
+	to extend the charter refinement period
+	by sending an announcement with the same visibility as the initial [=charter refinement notice=].
+	Reaching the end of the announced period (including any announced extension)
+	without a decision to initiate [=AC Review=] nor without an announcement of extension
+	is considered a de-facto [=Team decision=] to abandon the proposal.
+
 	[=Formal Objections=] filed during the [=charter refinement=] phase
 	are specially handled:
 	* Objections to abandoning the proposal can be appealed only if 5 or more Members,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1824,6 +1824,13 @@ Charter Refinement</h3>
 	either to initiate [=AC Review=] of the [=charter draft=],
 	or to abandon the proposal.
 
+	The [=Team=] <em class=rfc2119>may</em> [=Team Decision|decide=]
+	to extend the charter refinement period
+	by sending an announcement with the same visibility as the initial [=charter refinement notice=].
+	Reaching the end of the announced period (including any announced extension)
+	without a decision to initiate [=AC Review=] nor without an announcement of extension
```
TallTed, line 1831:
```suggestion
	with neither a decision to initiate [=AC Review=] nor an announcement of extension
```

TallTed (changes requested):

ianbjacobs:
Per 4.2, there is an 8-week window for objecting to a decision to abandon the proposal. The window starts with the announcement of the decision to abandon the proposal. I read the text proposed in this pull request as suggesting there could be a decision without an announcement. Even if one might conclude that the start of the 8-week window is the last day of the announced refinement period, it seems simpler to require an announcement in all 3 decision cases.

frivoal:
@ianbjacobs , this may not be apparent from the text, and if so we should tweak further, but (a) Yes, I agree that an announcement of extension should be required at the end of the period if we're neither moving to ac review nor abandoning, however, (b) given that a duration is announced ahead of time, and that missing deadlines is not especially hard to imagine, I think it does make sense to say what happens if the period expires with no decision.

I think the PR as it is does (b) reasonably well, but it might need another pass to do (a) well.

tantek (approved):

PR #1015, opened 2025-03-28
chore: replace informative with non-normative for consistency

### LINKED ISSUE #914: Be consistent with terms "non-normative" and "informative"
In various parts of the document the terms "non-normative" and "informative" are used, and at no point do we state that they in fact mean the same thing, giving rise to a potential misunderstanding that they have different semantics.

The terms are also not defined, so their ordinary English language meanings seem to apply.

Suggest either:
1. using one term throughout - I'd prefer "informative" with a statement that "statements that are not normative are referred to as informative" in the document conventions OR
2. adding into the document conventions a statement "the terms non-normative and informative are used interchangeably"

Bashamega:
fixes #914

***
<a href="https://pr-preview.s3.amazonaws.com/Bashamega/w3c-process/pull/1015.html" title="Last updated on Mar 28, 2025, 6:41 AM UTC (6cf0b96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1015/309bcaa...Bashamega:6cf0b96.html" title="Last updated on Mar 28, 2025, 6:41 AM UTC (6cf0b96)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1195,7 +1195,7 @@ Advisory Board and Technical Architecture Group Elections</h5>
 	are changes of affiliation).
 	(Other formal relationships such as other contracts should be disclosed as potential conflicts of interest.)
 	Each nomination <em class="rfc2119">should</em> include
-	a few non-normative paragraphs about the nominee.
+	a few informative paragraphs about the nominee.
 	If an identified candidate's nomination information is only partially complete
 	as of the deadline for nominations,
 	the [=Team=] <em class=rfc2119>may</em> allow extra time for that candidate's nomination to be completed,
@@ -2751,7 +2751,7 @@ Council Decision Report</h5>
 	The [=Council=] <em class=rfc2119>may</em> also issue a <dfn export>Supplemental Confidential Council Report</dfn>
 	with a more restricted <a href="#confidentiality-levels">level of confidentiality</a> than its main report
 	when it believes that additional commentary on confidential aspects of the case
-	would be non-normative.
+	would be informative.
 	However, the main [=Council Report=] <em class=rfc2119>should</em> be self-sufficient
 	and understandable without reference to Supplemental Confidential Council Reports.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1195,7 +1195,7 @@ Advisory Board and Technical Architecture Group Elections</h5>
 	are changes of affiliation).
 	(Other formal relationships such as other contracts should be disclosed as potential conflicts of interest.)
 	Each nomination <em class="rfc2119">should</em> include
-	a few informative paragraphs about the nominee.
+	a few non-normative paragraphs about the nominee.
```
frivoal, line 1198: I think in this instance, "informative" means "carrying information", not "non-normative", so the replacement is not appropriate.

index.bs
```diff
@@ -2751,7 +2751,7 @@ Council Decision Report</h5>
 	The [=Council=] <em class=rfc2119>may</em> also issue a <dfn export>Supplemental Confidential Council Report</dfn>
 	with a more restricted <a href="#confidentiality-levels">level of confidentiality</a> than its main report
 	when it believes that additional commentary on confidential aspects of the case
-	would be informative.
+	would be non-normative.
```
frivoal, line 2754: In this instance, "informative" definitely means "carrying (interesting) information", not "non-normative", so the replacement is not appropriate.

Bashamega:
Done @frivoal  :)

index.bs
```diff
@@ -3255,7 +3255,7 @@ Classes of Changes</h4>
 			Examples of changes in this class include
 			correcting non-normative examples
 			which clearly conflict with normative requirements,
-			clarifying informative use cases or other non-normative text,
+			clarifying non-normative use cases or other non-normative text,
```
frivoal, line 3258: I'm not sure this part of the correction is needed, though I don't think it is wrong. That said, if we do take it, the phrase becomes a bit redundant, and we could further simplify to something like this:
```suggestion
			clarifying various other non-normative text,
```

frivoal:
Thank you @Bashamega. Help improving the quality and consistency of the text is always very much appreciated!

fantasai (approved):

tantek (approved):

frivoal:
@nigelmegitt The other parts are not coming from the source code, but from bikeshed. I'll follow up to get that addressed there, but this PR is a net improvement.

As for the remaining instances of informative in the text, I think they're actually quite natural and better than alternative rephrasing (like "informational"). Unless you're specifically looking for them as we are here, I think it's unlikely anybody would notice anything unusual about them.

Issue #1004, opened 2025-03-20
Confusing wording wrt adding new features to RECs


fantasai:
> To make changes which introduce a new feature to a Recommendation that does not [allow new features](https://www.w3.org/policies/process/#allow-new-features), W3C must create a new [technical report](https://www.w3.org/policies/process/#technical-report) ...

This bit of text is a bit confusing, since we're not adding a new feature to the Recommendation, but to the technology that the Recommendation defines.

nigelmegitt:
I have always understood this as being about introducing a new feature to a Recommendation.

TallTed:
Perhaps this small change?

> To make changes which introduce a new feature to a Recommendation that does not [allow new features](https://www.w3.org/policies/process/#allow-new-features), W3C must create a new [technical report](https://www.w3.org/policies/process/#technical-report) (i.e., a new Recommendation)...

frivoal:
I'm confused at to what the confusion is.

If I understand correctly, in a way, I agree that Recommendations don't have features, technologies do, and the Recommendation just describes them. So sure, we could, to be technically correct, say something like:

> When making changes which introduce a new feature to a technology described in a Recommendation that does not [allow new features](https://www.w3.org/policies/process/#allow-new-features), W3C must create a new [technical report](https://www.w3.org/policies/process/#technical-report) (i.e., a new Recommendation) to describe the new feature, and must not include normative definitions of the feature in a revision of original Recommendation.

However, that phrasing is clunkier, and as I don't see any other possible interpretation anyway, it does not seem to me that it lifts any ambiguity.

@fantasai, Is there any particular misunderstanding you are worried about?

TallTed:
The reason for my `(i.e., a new Recommendation)` is that brains don't always make the leap to equate an unlinked-to-definition `Recommendation that does not` with a linked-to-definition [`technical report`](https://www.w3.org/policies/process/#technical-report).

@nigelmegitt is correct that the new document may never become a Rec although it is created for the Rec Track. 

I think that @frivoal's new wording goes a bit too far, as I believe a v2.0 that includes such a new feature where v1.x did not would be permitted, so —

[...]

PR #1026, opened 2025-04-15
Fix confusing wording wrt adding new features to RECs #1004

fantasai:
Addresses https://github.com/w3c/process/issues/1004

***
<a href="https://pr-preview.s3.amazonaws.com/fantasai/w3process/pull/1026.html" title="Last updated on Apr 15, 2025, 7:47 PM UTC (79ea26e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1026/6553a80...fantasai:79ea26e.html" title="Last updated on Apr 15, 2025, 7:47 PM UTC (79ea26e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4262,11 +4262,15 @@ Revising a Recommendation: New Features</h5>
 	enables third parties to depend on a stable feature-set for [=Recommendations=] that do not advertise that ability,
 	as was the case for all [=Recommendations=] prior to the 2020 revision of this Process.
 
-	To make changes which introduce a new feature
-	to a [=Recommendation=] that does not [=allow new features=],
-	W3C <em class="rfc2119">must</em> create a new [=technical report=],
-	following the full process of <a href="#rec-advance">advancing a technical report to Recommendation</a>
+	Note: When a [=Recommendation=] does not [=allow new features=],
+	new features can be added by creating a new [=technical report=]
+	and following the full process of advancing that [=technical report=] to [=Recommendation=]--
 	beginning with a new [=First Public Working Draft=].
+	Such [=technical reports=] could be written to represent
+	additional modules building on top of the original [=Recommendation=] of the core technology,
+	or an expanded replacement of the original [=Recommendation=] of the core technology
+	(in which case the new [=technical report=] will typically have the same name as the original,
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4262,11 +4262,15 @@ Revising a Recommendation: New Features</h5>
 	enables third parties to depend on a stable feature-set for [=Recommendations=] that do not advertise that ability,
 	as was the case for all [=Recommendations=] prior to the 2020 revision of this Process.
 
-	To make changes which introduce a new feature
-	to a [=Recommendation=] that does not [=allow new features=],
-	W3C <em class="rfc2119">must</em> create a new [=technical report=],
-	following the full process of <a href="#rec-advance">advancing a technical report to Recommendation</a>
+	Note: When a [=Recommendation=] does not [=allow new features=],
```
tantek, line 4265:
```suggestion
	When a [=Recommendation=] does not [=allow new features=],
```
Drop "Note: " because in typical W3C contexts it implies a non-normative sentence/paragraph/section, and I don't think that was the intent of this change.

tantek (changes requested):
Minor fix to avoid implying non-normative text.

The new "Such …" sentence I think belongs in a guide (the Guide?), not The Process, and IMO unnecessarily adds length to the Process (which I believe we are trying to continuously shrink and simplify over time).

index.bs
fantasai: No, that was actually intended. :) This paragraph is identifying an implication of existing normative text.

tantek (approved):
Thanks for the explicit intention of the "Note:".

I still think a Guide patch with examples should be sufficient to help reduce confusion, at least something to try before expanding the Process length. However, I still leave this up to editor discretion.

frivoal (approved):

cwilso (approved):

nigelmegitt:
> @tantek I understand what you're getting at, but also this has been a frequent point of confusion, so I think the examples help. But I could be persuaded otherwise depending what others think. CC @nigelmegitt

Sorry only just got to this, but I think the change is fine and should help with a point of confusion I have witnessed.

Issue #1033, opened 2025-04-22
Special rules for excessively short terms


fantasai:
Depending on the precise schedule of special elections, some people might end up with very short terms, if they are elected to fill the remainder of the term of someone who didn't have much time left. Should we automatically extend (by two years) terms that would be less than X months? If so, for which value of X?

Note: This is not about how long ahead of a regular election we should have a special election. That's already defined as 3 months (roughly the time it takes to organize an election).

fantasai:
My suggestion here would be the same as @martinthomson in https://github.com/w3c/process/pull/888/files#r2053267191: Yes, and 3 months.

frivoal:
Made https://github.com/w3c/process/pull/1044 to address this.

frivoal:
pinging @fantasai, as the person who opened this issue. If you'd like to push back against closing this, now's a great time.

Issue #688, opened 2022-12-13
Should W3C continue to stagger AB/TAG elections ?


michaelchampion:
I see this point raised  by @chaals in https://github.com/w3c/w3process/issues/484#issuecomment-749373454 :

> staggering terms, reduces the ability of any voting system to support diversity. This is the case under any voting system, even approval (which generally works against diversity). The purported benefit of staggered terms is to ensure that continuity provides some ongoing measure of institutional memory, rather than finding that everyone is new. That is a separate question, and I will raise a separate issue to discuss it as a proposal.

Apologies if Chaals has opened this issue and the CG or AB closed it ...But the current December 2022 situation where there are 4 vacant AB seats, 3 whose terms end in six months, makes this question very timely.

Strawman proposal:  Revise https://www.w3.org/2021/Process-20211102/#ABParticipation to eliminate  "Terms are staggered so that each year, either five or six terms expire.” And add some language about how to handle people whose terms would have expired in 2023 and 2024. (I don't have a proposal for that language).

Discussion:  I agree with Chaals that staggering terms reduces the ability of STV to support outcomes that allow representation of diverse geographical, business, philosophical, etc. groupings in proportion to the size of the electorate aligned on one or more of those dimensions.  The original motivation for staggering terms, which I also recall as "preserve institutional memory" seems rather hypothetical given the tendency for incumbents to be re-elected, even under STV.   It will take some work to find language to handle those elected for 2-year terms in 2022 if this change is adopted before the next AB election, or how to phase this in over multiple election cycles.

cwilso:
I'm in strong agreement that we should stop staggering terms.

igarashi50:
I agree on stopping staggering terms. I do not see the benefit of the "preserve institutional memory" for AB/TAG since of the tendency incumbents to be re-elected. Also, it is better for W3C to choose talented and diverse persons by electing all seats of AB/TAG with the STV voting.

chaals:
I agree (with my older opinion) that staggered terms means the elections produce less representative results each time.

I am concerned about institutional memory, despite the tendency for incumbents to win re-election. I have repeatedly watched groups (outside the W3C as well) go over the same ground after a few years, and those who are able to look carefully at what happened last time seem to cover it far more effectively. That includes how they understand whether there is "something different this time". However, I think there are better mechanisms available than ensuring some people get to hang around when the membership have hanged their mind about what they are looking for.

[...]

PR #1044, opened 2025-05-13
Extent excessively short terms

frivoal:
See #1033

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1044.html" title="Last updated on May 13, 2025, 2:14 AM UTC (9e7db76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1044/2691bf7...frivoal:9e7db76.html" title="Last updated on May 13, 2025, 2:14 AM UTC (9e7db76)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -927,8 +927,6 @@ Composition of the Advisory Board</h5>
 	The terms of elected Advisory Board participants are for <span class="time-interval">two years</span>.
 	Terms are staggered so that each year,
 	either five or six terms expire.
-	If an individual is elected to fill an incomplete term,
-	that individual's term ends at the normal expiration date of that term.
 	Regular Advisory Board terms begin on 1 July and end on 30 June.
 
 <h5 id=ab-comm>
@@ -1039,8 +1037,6 @@ Composition of the Technical Architecture Group</h5>
 	The terms of TAG participants last for <span class="time-interval">two years</span>.
 	Terms are staggered so that four elected terms
 	and either one or two appointed terms expire each year.
-	If an individual is appointed or elected to fill an incomplete term,
-	that individual's term ends at the normal expiration date of that term.
 	Regular TAG terms begin on 1 February and end on 31 January.
 
 <h5 id=tag-comm>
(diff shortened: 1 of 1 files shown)


chaals:
I'm not in favour of this.

If we need an extra person, then have the election. But the more seats are up for election at a time, the more representative the outcomes are of W3C as a whole (see also the discussion in #688), whereas the most common case of a single-seat special election means that to the extent there are "factions" the single biggest voting bloc gets an extra representative. I understand the disincentives to short-term participation (having been there and done it), but I don't think it makes sense for that relatively unbalanced representation to be extended beyond a scheduled election.

fantasai:
Maybe this needs to just be for the case where we have a full recall. Or maybe just for the case where half of the group's seats are vacated.

fantasai:
Btw, there's an open issue for discussion, perhaps we should move the conceptual conversation there?
https://github.com/w3c/process/issues/1033

frivoal:
@chaals : we already have a rule saying that we should not have special to fill vacant seats if the next regularly scheduled election is fewer than three months away. This PR doesn't change that, but deal with what happens if we do anyway.

Also, fantasai is possibly right that this matters differently in the case of one or a few vacant seats (→ should just wait), vs cases of many/most seats being empty.

cwilso:
I'm in strong agreement with Chaals.  We should not extend the term because it interferes with the election dynamics.  

Overall, I don't think it's necessary to fix this "problem"; it's left to the Chair(s) whether to call a special election or wait until the next regularly scheduled election, and I think that's good enough.  (This has been exercised both ways - calling a special election in late 2022 when four seats on the AB were vacated, and not calling a special election in early 2025, when a seat was vacated.)

The only reason I can see for this being necessary would be a total recall of the elected body, which is similarly not really something I think we need to solve.

frivoal:
Keeping this open, but removing my own agenda+: there is currently no consensus, and I am not trying to get to consensus in the short term (i.e. within the current iteration of the Process).

frivoal:
Given the lack of consensus, I'd like to propose to close.

frivoal:
Process CG agreed to close during the 2025-10-08 meeting.

Issue #1041, opened 2025-05-09
Multi-point Multi-FO Councils


fantasai:
The [Council Deliberations section](https://www.w3.org/policies/process/drafts/#council-deliberations) is written as if there is only one objection, which creates some oddities when there are a) multiple FOs and/or b) FOs with multiple points.

Two possible approaches to cleaning this up are:
-  Rewrite to say that the Council upholds or overturns *a decision*.
-  Rewrite to say that the Council upholds or overturns individual points in the objections.

(A third approach of upholding or overturning individual FOs would mean treating 6 points filed as a single FO and 6 FOs on different points differently, which seems bad.)

Functionally the Council is either upholding or overturning a decision. However its rationale (and therefore the expected follow-up actions) depend on which objections the Council felt were worth upholding or overruling.

What makes sense to do here?

nigelmegitt:
I think it should be based on a _decision_.

dwsinger:
Interestingly I think it's a minor re-write and a plausible improvement:

Otherwise, after sufficient deliberation, the [W3C Council](https://www.w3.org/policies/process/drafts/#w3c-council) decides whether to uphold or overturn the original decision. The [W3C Council](https://www.w3.org/policies/process/drafts/#w3c-council) may [uphold](https://www.w3.org/policies/process/drafts/#uphold) the original decision even if it agrees with some of the supportive arguments in the [Formal Objection(s)](https://www.w3.org/policies/process/drafts/#FormalObjection).

When [overturning](https://www.w3.org/policies/process/drafts/#overturn) a decision, it should recommend a way forward. If the overturned decision…

frivoal:
Decisions can be unambiguously identified, and so can Formal Objections. Points in an objection, on the other hand, are a more subjective division, and having a requirement to individually address something we might not be able to all agree about how many there are seems to invite trouble to me. We could unambiguously go with rulling on each each objection, but that sees to miss the point of judging them together and of having a per-decision council. It may very well be that several objections make good points, which individually, while valid, would not be strong enough to justify blocking/overturning a decision, but do in aggregate. In such cases, it seems odd to pick any particular objection to say that that's the one sustaining (causing the decision to be overturned), and much more sensible to talk directly about the decision being overturned (due to reasons in the objections).

I agree that the current text is a little incoherent in the way it's phrased, so a tweak of the language is likely in order. I think something along the lines of what David said seems about right.

frivoal:
[...]

PR #1045, opened 2025-05-13
Clarify council text in case of multiple objections to a single decision

frivoal:
Even though the logic already was that the Council processes all objections to a single decision together, parts of the original text were written as if there was only a single objection, and were awkward or in case of several.

This rephrasing makes the text coherent.

See #1041

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1045.html" title="Last updated on Jun 11, 2025, 3:10 PM UTC (ac10398)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1045/2691bf7...frivoal:ac10398.html" title="Last updated on Jun 11, 2025, 3:10 PM UTC (ac10398)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2665,7 +2665,7 @@ Council Deliberations</h5>
 	even if it agrees with some of the arguments of a [=Formal Objection=].
 
 	<p id=fo-mitigations>
-	When [=overturning=] an decision,
+	When [=overturning=] a decision,
 	it <em class=rfc2199>should</em> recommend a way forward.
 	If the overturned decision has already had consequences
 	(e.g., if the objection concerns material already in a published document)
--- a/index.bs
+++ b/index.bs
@@ -2662,7 +2662,7 @@ Council Deliberations</h5>
 	the [=W3C Council=] decides whether to
 	<dfn>confirm</dfn> or <dfn>overturn</dfn> the decision being objected to.
 	The [=W3C Council=] <em class=rfc2119>may</em> [=confirm=] the decision
-	even if it agrees with some of the arguments of a [=Formal Objection=].
+	even if it agrees with some of the arguments made as part of a [=Formal Objection=].
 
 	<p id=fo-mitigations>
 	When [=overturning=] a decision,
--- a/index.bs
+++ b/index.bs
@@ -2661,8 +2661,8 @@ Council Deliberations</h5>
 	after sufficient deliberation,
 	and with due consideration of each argument in the applicable [=Formal Objections=],
 	the [=W3C Council=] decides whether to
-	<dfn>confirm</dfn> or <dfn>overturn</dfn> the decision being objected to.
-	The [=W3C Council=] <em class=rfc2119>may</em> [=confirm=] the decision
+	<dfn>uphold</dfn> or <dfn>overturn</dfn> the decision being objected to.
+	The [=W3C Council=] <em class=rfc2119>may</em> [=uphold=] the decision
 	even if it agrees with some of the arguments made as part of a [=Formal Objection=].
 
 	<p id=fo-mitigations>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2655,17 +2655,17 @@ Council Deliberations</h5>
 	<p id=broker-consensus>
 	The Council <em class=rfc2119>may</em>
 	further attempt to broker consensus,
-	which, if successful, disposes the formal objection.
+	which, if successful, disposes that formal objection.
```
nigelmegitt, line 2658: Since more than one FO can be handled by a Council in relation to a single decision:

```suggestion
	further attempt to broker consensus on one or more of the formal objections
	which, if successful, would allow them to be resolved.
```

index.bs
```diff
@@ -2655,17 +2655,17 @@ Council Deliberations</h5>
...
-	<dfn lt="uphold | upholds | upheld | upholding">uphold</dfn> or <dfn>overrule</dfn> the objection.
-	The [=W3C Council=] <em class=rfc2119>may</em> [=overrule=] the [=Formal Objection=]
-	even if it agrees with some of the supportive arguments.
+	<dfn>confirm</dfn> or <dfn>overturn</dfn> the decision being objected to.
+	The [=W3C Council=] <em class=rfc2119>may</em> [=confirm=] the decision
+	even if it agrees with some of the arguments of a [=Formal Objection=].
 
 	<p id=fo-mitigations>
-	When [=upholding=] an objection,
+	When [=overturning=] an decision,
```
nigelmegitt, line 2668: typo

```suggestion
	When [=overturning=] a decision,
```

nigelmegitt (commented):
Small suggestions.

nigelmegitt (approved):

nigelmegitt:
> These changes focus on each decision, not on each FO (nor on each facet of each FO), which is the reverse of the perspective generally desired in today's concall.

I wasn't on the call, but I disagree with that perspective - the focus should be on Decisions not FOs in my opinion.

index.bs
```diff
@@ -2654,18 +2654,18 @@ Council Deliberations</h5>
...
 
 	Otherwise,
 	after sufficient deliberation,
 	the [=W3C Council=] decides whether to
-	<dfn lt="uphold | upholds | upheld | upholding">uphold</dfn> or <dfn>overrule</dfn> the objection.
-	The [=W3C Council=] <em class=rfc2119>may</em> [=overrule=] the [=Formal Objection=]
-	even if it agrees with some of the supportive arguments.
+	<dfn>confirm</dfn> or <dfn>overturn</dfn> the decision being objected to.
+	The [=W3C Council=] <em class=rfc2119>may</em> [=confirm=] the decision
+	even if it agrees with some of the arguments of a [=Formal Objection=].
```
chaals, line 2665:
```suggestion
	even if it agrees with some of the arguments made as part of a [=Formal Objection=].
```

phrasing. YMMV...

chaals:
> > These changes focus on each decision, not on each FO (nor on each facet of each FO), which is the reverse of the perspective generally desired in today's concall.
> 
> I wasn't on the call, but I disagree with that perspective - the focus should be on Decisions not FOs in my opinion.

+1 to @nigelmegitt

chrisn:
This looks like a good rephrasing that clarifies while preserving the intent of the original text. I agree with @TallTed though (assuming I'm understanding you correctly) that Councils need to consider all the arguments made within each FO to properly come to a conclusion on whether to uphold or overturn a decision.

index.bs
```diff
@@ -2654,18 +2654,18 @@ Council Deliberations</h5>
...
 	<p id=broker-consensus>
 	The Council <em class=rfc2119>may</em>
-	further attempt to broker consensus,
-	which, if successful, disposes the formal objection.
+	further attempt to broker consensus on one or more of the formal objections,
+	which, if successful, would allow them to be resolved.
 
 	Otherwise,
 	after sufficient deliberation,
 	the [=W3C Council=] decides whether to
```
fantasai, line 2662: I agree with @nigelmegitt's explanation / recollection in https://github.com/w3c/process/issues/1041#issuecomment-2880592714 that this PR is the right way to resolve this issue, but @cwilso raised the issue on the call (and @chrisn in https://github.com/w3c/process/pull/1045#issuecomment-2886010001) that the text doesn't sufficiently emphasize that each argument of the FO needs consideration. (This point is only mentioned in the section on what goes in the Council Report.) So my suggestion is two changes.

1. Clarify that each argument of each FO needs to be considered by the Council, thus:

```suggestion
	after sufficient deliberation,
	and with due consideration of each argument in the applicable [=Formal Objections=],
	the [=W3C Council=] decides whether to
```

2. Replace the word “confirm” with “uphold”, which feels like it carries slightly less incumbent bias.

nigelmegitt (commented):
Post merge, but the agreed changes look good to me. Slightly surprised there isn't a defined term for `decision`, but that's a minor nit.

Issue #1018, opened 2025-04-02
What happens in Council votes when two members share the same affiliation?


TallTed:
_Originally posted by @TallTed in https://github.com/w3c/process/pull/1017#discussion_r2021638069_

https://github.com/w3c/process/blob/d9955a44a606e2f261e351e2b2fbca5d1deb68b3/index.bs#L2708-L2711

The lines shown above appear to suggest that if two members of a Council who share the same affiliation cast *different* ballots, then they are both counted, while if they cast *identical* ballots, then neither is counted. I don't believe that to be the intent. I believe that all members of a Council who share the same affiliation are meant to work among themselves to condense their votes to a single ballot, which will be counted.

frivoal:
> I don't believe that to be the intent.

It is:
* most of the time, the Council does not decide by voting, it decides by consensus, so this provision does not apply.
* If we do vote, and representatives from the same member cannot agree with each other, their votes cancel out; if they do agree, they count, but as a single one, not with the added weight being several individuals.

> I believe that all members of a Council who share the same affiliation are meant to work among themselves to condense their votes to a single ballot, which will be counted.

That's what they are encouraged to do. But if they cannot work it out, they cancel out.

> What if there are more than two members of a Council who share the same affiliation?

It is going to be rare, due to https://www.w3.org/policies/process/#AB-TAG-constraints. But indeed, it is not impossible, and the text isn't clear.

> What if there are more than two options to choose between in the vote?

I don't believe that's possible. Council decisions are binary: uphold or overturn.

fantasai:
I think we can just write
> if _multiple_ members of a Council who share the same affiliation cast _identical ballots_

brentzundel:
This will be addressed by #1047

PR #1047, opened 2025-05-13
Handle votes by >2 council members from the same affiliation

frivoal:
In rare cases, there can be more than 2 members of the council with the same affiliation. Generalize the phrasing about votes to handle that too.

See #1018

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1047.html" title="Last updated on May 13, 2025, 4:25 AM UTC (5369982)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1047/2691bf7...frivoal:5369982.html" title="Last updated on May 13, 2025, 4:25 AM UTC (5369982)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2696,9 +2696,8 @@ Council Deliberations</h5>
 	then members of that group at the time the decision or proposal was made
 	must abstain in such a vote.
 	In case of a vote,
-	if two members of a Council who share the same affiliation cast an identical ballot,
-	then their ballots count as a one vote,
-	not two.
+	if multiple members of a Council who share the same affiliation cast identical ballots,
+	then these identical ballots count as a single vote.
 
 	In the case of non-unanimous decisions,
 	members of a [=W3C Council=] who disagree with the decision


plehegar (approved):

index.bs
```diff
@@ -2696,9 +2696,8 @@ Council Deliberations</h5>
 	then members of that group at the time the decision or proposal was made
 	must abstain in such a vote.
 	In case of a vote,
-	if two members of a Council who share the same affiliation cast an identical ballot,
-	then their ballots count as a one vote,
-	not two.
+	if multiple members of a Council who share the same affiliation cast identical ballots,
+	then these identical ballots count as a single vote.
```
TallTed, line 2700: As I understand it, a maximum of 3 members of a council might share affiliation.

|  1  |     |     | net |
|:---:|:---:|:---:|:---:|
|  +  |     |     |  +  |
|  –  |     |     |  –  |

|  1  |  2  |     | net |
|:---:|:---:|:---:|:---:|
|  +  |  +  |     |  +  |
|  +  |  –  |     |  0  |
|  –  |  +  |     |  0  |
|  –  |  –  |     |  –  |

|  1  |  2  |  3  | net |
|:---:|:---:|:---:|:---:|
|  +  |  +  |  +  |  +  |
|  +  |  +  |  –  |  0  |
|  +  |  –  |  +  |  0  |
|  +  |  –  |  –  |  0  |
|  –  |  +  |  +  |  0  |
|  –  |  +  |  –  |  0  |
|  –  |  –  |  +  |  0  |
|  –  |  –  |  –  |  –  |

Again as I understand it, if multiple members of a council share affiliation, they *must* vote in agreement for their votes to collapse to one (`1`). Any time any member votes against another sharing their affiliation — regardless of how many sharing that affiliation vote together — all votes from that affiliation collapse to zero (`0`).

Is that what's intended?

nigelmegitt: That result makes sense to me @TallTed - though I'm not sure by what mechanism there could be 3 members with the same affiliation, except in unusual transition cases before people with the same affiliation have stepped down or gone up for re-election, etc.

fantasai: @nigelmegitt That's exactly the situation that could cause this to arise, yes. Particularly given that the Council works primarily by consensus and not by vote, chances of us running into this situation, as cwilso said on the call, is pretty close to zero.

fantasai: @TallTed I think limiting the total voting power to 1 is probably preferable here to zeroing out any conflicts. The question is, can we do it without complicating the wording excessively. :)

frivoal: Right. The situation will be exceedingly rare (if it ever happens at all), because:
* most councils never vote at all, as they should (and typically do) decide by consensus
* 3 members with a same affiliation only happens in exceedingly rare transition periods

And yes, I agree that the goal is to limit the influence of one company to no more than 1 vote in either direction. @TallTed's summary table is indeed what I think the current text does (or tries to do), and it fulfills that goal. I'd be open to different systems that fulfill that goal too, if they can be described simply. This is just handling a rare corner case, so in practice it essentially doesn't matter, but it would be good to have something (a) simple and well defined in case the situation ever arises.

chrisn: That was my interpretation too.

TallTed: I believe we have a number of other situations where there's "one vote per member organization", and there's no problem with expressing those as such. Why not just say the same here?

chrisn: I agree, I think that would be clearer overall, and the Process then doesn't need to specify how the people involved come to a decision on how to vote, only that they have to.

frivoal: The difference is that in working groups and the like, people actually represent their organization, and thus their hierarchy can function and make them decide how to vote somehow. We don't need to care how.

However, in the context of a council, while people do have an affiliation, they do not act as representative of their organizations: AB/TAG members are elected “with the expectation that they will use their best judgment to find the best solutions for the Web, not just for any particular network, technology, vendor, or user”, and “their activities [as member representatives] are separate and distinct from their activities on the Advisory Board or TAG”: they “do not advocate for the commercial interests of their employers”.

So, If they cannot agree, they cannot go to their boss or to their AC rep, tell them what the situation is, and ask them to decide what the company position is: the company position is not welcome, and the council's confidentiality forbids this discussion anyway.

That doesn't mean we can entirely act as if they did not have the same affiliation: they do, and even when they chose to act in the best interest of the web rather than of their employer, they have the perspective and biases of their employer. This makes it desirable to limit them to (at most) one vote.

If they agree, great. But since we cannot invoke the employer or anything external to the council itself to make them agree if they don't, we need a rule that says what happens when that is the case, and which limits them to (at most) one vote. "The first one by alphabetical order wins" would work, but it would feel very arbitrary. The one proposed here (as a variation of the earlier rule which forget that there might be cases when they are more than two, even if that's vanishingly rare) seems fine by me, though we could come with alternatives (for instance, sum their votes as +1s / 0s / -1s, and clamp the result to [-1, +1]).

Issue #1066, opened 2025-06-06
Use “Staff Contact” instead of “Team Contact” consistently in communications with external audiences


tamsinewing555:
There are 2 issues:
1. Inconsistent use of the terms 'Staff Contact' and 'Team Contact' in W3C communications.
2. Different meanings in the terms

- **Staff** (member): A person employed by a particular organization
- **Team** (member): A person who works collaboratively with others towards a shared goal — in the context of W3C project work, it could mean either someone who is employed by W3C or someone employed by another organization.

I am unconcerned about how the terms are used in _internal_ communications as I understand there are many contexts that make sense at W3C, where people have fully internalized the distinctions and can relate to them.

However, using both terms inconsistently with _external_ target audiences can be confusing.

I therefore propose that we only use the term 'Staff Contact' in external communications.

Rationale:
When we have a contact for a group made up of both internal and external people  (i.e., working groups and task forces), using the term 'Staff Contact' rather than 'Team Contact' makes it clear to people unfamiliar with how W3C works that this is someone employed by W3C.

I understand the group contact role is only ever held by someone employed by W3C.

Examples of 2 scenarios:

1. **Linking between web pages** (such as linking from the WAI subsite pages to information on w3.org/groups pages as the source of truth) —  using the same term for the role of the contact person on each site is best practice from a user experience perspective.
2. **WAI email announcement** (these often have multiple signatories, some of whom are employed by W3C and some who are external to W3C) — using the term 'Staff Contact' makes it clear that this role is held by someone employed by W3C.

frivoal:
Consistency of terminology is good, and would help reduce confusion. At the same time, there is *a lot* of history in calling the people employed by W3C and its partners (as well as quasi-employess contractors) "The Team". Shifting away from that is going to be non-trivial.

Even if you simply start at https://www.w3.org/staff/ (and yes, I know the URL say "staff"), the first sentence is "The W3C Team is composed of […]". The word "Team" is even used in some legal agreements (like the Partnership agreements).

I agree that using Team in this sense is jargonny and non obvious until explained, and that we may gain from shifting to Staff. So I don't oppose the proposed change.

But I want to highlight that this is not a simple correction of one term in the W3C Process. Usage is pervasive and deeply embedded. If we want to replace usage of Team by Staff (and not simply perpetuate the fact that both are used interchangeably), I think this is going to take some deliberate and coordinated effort.

koalie:
[...]

PR #1068, opened 2025-06-10
Rename "Team Contact" into "Staff Contact"

frivoal:
See issue #1066

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1068.html" title="Last updated on Jun 10, 2025, 6:22 AM UTC (c5adca4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1068/4cbc515...frivoal:c5adca4.html" title="Last updated on Jun 10, 2025, 6:22 AM UTC (c5adca4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -916,9 +916,9 @@ Composition of the Advisory Board</h5>
 	and <em class=rfc2119>may</em> be run at other times when initiated by the current chairs or the Team,
 	for example if a chair steps down or if a minority of the participants make such a request.
 
-	The team also appoints a <a href="https://www.w3.org/guide/teamcontact/role">Team Contact</a>,
+	The team also appoints a [=Staff Contact=],
 	as described in <a href="#ReqsAllGroups"></a>.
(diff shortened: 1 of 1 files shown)


frivoal:
Note: this PR assumes that instances of "Team Contact" in the guide and such places have not yet been replaced, and in particular, this page: https://www.w3.org/guide/teamcontact/role, which has that term not just in its content, but also in its URL.

I'm thinking we can do it in two steps: First we establish (as this PR does) the fact that Staff Contact is a legitimate term, then we update all sorts of places, including this guide article, to say Staff Contact instead of Team contact (I realize this has started, but it's not yet complete), and during a later process cycle, once the Guide and other prominent places no longer use the word Team contact, we do as second pass, only keep Staff Contact in the main body of text, and move Team Contact to https://www.w3.org/policies/process/drafts/#old-terms

Issue #1063, opened 2025-05-28
Republishing a CR Snapshot in order to start the AC Review


plehegar:
[[
If all the criteria above are fulfilled, the [Team](https://www.w3.org/policies/process/drafts/#team) must begin an [Advisory Committee Review](https://www.w3.org/policies/process/drafts/#advisory-committee-review) on the question of whether the identified [Candidate Recommendation Snapshot](https://www.w3.org/policies/process/drafts/#candidate-recommendation-snapshot) is appropriate to [publish](https://www.w3.org/policies/process/drafts/#publishing) as a [W3C Recommendation](https://www.w3.org/policies/process/drafts/#RecsW3C). If the document’s most recent publication is not a [Candidate Recommendation Snapshot](https://www.w3.org/policies/process/drafts/#candidate-recommendation-snapshot), then it must be republished as such in order to provide a basis for review.
]]
https://www.w3.org/policies/process/drafts/#initiating-rec-review

In other words, if you fixed a typo and republish a CR Draft, you are now forced to republish a CR Snapshot and trigger a 60 exclusion period because of the Call for Exclusion generated by the CR Snapshot.

In addition, the following sentence was removed:
[[
A [Proposed Recommendation](https://www.w3.org/policies/process/#RecsPR) may identify itself as intending to allow new features ([class 4 changes](https://www.w3.org/policies/process/#correction-classes)) after its initial publication as a [Recommendation](https://www.w3.org/policies/process/#RecsW3C), as described in [§ 6.3.11.4 Revising a Recommendation: New Features](https://www.w3.org/policies/process/#revised-rec-features).
]]

which means that a group would be forced to republish a CR Snapshot if they wanted to add that sentence before starting the AC Review.

That seems an overburden for the working groups.

We should allow doing an AC Review on a CR Draft, as long as no substantive changes were made since the last CR Snapshot.

cc @deniak 

(btw, I have a vague sentiment that I asked this question before but I couldn't find the answer I got, so apologizes for asking again)

plehegar:
cc https://github.com/w3c/modern-tooling/issues/112

frivoal:
In the opening comment here, and during the call yesterday, the question was raised about what happened to the sentence that talked about new features only be allowed in RECs that explicitly say so, and the constraints on when that statement can be added.

It's at [https://www.w3.org/policies/process/drafts/#RecsW3C](https://www.w3.org/policies/process/drafts/#RecsW3C):

[...]

PR #1070, opened 2025-06-12
Allow transition to REC from CRD in certain cases

frivoal:
See #1063

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4153,15 +4153,25 @@ Requirements for Transition</h5>
 	and <em class=rfc2119>must not</em> include any such marking
 	if not already present.
 
+	If the document's most recent publication is a [=Candidate Recommendation Draft=],
+	the [=Team=] <em class=rfc2119>must</em> verify
+	that it contains no changes since the previous [=Candidate Recommendation Snapshot=] other than:
+	* [=editorial changes=],
+		including the addition or removal of a statement as to whether the document will [=allow new features=]
+		after its publication as a [=Recommendation=].
+	* dropping [=at risk=] features
+
+	Otherwise, the [=Working Group=] <em class=rfc2119>must</em> republish it
+	as a [=Candidate Recommendation Snapshot=]
+	prior to initiating the review.
+
 <h5 id="initiating-rec-review">
 Initiating Review</h5>
 
 	If all the criteria above are fulfilled,
 	the [=Team=] <em class="rfc2119">must</em> begin an [=Advisory Committee Review=]
-	on the question of whether the identified [=Candidate Recommendation Snapshot=]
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -4153,15 +4153,25 @@ Requirements for Transition</h5>
 	and <em class=rfc2119>must not</em> include any such marking
 	if not already present.
 
+	If the document's most recent publication is a [=Candidate Recommendation Draft=],
+	the [=Team=] <em class=rfc2119>must</em> verify
+	that it contains no changes since the previous [=Candidate Recommendation Snapshot=] other than:
+	* [=editorial changes=],
+		including the addition or removal of a statement as to whether the document will [=allow new features=]
```
nigelmegitt, line 4160: tl;dr: hard to express discomfort, no proposal to change the PR wording here.

I have a feeling that adding or removing such a statement is more substantive a change than it might at first appear, but I'm having trouble expressing why. I can't think of any patent implications though, so I guess this is okay.

My concern is to do with this having a significant potential long term impact on the way the report is perceived, and how it can be amended over time, that warrants some care, but I guess it's highly unlikely that AC would ever want to overturn a WG decision about it. As long as the WG's working mode doesn't allow changes like this to be introduced in CRDs without a WG Decision, it's okay...

plehegar (approved):

index.bs
fantasai: Wrt the edits to the Process, I think we should just pull it out into its own bullet in this list, separate from [=editorial changes=].

But also, I don't think we're clear anywhere that this is not something that the WG can change out from under the AC during its review?

fantasai: Maybe we can clarify that changing from allowing to disallowing new features is a [=substantive change=].

frivoal: I've added https://github.com/w3c/process/pull/1070/commits/f9c1c89aab1362434ea4cda0a14f16e65233c413 to address this concern.

nigelmegitt: That's definitely an improvement, thanks @frivoal if for no other reason than it avoids a potential misunderstanding of the combination of negative + includes. Not sure it addresses the point about changing from under the AC during its review though?

frivoal: > Adding or removing this allowance is a [=substantive change=].

does that, since you cannot silently make substantive changes post AC review without another round of actively checking for consensus.

frivoal: The review is about a single dated document. If changes and new publications are made during or after the review, they aren't on the same dated revision of the document, and they are irrelevant to that review, unless the provisions of https://www.w3.org/policies/process/drafts/#ACReviewAfter are used to apply additional changes to the reviewed material.

brentzundel (approved):

plehegar (approved):

PR #1176, opened 2026-07-03
Drop the requirement that AC meetings be face-to-face. Fixes #1084.

### LINKED ISSUE #1084: Reconsider requirement for AC meetings to be face-to-face
(Split off from w3c/AB-public#283)

[§ 3.2.4. Advisory Committee Meetings](https://www.w3.org/policies/process/#ACMeetings) contains the following sentence:

> The <a href=https://www.w3.org/policies/process/#team>Team</a> organizes a <a href=https://www.w3.org/policies/process/#ftf-meeting>face-to-face meeting</a> for the <a href=https://www.w3.org/policies/process/#advisory-committee>Advisory Committee</a> twice a year.

This obliges the Team to organize two AC meetings a year, and it requires those AC meetings to be face-to-face.

In practice, the in-person requirement has been put aside several times over the years. Should this requirement be revisited?

hober:
Per https://github.com/w3c/process/issues/1084#issuecomment-4865575497

***
<a href="https://pr-preview.s3.amazonaws.com/hober/process/pull/1176.html" title="Last updated on Jul 3, 2026, 7:24 AM UTC (eac2d8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1176/658acd5...hober:eac2d8c.html" title="Last updated on Jul 3, 2026, 7:24 AM UTC (eac2d8c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -821,7 +821,7 @@ Advisory Committee Mailing Lists</h4>
 <h4 id="ACMeetings">
 Advisory Committee Meetings</h4>
 
-	The [=Team=] organizes a <a href="#ftf-meeting">face-to-face meeting</a> for the [=Advisory Committee=]
+	The [=Team=] organizes a <a href="#GeneralMeetings">meeting</a> for the [=Advisory Committee=]
 	<span class="time-interval">twice a year</span>.
 	The [=Team=] appoints the Chair of these meetings (generally the CEO).
 	At each Advisory Committee meeting,


index.bs
```diff
@@ -821,7 +821,7 @@ Advisory Committee Mailing Lists</h4>
 <h4 id="ACMeetings">
 Advisory Committee Meetings</h4>
 
-	The [=Team=] organizes a <a href="#ftf-meeting">face-to-face meeting</a> for the [=Advisory Committee=]
+	The [=Team=] organizes a <a href="#GeneralMeetings">meeting</a> for the [=Advisory Committee=]
 	<span class="time-interval">twice a year</span>.
```
frivoal, line 825: Since the intent is to provide flexibility, rather than call for meetings to be primarily online, I'd suggest this rather than dropping the mention of F2F altogether.
```suggestion
	<span class="time-interval">twice a year</span>.
	These meetings should be [=face-to-face meetings|face-to-face=].
```

index.bs
frivoal, line 553:
```suggestion
			Nevertheless, facilities <em class=rfc2119>must</em> be provided to enable remote attendance.

```
Given that this is not specified anywhere, and that discussing face-to-face meetings regularly surfaces the question of whether they are different from hybrid meetings, I thought we might as well make it explicit that face-to-face meetings always provide for remote participation.

brentzundel:
I'm not opposed to Florian's suggestions, but I prefer the simplicity of the original change. I think it's fine for there to be some room for trust and interpretation of the Process.

This is a case where if the Team begins acting consistently in a way that is not in line with the best interests of the membership (e.g., by making the AC meeting a one-hour teleconference rather than a several-day in-person event), there are other avenues that exists for fixing that. And we can also strengthen the requirements in the Process at that point if other avenues prove insufficient.

tidoust (approved):
I also prefer the original change without amendments. That said, I'm fine adding the requirement about providing remote facilities to face-to-face meetings (it used to be harder to guarantee, e.g., for group meetings, but that should no longer be a problem).

nigelmegitt:
I agree with @brentzundel and @tidoust but would go a little further and call out a potential issue with @frivoal 's suggestion about "facilities must be provided to enable remote attendance", which is that it is a poorly scoped requirement in that it's unclear where W3C's responsibilities end. What counts as "facilities"?

For example, if an AC member is on a remote island with no access to the internet, does this obligate W3C somehow to provide internet access so they can attend remotely?

frivoal:
> that it is a poorly scoped requirement in that it's unclear where W3C's responsibilities end. What counts as "facilities"?
>
> For example, if an AC member is on a remote island with no access to the internet, does this obligate W3C somehow to provide internet access so they can attend remotely?

That seems like a stretch. Sure, we could maybe be a little more specific, but the phrasing isn't "make sure anyone on earth (or in space) can attend, regardless of their location, physical condition, or lack of connectivity". Just stating that there needs to be a way to attend without being present on site. Basically, it's what we do already, but since there was some discussion that asking for something to be face-to-face was not ok because we need hybrid, I just thought it'd be useful to state that when we do face-to-face, we always mean hybrid.

> This is a case where if the Team begins acting consistently in a way that is not in line with the best interests of the membership (e.g., by making the AC meeting a one-hour teleconference rather than a several-day in-person event), there are other avenues that exists for fixing that.

I mean, kind of, but also kind of no. The Process is invoked from the Member agreement, and Members can lean on it to state "we were owed this" if it isn't happening. But if there was a requirement in the Process, and we then emove it, that is explicit permission not to do it anymore. Some can still argue that it would be nice to do anyway, but this goes from being a requirement to something people can honestly disagree about.

If the intent is to remove the expectation, and give the Team space to decide what to do about the meeting, then sure. But if intend for the expectation to still be there, and just want to make some room for occasional exceptions (Pandemic, what have you), then I think we should say that.

nigelmegitt:
>That seems like a stretch. Sure, we could maybe be a little more specific, but the phrasing isn't "make sure anyone on earth (or in space) can attend, regardless of their location, physical condition, or lack of connectivity". Just stating that there needs to be a way to attend without being present on site. Basically, it's what we do already, but since there was some discussion that asking for something to be face-to-face was not ok because we need hybrid, I just thought it'd be useful to state that when we do face-to-face, we always mean hybrid.

Yes! It was intended as a _reductio ad aburdam_ argument. It's extreme, but within the range of potential interpretations of the current proposed wording. I note that the W3C has at times invested a lot of money in some facilities, such as the dedicated video conference screens, cameras and microphones that are used annually at TPAC. It's a long time ago now, but I also recall that W3C did set up specialist dedicated internet connectivity for TPAC 2013 in Shenzhen, as an example.

So maybe it _is_ reasonable to have an open scope here, but I still think it would be unwise to make it a MUST in that case. Also note that it in the passive voice, "must be provided" doesn't say _who_ must provide them.

So two potential improvements, take your pick or add your own:

1. Should not must, and allow for provision or arrangement:

```
Nevertheless, facilities <em class=rfc2119>should</em> be provided or arranged by W3C to enable remote attendance.
```

2. Must but slightly tighter scoping, with must and allowing for provision or arrangement:

```
Nevertheless, teleconferencing facilities <em class=rfc2119>must</em> be provided or arranged by W3C to enable remote attendance.
```

brentzundel (approved):

PR #1178, opened 2026-07-09
Remove AB from TAG appointee ratification

### LINKED ISSUE #810: AB Role in TAG Appointment
Decisions about TAG appointments are currently [ratified by a two-thirds majority of both the TAG and the AB](https://www.w3.org/2023/Process-20231103/#TAG-appointments).  I do not see a reason for the AB to be involved in this decision at that level.

For boards and groups operating at the level of the TAG, an appointment process is a good way to ensure that there is sufficient diversity of members available.  Particularly when elections are used to fill seats, the process of filling seats can leave gaps in expertise or other qualities that would diminish the capacity of the group to function effectively.

A more regular process would be to give the group itself the ability to fill a small number of positions.  For some reason, the W3C process gives this responsibility to the Team.  That is somewhat irregular, but not really the subject of this issue.

The decision by the Team then needs to be approved by the TAG with a two-thirds majority, which is entirely appropriate.  If the TAG believes that they cannot work with an appointed person, or the appointments do not address a diversity shortfall, then the TAG needs to be able to have a say.

The AB is given the same approval power, something that seems poorly justified.  Obviously, the AB and TAG frequently work together, but while the two functions are complementary, the AB has no significant stake in the composition of the TAG or the technical work that the TAG does.  The function of the AB is to advise the team, yes, but that is generally limited to procedural and operational matters.

Giving the AB input into TAG appointments at the same level as the TAG is unnecessary.  I see no substantive reason that the AB needs a role that is significantly greater than any other consortium member.

If the goal of having the AB vote on appointments is to provide better accountability to the members of the organization, the formal objection process is a more appropriate way to ensure that membership has a way to object to particular appointments.  That option is available to the AB just like any community member.

I propose that the process be amended to remove the AB from this process.

frivoal:
Per AB decision, as reported in https://github.com/w3c/process/issues/810.

***
<a href="https://pr-preview.s3.amazonaws.com/frivoal/w3process/pull/1178.html" title="Last updated on Jul 9, 2026, 5:04 PM UTC (bd04097)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1178/658acd5...frivoal:bd04097.html" title="Last updated on Jul 9, 2026, 5:04 PM UTC (bd04097)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1291,7 +1291,7 @@ Technical Architecture Group Appointments</h5>
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
 	by the [=TAG=],
-	requiring a two-thirds approval.
+	requiring two-thirds approval.
 	In the case of regularly scheduled elections,
 	the [=TAG=] participants in this ratification are
 	its members for the upcoming term.
--- a/index.bs
+++ b/index.bs
@@ -1290,8 +1290,8 @@ Technical Architecture Group Appointments</h5>
 
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
-	by both the [=AB=] and the [=TAG=],
-	requiring a two-thirds approval from each.
+	by the [=TAG=],
+	requiring a two-thirds approval.
 	In the case of regularly scheduled elections,
 	the [=TAG=] participants in this ratification are
 	its members for the upcoming term.


index.bs
```diff
@@ -1289,8 +1290,8 @@ Technical Architecture Group Appointments</h5>
 
 	The [=Team=]'s choice of appointee(s)
 	is subject to ratification by secret ballot
-	by both the [=AB=] and the [=TAG=],
-	requiring a two-thirds approval from each.
+	by the [=TAG=],
+	requiring a two-thirds approval.
```
hober:
```suggestion
	requiring two-thirds approval.
```

hober (approved):
r+ with nit

brentzundel (approved):

tidoust (approved):

torgo (approved):

hidde (approved):

PR #1189, opened 2026-09-09
Proposal for Process Experimentation

ianbjacobs:
The W3C Process Document is managed by the Advisory Board. Changes to the Process Document are reviewed by the Advisory Committee and adopted by consensus. The Membership has not explicitly granted the Advisory Board the authority to conduct experiments related to the existing process.

This is a proposal to legitimize temporary alternative Processes authorized by the Advisory Board.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/process/pull/1189.html" title="Last updated on Sep 15, 2026, 3:27 PM UTC (e7498a8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/process/1189/f92ce2c...e7498a8.html" title="Last updated on Sep 15, 2026, 3:27 PM UTC (e7498a8)">Diff</a>


index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
 	and the Process in the <a href="https://www.w3.org/community/w3process/">W3C Process Community Group</a>.
 
+<h2 id="process-experimentation">Process Experimentation</h2>
+
+The Advisory Board may authorize Process Experiments.  
+An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document, 
+Patent Policy (in partnership with the PSIG),
```
hober: I'd like to explicitly exclude the patent policy from this, at least in v1.

* What would an experiment here even be? Running a year long experiment with RAND instead of RF, for instance, is simply not something we should entertain.
* I don't think it's a likely area for experimentation in the short term, anyway.

Let's prove the usefulness of this experiment mechanism with other aspects of our governing documents, and leave the patent policy out of it for now.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document, 
+Patent Policy (in partnership with the PSIG), 
+and Code of Conduct (in partnership with the Positive Work Environment CG), 
+subject to the scope limitations defined below.
+
+<h3>Notification</h3>
+
+The Advisory Boards must notify the Membership and public of its decision to authorize a Process Experiment
```
hober, line 5336: Why "and public"? If the intent is to reach W3C participants who are not AC reps, surely we have a way to do that that isn't broadcasting wacky stuff to the public at large?

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+and include the language that temporarily replaces those provisions.</li>
+<li>Specify an end date that may not be more than 1 year after the start date of the experiment.</li>
+</ul>
+
+<h3 id="experiment-scope">Scope limitations</h3>
+
+Authorized Process Experiments must not:
+
+<ul>
+<li>Alter the rules that give it the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
```
hober:
```suggestion
<li>Alter the rules that give the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
```

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+</ul>
+
+<h3>Extensions</h3>
+
+An Authorized Process Experiment expires automatically on its specified end date but may be extended at most one time for up to one year. 
+If the Advisory Board wishes to extend an experiment:
+
+<ul>
+<li>the Advisory Board must issue a <a href="#process-experiment-report">Process Experiment Report</a></li>
+<li>the extension must be approved by majority vote of the Advisory Committee by a ballot open for 10 business days at a minimum.</li>
```
hober: First, a simplification:

```suggestion
<li>the extension must be approved by the Advisory Committee by a ballot open for 10 business days at a minimum.</li>
```

Second, when the AC is polled using WBS, one of the options included is the ability to FO. Should this text say something about that not being the case? Or is it okay for FOs to happen to extension requests?

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+</ul>
+
+<h3>Process Experiment Reports</h3>
+
+Upon conclusion or termination of a Process Experiment and prior to any extension, 
+the Advisory Board must share a <dfn id="process-experiment-report">Process Experiment Report</dfn> with the Membership and public.
```
hober, line 5376: Same question as above re "and public".

hober (changes requested):
This is great! It's not quite ready to land, but I'd be okay with landing something awfully close to this.

index.bs
ianbjacobs: I included it to avoid ambiguity about inclusion of policies by reference. I don't mind at all an explicit choice to exclude it from early experimentation.

index.bs
ianbjacobs: The experimental process provisions themselves will likely be developed in public (in the Process CG) and will need to be public because they have an impact on people beyond those with Member access. After that, it seems both proper and simpler to make the other bits (e.g., rationale, duration) public as well.

index.bs
ianbjacobs: I am not anticipating a usual AC review form, but an up or down vote in this case. 

In pursuit of ability, and because the AC is making the decision directly (without any party further assessing the outcome of the Member perspectives), I don't see the same need for FOs.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
 
+<h2 id="process-experimentation">Process Experimentation</h2>
+
+The Advisory Board may authorize Process Experiments.  
+An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document, 
+Patent Policy (in partnership with the PSIG), 
+and Code of Conduct (in partnership with the Positive Work Environment CG),
```
frivoal: As per https://www.w3.org/policies/process/#GAProcess and the note it contains, the delegation to PSIG about the patent policy and to the PWE-CG for the code of conduct are not hard coded in the process. They are delegations from the AB, and I don't think this text should elevate them as a side effect.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+and Code of Conduct (in partnership with the Positive Work Environment CG), 
+subject to the scope limitations defined below.
+
+<h3>Notification</h3>
+
+The Advisory Boards must notify the Membership and public of its decision to authorize a Process Experiment 
+no less than three weeks before the planned start of the experiment. 
+Provided the experiment remains within <a href="#experiment-scope">scope limitations</a>, 
+the authorization itself is not subject to the usual Formal Objection mechanisms, but it is subject to Advisory Committee Appeal.
+The usual Process safeguards (objections, appeals) remain in place for decisions made during the course of an experiment.
```
frivoal: This seems to imply that experiments cannot override or modify the ability to object to certain things, which I don't think is what you intend, since I understand that some of the intended experiments are precisely about how to handle disagreements in a variety of situations, including charter creation.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+<li>Specify an end date that may not be more than 1 year after the start date of the experiment.</li>
+</ul>
+
+<h3 id="experiment-scope">Scope limitations</h3>
+
+Authorized Process Experiments must not:
+
+<ul>
+<li>Alter the rules that give the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
+<li>Affect the W3C Member Agreement or incorporation documents (bylaws, certificate of incorporation).</li>
```
frivoal: "W3C Member Agreement or incorporation documents (bylaws, certificate of incorporation)" are not merely things the AB must not experiment with, but things the AB cannot experiment with. These are not under AB, AC, or Process purview. They don't belong in this list.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+<li>Alter patent policy commitments.</li>
```
frivoal, line 5359: What does that mean? a strict interpretation would be that nothing about the REC track or about decision making can be changed, because that would change (possibly significantly) what can be in documents about which patent commitments are made. I suspect you don't mean that, as the REC track is probably something we do want to run experiments about.

It might mean that section 5 of the patent policy cannot be changed by an experiment. That too seems unlikely to be what's intended, as it feels far too narrow to actually protect the continuity of commitments.

So something in between is likely intended, but I am not sure what.

index.bs
frivoal: The process is public. If some of its clauses are temporarily suspended or altered, the same people who can read the process need to be able to discover and read the suspensions and alterations.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+<li>Describe what problem(s) the experiment seeks to address.</li>
+<li>Include hypotheses for how the experiment will address these problems and what success would look like.</li>
+<li>Identify the policy provisions that are temporarily suspended during the experiment, 
+and include the language that temporarily replaces those provisions.</li>
+<li>Specify an end date that may not be more than 1 year after the start date of the experiment.</li>
+</ul>
+
+<h3 id="experiment-scope">Scope limitations</h3>
+
+Authorized Process Experiments must not:
```
frivoal, line 5354: What happens if the experiments (inadvertently maybe) do some of the things they supposedly must not do? Are they automatically void? Does that grant someone (who) the ability to shut them down?

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+<ul>
+<li>Alter the rules that give the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
+<li>Affect the W3C Member Agreement or incorporation documents (bylaws, certificate of incorporation).</li>
+<li>Alter patent policy commitments.</li>
+<li>Have retroactive effect.</li>
+</ul>
+
+<h3>Extensions</h3>
+
+An Authorized Process Experiment expires automatically on its specified end date but may be extended at most one time for up to one year.
```
frivoal, line 5365: I don't know how this can actually work. It seems that the AB can make an experiment to be slightly different in some way but substantially the same, and then it would be a new experiment, not an extension of an existing one, and then they'd need no permission.

Few things are as permanent as temporary measures.

index.bs
```diff
@@ -5322,6 +5322,79 @@ Process Evolution</h2>
...
+Patent Policy (in partnership with the PSIG), 
+and Code of Conduct (in partnership with the Positive Work Environment CG), 
+subject to the scope limitations defined below.
+
+<h3>Notification</h3>
+
+The Advisory Boards must notify the Membership and public of its decision to authorize a Process Experiment 
+no less than three weeks before the planned start of the experiment. 
+Provided the experiment remains within <a href="#experiment-scope">scope limitations</a>, 
+the authorization itself is not subject to the usual Formal Objection mechanisms, but it is subject to Advisory Committee Appeal.
```
frivoal, line 5339: The inclusion of AC Appeal is a step in the right direction, but a small one, I am concerned that the ability for the AB to modify the process largely at will and with no authorization is a giant shift in our power structure, and a shift away from a "rule of law" type of governance.

Sure, there are some time limitations (but I don't think they're much of a limitation in practice, see https://github.com/w3c/process/pull/1189/changes#r3974937741), and there are some very small scope limitations, but this effectively gives the AB the ability to change any and all parts of the Process with hardly any oversight. The name "Advisory Board", alongside with its description that says "has no decision-making authority within W3C; its role is strictly advisory" then becomes really inappropriate. It becomes something more suited to be called maybe an Executive Committee or Directorate.

Sure, we do have elections which can bring some measure of accountability into this, but even then, it is still a massive shift from being governed by collectively approved rules to enabling governance by executive orders.


There is a broader lesson about governance here. We may think we don't need detailed rules because we have longstanding norms of good behavior, or that we can always vote out anyone who abuses their power. But norms can be ignored, and electoral safeguards are only useful if they actually work.

Likewise, giving the AB broad power to override the Process may seem efficient, provided we trust it to use that power responsibly. But when rules can be set aside at will, it becomes unclear whether an exercise of power is actually lawful, and who has the authority to say so when it isn't.

This is why the rule of law matters: it is designed for when good governance norms fail.

frivoal (commented):
I support the idea experimentation, but I am skeptical of the approach proposed here.

index.bs
ianbjacobs: I have removed them from the scope.

index.bs
ianbjacobs: I have removed them from the scope.

index.bs
ianbjacobs: Can you suggest a clarification? I am trying to say "can't raise an FO on conducting an experiment, but can create an FO within the context of the experiment"

index.bs
ianbjacobs: I have removed it from the list

index.bs
ianbjacobs: Let's back up a bit:

* My thought is that an experiment should not introduce ambiguity about whether someone has made a commitment to the W3C patent policy or whether a specification benefits from protections under those commitments. The first question is: is that a worthwhile goal? If not, or if we think that the people designing the experiments should have flexibility, then we can simplify the proposal. My assumption was that we would not want ambiguity or doubt around patent policy guarantees.
* I have not studied the question in depth yet, but my initial thought was that we would get an understanding of the boundaries of experimentation related to the patent policy by looking at the patent policy. It includes provisions that (unless modified) set expectations. For example, it talks about charters referring to the policy, so we should make sure that's still a thing in a charter. It talks about "Working Group participants" making commitments, so we should not (at least for now) move to a model where specifications are developed outside of Working Groups ("workstreams" come to mind). As you have pointed out, the Patent Policy relies on a "Patent Review Draft" so we need one of those, however we get it. In short, my hope is that looking closely at the patent policy will tell us what we need to preserve in the process document.

index.bs
ianbjacobs: I agree with you that we should have a way to void the "substantive" outcome of an experiment, such as a final decision after experimenting with some new approach to handling formal objections. 

The proposal requires the AB to publish a Report at the conclusion of an experiment. We could take this further:

* The AB announces the report. 
* If the AB itself concludes that the experiment did not remain within the scope limitations, via the Report the AB declares the substantive outcome to be void. Otherwise, the AB declares the outcome to be valid.
* The AC can appeal that declaration (either way).

Goals of this proposal:

* Empower the AB to acknowledge (possibly after receiving input from the community) issues with the experiment, to stop it early if needed, and to limit damages (perhaps mostly to frustrated parties who were in the experiment).
* Provide some agility to the AB to validate experimental results (through the appeal bar) while still giving Members final decision-making power.

index.bs
ianbjacobs: The AC gets to approve the extension. If the AC believes the AB is trying to pull a fast one and work around the process, then they can vote down the extension.

hober: Relying on vigilance from the AC is an imperfect mechanism, but the right one despite its imperfections, if you ask me. If the membership can't bother to lead a member-led organization, we have bigger fish to fry than the efficacy of this new mechanism.

index.bs
frivoal: Maybe something like this:
```suggestion
The usual Process safeguards (objections, appeals), to the extent they are not themselves modified by the experiment, remain in place for decisions made during the course of an experiment.
```

index.bs
frivoal: The AC gets to approve (or not) the extension if it is declared to be an extension, but the AB gets to decide whether to describe it as an extension or not in the first place. If the AB declares something to be a different experiment rather than an extension, nobody is asked to check how much of a stretch that is.

ianbjacobs: We now enable the AC to appeal any experiment. So if they think the AB is pulling a fast one by calling it a new experiment, they can appeal it.

This raises the question of whether an extension mechanism is necessary if the AC can appeal every experiment. But I think it's useful to set expectations, and I have in mind:

* Experiments must be time-limited.
* If more time is needed because the experiment is promising, one extension should not seen as ok.
* But at any time the AC an appeal an experiment, and the AB should be held to account if they appear to be abusing the process.

index.bs
hober: > I am concerned that the ability for the AB to modify the process largely at will and with no authorization

I find this characterization of the proposed experiment mechanism to be at odds with the text of the proposal. The text requires (with RFC 2119 MUST) the AB to tell the AC about each and every experiment they’d like to run, weeks before they run them, and the AC is empowered to appeal any and all of them.

If the AB proposes to run an experiment in which it suspends all rules and lets anarchy prevail for a year, and we can’t find a single AC rep to appeal that, we have much bigger problems. That said, I'm encouraged by the effort elsewhere in the discussion on this PR where we're all trying to make sure experiments have the right sort of guardrails etc.

> is a giant shift in our power structure, and a shift away from a "rule of law" type of governance.

No, it’s not. In fact, I’d argue the opposite. It’s enabling the Membership to more fully realize a Member-led consortium, by giving it necessary tools to experiment with process changes, so that we can identify concrete changes that we can be confident will actually improve the effectiveness of the organization re: meeting the Membership’s needs.

> This is why the rule of law matters: it is designed for when good governance norms fail.

I agree!

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+and include the language that temporarily replaces those provisions.</li>
+<li>Specify an end date that may not be more than 1 year after the start date of the experiment.</li>
+</ul>
+
+<h3 id="experiment-scope">Scope limitations</h3>
+
+Authorized Process Experiments must not:
+
+<ul>
+<li>Extend to policies included by reference in the Process Document.
```
tidoust:
```suggestion
<li>Extend to policies included by reference in the Process Document.</li>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+<h2 id="process-experimentation">Process Experimentation</h2>
+
+The Advisory Board may authorize Process Experiments.  
+An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document
+subject to the scope limitations defined below. 
+
+<h3>Notification</h3>
```
tidoust: Seems like all headings have explicit IDs in the Process document

```suggestion
<h3 id="experiment-notification">Notification</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+Authorized Process Experiments must not:
+
+<ul>
+<li>Extend to policies included by reference in the Process Document.
+<li>Alter the rules that give the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
+<li>Alter patent policy commitments.</li>
+<li>Have retroactive effect.</li>
+</ul>
+
+<h3>Extensions</h3>
```
tidoust:
```suggestion
<h3 id="experiment-extensions">Extensions</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+An Authorized Process Experiment expires automatically on its specified end date but may be extended at most one time for up to one year. 
+If the Advisory Board wishes to extend an experiment:
+
+<ul>
+<li>the Advisory Board must issue a <a href="#process-experiment-report">Process Experiment Report</a></li>
+<li>the extension must be approved by majority vote of the Advisory Committee by a ballot open for two weeks at a minimum.</li>
+</ul>
+
+<h3>Process Experiment Reports</h3>
```
tidoust:
```suggestion
<h3 id="experiment-reports">Process Experiment Reports</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<ul>
+<li>Describe at a high level what happened, what was learned, and what evidence was gathered.</li>
+<li>Indicate whether the hypotheses in the Process Experiment Authorization were supported, 
+and what unexpected events occurred.</li>
+<li>Advise whether the experiment be extended, modified, or proposed for permanent adoption, 
+or should no further action be taken.</li>
+</ul>
+
+<h3>Experiment outcomes</h3>
```
tidoust:
```suggestion
<h3 id="experiment-outcomes">Experiment outcomes</h3>
```

tidoust (approved):

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
 	the [[COC|Code of Conduct]] in the <a href="https://www.w3.org/community/pwe/">Positive Work Environment Community Group</a>,
 	and the Process in the <a href="https://www.w3.org/community/w3process/">W3C Process Community Group</a>.
 
+<h2 id="process-experimentation">Process Experimentation</h2>
```
frivoal:
```suggestion
<h2 id="process-experimentation">
Process Experimentation</h2>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+The Advisory Board may authorize Process Experiments.  
+An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document
+subject to the scope limitations defined below.
```
frivoal:
```suggestion
	The Advisory Board may authorize Process Experiments.  
	An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
	one or more provisions of an operative policy for the authorized duration and scope of the experiment.

	Authorized Process Experiments may be conducted over the Process Document
	subject to the scope limitations defined below. 
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+<h2 id="process-experimentation">Process Experimentation</h2>
+
+The Advisory Board may authorize Process Experiments.  
+An <dfn id="authorized-process-experiment">Authorized Process Experiment</dfn> modifies or replaces 
+one or more provisions of an operative policy for the authorized duration and scope of the experiment.
+
+Authorized Process Experiments may be conducted over the Process Document
+subject to the scope limitations defined below. 
+
+<h3 id="experiment-notification">Notification</h3>
```
frivoal:
```suggestion
<h3 id="experiment-notification">
Notification</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+The Advisory Boards must notify the Membership and public of its decision to authorize a Process Experiment 
+no less than three weeks before the planned start of the experiment. 
+Provided the experiment remains within <a href="#experiment-scope">scope limitations</a>, 
+the authorization itself is not subject to the usual Formal Objection mechanisms, but it is subject to Advisory Committee Appeal.
```
frivoal:
```suggestion
	The Advisory Boards must notify the Membership and public
	of its decision to authorize a Process Experiment 
	no less than three weeks before the planned start of the experiment. 
	Provided the experiment remains within <a href="#experiment-scope">scope limitations</a>, 
	the authorization itself is not subject to the usual Formal Objection mechanisms,
	but it is subject to Advisory Committee Appeal.
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+The usual Process safeguards (objections, appeals), to the extent they are not themselves modified by the experiment, remain in place for decisions made during the course of an experiment.
+
+Notification of a Process Experiment Authorization must:
```
frivoal:
```suggestion
	The usual Process safeguards (objections, appeals),
	to the extent they are not themselves modified by the experiment,
	remain in place for decisions made during the course of an experiment.

	Notification of a Process Experiment Authorization must:
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<ul>
+<li>Describe what problem(s) the experiment seeks to address.</li>
+<li>Include hypotheses for how the experiment will address these problems and what success would look like.</li>
+<li>Identify the policy provisions that are temporarily suspended during the experiment, 
+and include the language that temporarily replaces those provisions.</li>
+<li>Specify an end date that may not be more than 1 year after the start date of the experiment.</li>
+</ul>
+
+<h3 id="experiment-scope">Scope limitations</h3>
```
frivoal:
```suggestion
<h3 id="experiment-scope">
Scope limitations</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+Authorized Process Experiments must not:
```
frivoal:
```suggestion
	Authorized Process Experiments must not:
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<ul>
+<li>Extend to policies included by reference in the Process Document.</li>
+<li>Alter the rules that give the Advisory Board the authority to conduct experiments or permanently change a policy.</li>
+<li>Alter patent policy commitments.</li>
```
frivoal:
```suggestion
	<ul>
		<li>Extend to policies included by reference in the Process Document.
		<li>Alter the rules that give the Advisory Board the authority to conduct experiments
			or permanently change a policy.
	<li>Alter patent policy commitments.
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+<li>Have retroactive effect.</li>
+</ul>
```
frivoal:
```suggestion
		<li>Have retroactive effect.
	</ul>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<h3 id="experiment-extensions">Extensions</h3>
```
frivoal:
```suggestion
<h3 id="experiment-extensions">
Extensions</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+An Authorized Process Experiment expires automatically on its specified end date but may be extended at most one time for up to one year.
```
frivoal:
```suggestion
	An Authorized Process Experiment expires automatically on its specified end date
	but may be extended at most one time for up to one year. 
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<h3 id="experiment-extensions">Extensions</h3>
+
+An Authorized Process Experiment expires automatically on its specified end date but may be extended at most one time for up to one year. 
+If the Advisory Board wishes to extend an experiment:
+
+<ul>
+<li>the Advisory Board must issue a <a href="#process-experiment-report">Process Experiment Report</a></li>
+<li>the extension must be approved by majority vote of the Advisory Committee by a ballot open for two weeks at a minimum.</li>
+</ul>
```
frivoal:
```suggestion
	If the Advisory Board wishes to extend an experiment:

	<ul>
		<li>the Advisory Board must issue a [=Process Experiment Report=]
		<li>the extension must be approved by majority vote of the Advisory Committee
			by a ballot open for two weeks at a minimum.
	</ul>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<h3 id="experiment-reports">Process Experiment Reports</h3>
```
frivoal:
```suggestion
<h3 id="experiment-reports">
Process Experiment Reports</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+Upon conclusion or termination of a Process Experiment and prior to any extension, 
+the Advisory Board must share a <dfn id="process-experiment-report">Process Experiment Report</dfn> with the Membership and public.
```
frivoal:
```suggestion
	Upon conclusion or termination of a Process Experiment and prior to any extension, 
	the Advisory Board must share a <dfn id="process-experiment-report">Process Experiment Report</dfn> with the Membership and public.
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+A Process Experiment Report must:
+
+<ul>
+<li>Describe at a high level what happened, what was learned, and what evidence was gathered.</li>
+<li>Indicate whether the hypotheses in the Process Experiment Authorization were supported, 
+and what unexpected events occurred.</li>
+<li>Advise whether the experiment be extended, modified, or proposed for permanent adoption, 
+or should no further action be taken.</li>
+</ul>
```
frivoal:
```suggestion
	A Process Experiment Report must:

	<ul>
		<li>Describe at a high level what happened, what was learned, and what evidence was gathered.
		<li>Indicate whether the hypotheses in the Process Experiment Authorization were supported, 
			and what unexpected events occurred.
		<li>Advise whether the experiment be extended, modified, or proposed for permanent adoption, 
			or should no further action be taken.
	</ul>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+<h3 id="experiment-outcomes">Experiment outcomes</h3>
```
frivoal:
```suggestion
<h3 id="experiment-outcomes">
Experiment outcomes</h3>
```

index.bs
```diff
@@ -5322,6 +5322,77 @@ Process Evolution</h2>
...
+
+Not all experiments will be deemed successful or lead to permanent changes to the Process Document. 
+The validity of actions or decisions taken as part of an authorized Process Experiment does not depend on the outcome of the experiment.
+
+An Authorized Process Experiment does not amend the operational policy permanently; 
+making any change permanent requires the usual approval process.
```
frivoal:
```suggestion
	Not all experiments will be deemed successful or lead to permanent changes to the Process Document. 
	The validity of actions or decisions taken as part of an authorized Process Experiment
	does not depend on the outcome of the experiment.

	An Authorized Process Experiment does not amend the operational policy permanently; 
	making any change permanent requires the usual approval process.
```

frivoal (changes requested):
Submitting a batch of suggested changes to align with the coding style documented at the beginning of the file, to keep these edits consistent with the rest of the document

ianbjacobs:
Based on discussion at the 23 September Process CG call, I have included a provision in the notification section that the affected audience be identified. When the audience is "narrow" the AB must socialize the experiment with the audience (e.g., a WG) before notification. When the audience is "wide" the AB must provide rationale for that scope.

hober (approved):
This is awesome; thank you for all your hard work! :)

